PR #3, opened 2026-04-25
Harden sweep close safety

stainlu:
## Summary
- make review shards proposal-only and keep apply-existing as the only close path
- block protected-label items from normal planning and from apply-time closes
- add runtime decision validation for schema shape, evidence, close comments, close reason/type combos, and implemented-on-main source evidence
- reduce review job credential exposure, pin Codex, and upload only review markdown artifacts

## Verification
- npm run check


steipete (changes requested):
Codex review found one small docs mismatch before this lands. The code path now intentionally removes `OPENAI_API_KEY` / `CODEX_API_KEY` from the `codex exec` environment and the workflow no longer passes those env vars into the review step, but README still says the workflow "passes `CODEX_API_KEY` to `codex exec`". Please update that setup paragraph to match the new login-only flow.\n\nVerification on a current-main merge with 601d8b04: `npm ci && npm run check` passes.

steipete (approved):
Codex re-review: requested README auth wording is fixed. Verified on a current-main merge (`93c298d7`) with `npm ci && npm run check`; green.

PR #5, opened 2026-04-25
Retry transient GitHub API failures

stainlu:
## Summary
- classify GitHub failures as throttle vs transient vs non-retryable
- keep the existing long backoff and throttle heartbeat path for secondary/rate limits
- add shorter retries for transient GitHub API/network failures such as unexpected EOF, connection resets, gateway errors, and temporary timeouts
- document the transient retry behavior in the workflow overview

## Verification
- npm run check
- npm run reconcile -- --dry-run

## Dry-run evidence
During the reconcile dry run, the scanner hit a transient failure on page 62 and recovered instead of crashing:

```text
Transient GitHub API failure; retrying gh api repos/openclaw/openclaw/issues?state=open&sort=created&direction=asc&per_page=100&page=62 in 2s
```

The dry run then completed successfully with 9,903 open items seen across 100 scanned pages.


steipete:
Landed this on main as 41c353a8 with the conflict resolved and the retry classifier tightened. I kept the transient retry behavior, preserved author credit, and added a regression so a non-retryable error on a path like /issues/502/comments is not misclassified as a 502 gateway retry.\n\nVerified with npm run check.

PR #6, opened 2026-04-25
fix: isolate Codex home for local review runs

ZHOUKAILIAN:
## Summary
- isolate the Codex child process from user-level `~/.codex` skills and AGENTS baggage during ClawSweeper review runs
- keep only `auth.json` and `config.toml` in a temporary minimal `.codex` home for the spawned Codex process
- avoid secondary crashes when `spawnSync` returns missing stdout/stderr by using a safe output tail helper
- add regression tests for both behaviors

## Why
Local review runs were failing for two separate reasons:
1. ClawSweeper could crash on `undefined.slice(...)` when Codex process output was missing on failure paths.
2. The spawned `codex exec` inherited the user's full `~/.codex` environment, which could load personal AGENTS / superpowers skills and significantly distort or slow local review behavior compared with CI.

This change makes local review execution more stable and closer to CI by isolating the spawned Codex environment while preserving the auth/config it still needs.

## Verification
- `npm run check`
- local smoke run:
  - `node dist/clawsweeper.js review --openclaw-dir ../openclaw --item-number 45049 --batch-size 1 --max-pages 1 --codex-timeout-ms 120000`

### DIFF
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -5,13 +5,16 @@ import {
   chmodSync,
   existsSync,
   mkdirSync,
+  mkdtempSync,
   readFileSync,
   readdirSync,
   renameSync,
+  rmSync,
   statSync,
   unlinkSync,
   writeFileSync,
 } from "node:fs";
+import { tmpdir } from "node:os";
 import { basename, dirname, join, relative, resolve } from "node:path";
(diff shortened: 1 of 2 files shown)


ZHOUKAILIAN:
Closing this in favor of #7, which keeps only the narrow undefined.slice crash fix.

PR #7, opened 2026-04-25
fix: guard missing Codex process output

ZHOUKAILIAN:
## Summary
- add a `safeOutputTail` helper for Codex child-process output
- avoid calling `.slice(...)` on missing `stdout`/`stderr` in ClawSweeper review failures
- add a regression test covering missing process output

## Why
Local review runs could crash while formatting Codex failure details if `spawnSync` returned missing `stdout` or `stderr`. That secondary crash hides the real Codex failure and breaks review artifact generation.

## Verification
- `npm run check`


steipete:
Thanks!

PR #27, opened 2026-05-01
Normalize repository profile lookup

xthunder0:
## Summary

- Normalize each configured `targetRepo` during `repositoryProfileFor()` lookup so mixed-case profile entries match normalized input.
- Add regression coverage for mixed-case caller input and mixed-case configured profile entries.

## Context

This routes the downstream stop-gap from cloverthe-ai/clover-sweeper#16 upstream for cloverthe-ai/clover-sweeper#15. The latent bug is that lookup normalized only the caller input, then compared it against raw `candidate.targetRepo`, so a mixed-case profile entry could fail to match itself and produce an unsupported-repo error that listed the same repo as known.

Once this lands upstream, the downstream patch can be retired on the next clover-sweeper re-vendor cycle.

## Verification

- `pnpm run build && node --test test/repository-profiles.test.ts`
- `pnpm run check`

Note: both checks ran on this machine with Node v22.22.1, which emits the repo's existing `>=24` engine warning. No live sweep/apply/close commands were run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ checkpoint, and status-only commits are intentionally omitted.
   `docs.openclaw.ai` links where appropriate.
 - Clarified `incoherent` close-reason wording so rendered reports no longer
   collide with `not_actionable_in_repo` (#29). Thanks @xthunder0.
+- Normalized repository profile lookup against configured target repos so
+  mixed-case profile entries resolve correctly (#27). Thanks @xthunder0.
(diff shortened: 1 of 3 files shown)


steipete:
Landed via temp rebase onto main.

- Gate: pnpm run check
- Land commit: 01c6097002d1c84fa76e9fe0d60f6e475ec4665b
- Merge commit: 62e607cbb875cfde12974714a73891482fac650d

Thanks @xthunder0!

PR #33, opened 2026-05-03
profiles: add iambuddy829-del/harnes target (private adopter)

iambuddy829-del:
## Summary

Adds a fourth `RepositoryProfile` so ClawSweeper can review and (with App + labels) auto-merge trusted PRs on `iambuddy829-del/harnes`.

## Context

I'm adopting OpenClaw + Codex-CLI to manage Harnes, a defensive audit/readiness harness for our CanTime ops Mac. Codex-generated PRs land there frequently and currently get manually reviewed. The goal of adopting ClawSweeper is to gate those trusted PRs through the same exact-head review + bounded autofix/automerge loop you already run for `openclaw/openclaw`.

## Profile choice

Mirrors the `openclaw/clawsweeper` self-review profile (most conservative):

- `pull_request` close rules limited to `["implemented_on_main"]`
- `issue` close rules empty (no auto-close at all)
- No `docsUrl` / `communityUrl` (private project, no public docs)

## Open question for maintainers

Would you prefer a config-driven approach — e.g. a `CLAWSWEEPER_EXTRA_PROFILES_PATH` env var pointing to a JSON file — so external adopters can self-onboard without an upstream PR per repo? Happy to refactor along that line if it fits your roadmap better.

If you'd rather keep `REPOSITORY_PROFILES` as the canonical curated list, this PR adds Harnes as a maintainer-blessed entry (same model as the three existing profiles).

## Test plan

- [x] TypeScript shape compliant (`RepositoryProfile` interface satisfied)
- [x] Existing `test/repository-profiles.test.ts` still green (uses `REPOSITORY_PROFILES[0]` — unchanged)
- [ ] Maintainers: full `pnpm test` + `pnpm build:all` per CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #37, opened 2026-05-04
feat: support external repository profiles

stainlu:
## Summary
- Add a guarded external repository profile loader for ClawSweeper operators.
- Keep built-in OpenClaw profiles curated in source while allowing external adopters to configure review targets via `CLAWSWEEPER_EXTRA_PROFILES_JSON` or local `CLAWSWEEPER_EXTRA_PROFILES_PATH`.
- Keep the external profile apply surface conservative: issue auto-close is always disabled, and pull-request auto-close may only opt into `implemented_on_main`.

## Why
ClawSweeper is starting to get adopter profile PRs for private or external repositories, for example https://github.com/openclaw/clawsweeper/pull/33. Those are policy-heavy and do not scale well as upstream source changes. A config-backed profile surface lets external users self-onboard without adding one-off private targets to the curated upstream profile list.

## What Changed
- Split built-in profiles from the merged runtime profile list.
- Parse and validate extra profile JSON with duplicate target/slug protection.
- Pass `CLAWSWEEPER_EXTRA_PROFILES_JSON` from GitHub repository variables into the sweep and commit-review workflows.
- Document the external profile contract in the README.
- Add focused coverage for defaults, allowed PR close policy, rejected issue close policy, and duplicate built-in targets.

## Safety
- External profiles cannot override built-in OpenClaw targets.
- External profiles cannot enable issue auto-close.
- External PR close rules are limited to `implemented_on_main` only.
- Unknown JSON keys fail closed instead of being ignored.

## Verification
- `pnpm run build`
- `node --test test/repository-profiles.test.ts`
- `pnpm run lint:src`
- `pnpm run format:check`
- `git diff --check -- src/repository-profiles.ts test/repository-profiles.test.ts README.md .github/workflows/sweep.yml .github/workflows/commit-review.yml`

`pnpm run check` was also attempted locally. It passed build, lint, unit tests, and most repair tests, then hit the existing local Node 22 notifier-test cancellation in `test/repair/notify-events.test.ts` and `test/repair/notify-github-activity.test.ts`. This repo declares Node >=24; the local host only has Node v22.22.0.


steipete:
Thanks for putting this together. The config-backed profile idea is reasonable for self-hosted ClawSweeper deployments, but we do not want the upstream/OpenClaw-hosted instance to look like a free third-party review service. External projects should fork/deploy ClawSweeper in their own org and configure their own targets there, rather than asking OpenClaw to run reviews for them.

I clarified that boundary in the README on main: https://github.com/openclaw/clawsweeper/commit/20ed692452

Closing this PR so we do not add an upstream external-target surface right now. The useful part of the direction is preserved as guidance for self-hosting, and we can revisit a self-hosting-specific config surface later if there is a concrete operator need.

PR #42, opened 2026-05-05
perf: reduce live worker capacity probes

stainlu:
Reduces repair live-capacity GitHub Actions API probes by fetching recent workflow runs once and filtering active statuses locally. Also avoids an immediate duplicate capacity probe in the wait-for-capacity dispatch loop.\n\nValidation: build:repair, focused live-worker-capacity test, targeted oxlint, targeted oxfmt, git diff --check. Direct build in this checkout is blocked by non-writable generated dist files, so validation ran in /private/tmp/clawsweeper-perf-build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ checkpoint, and status-only commits are intentionally omitted.
 - Fixed live worker scheduling to filter GitHub Actions runs through supported
   `workflowName` JSON fields instead of silently falling back to zero active
   workers when `gh run list --workflow` is unavailable.
+- Reduced repair live-capacity polling from one GitHub Actions API request per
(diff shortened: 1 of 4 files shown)


steipete:
This optimization is still directionally useful for rate-limit pressure, but the branch is dirty against current `main`. Please rebase before review/landing.

During the refresh, make sure it composes with the current workflow-active-run fixes on main: live worker/run counting should use supported GitHub data and local filtering, not unsupported `gh run list --workflow` / status-flag fallbacks. After rebase, please rerun `pnpm run check`.

stainlu:
Rebased on current main and reran `pnpm run check` under Node 24. This now uses the workflow-runs API with local active-status/prefix filtering and avoids the duplicate dispatch capacity probe.

PR #43, opened 2026-05-05
perf: cache comment router label lookups

stainlu:
Caches comment-router open-label issue lookups per process so autofix/automerge durable comment discovery and command synthesis do not repeat identical GitHub searches. Validation: build:repair, comment-router-core test, targeted oxlint, targeted oxfmt, git diff --check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ checkpoint, and status-only commits are intentionally omitted.
 - Fixed live worker scheduling to filter GitHub Actions runs through supported
   `workflowName` JSON fields instead of silently falling back to zero active
   workers when `gh run list --workflow` is unavailable.
+- Cached comment-router open-label issue lookups per run so repair-loop comment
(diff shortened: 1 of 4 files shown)


steipete:
The label-lookup cache is a small, sensible reduction in repeated GitHub calls, but the branch is dirty against current `main`. Please rebase and rerun `pnpm run check` before this can land.

The important behavior to preserve is per-process caching with copied return arrays, so callers cannot mutate cached state between lookups.

stainlu:
Rebased on current main and reran `pnpm run check` under Node 24. The cache still returns copied arrays so callers cannot mutate cached label-number state.

steipete:
Landed this manually after resolving the shared `CHANGELOG.md` conflict from the capacity-probe PR. The implementation and regression test stayed intact.

Landed on `main`: https://github.com/openclaw/clawsweeper/commit/ec1ea902a3

Thanks @stainlu.

PR #45, opened 2026-05-05
perf: cache review prompt assets

stainlu:
## Summary
- cache the static review prompt and decision schema lazily within each ClawSweeper process
- reuse cached bytes for review policy hashing, item prompt rendering, and the check command
- keep the schema path passed to Codex stable

## Validation
- pnpm run build
- node --test test/clawsweeper.test.ts
- pnpm run lint:src
- pnpm run lint:scripts
- pnpm run format:check
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,8 @@ checkpoint, and status-only commits are intentionally omitted.
   commit lists no longer process entries that are omitted from Codex input.
 - Thanks @stainlu for the repair prompt, GitHub pagination, lazy context
   compaction, and review telemetry PRs.
+- Cached the static review prompt and decision schema within each ClawSweeper
+  process instead of re-reading them during review planning and item prompts.
 
 ## 0.2.0 - 2026-05-03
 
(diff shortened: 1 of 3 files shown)


steipete:
This caching change is useful, but it needs a rebase against current `main`. The prompt/schema helper area has moved with the review telemetry work, so please adapt this to the current helper names instead of reintroducing duplicate file reads or parallel cache helpers.

After rebase, please rerun `pnpm run check`. The behavior to preserve is simple per-process caching of the static review prompt and decision schema while keeping the Codex `--output-schema` path unchanged.

stainlu:
Rebased on current main and reran `pnpm run check` under Node 24. This now caches the current `reviewPromptTemplate()` / `reviewDecisionSchemaText()` helpers and keeps the Codex `--output-schema` path unchanged.

steipete:
Landed manually on `main` as `fb9c2f32fd` (`perf: cache review prompt assets`) after resolving the current `CHANGELOG.md`, `src/clawsweeper.ts`, and test import conflicts. Kept `REVIEW_POLICY_VERSION` at v14, preserved the stable Codex `--output-schema` path, and kept the tracked-file regression. Full local `pnpm run check` passed before push. Thanks @stainlu.

PR #52, opened 2026-05-07
[Fix] Guard likely owner commit links

samzong:
### What's changed?

- Trim likely owner commit entries before rendering them.
- Render only SHA-like values as commit links in `Likely related people`.
- Add unit coverage for mixed provenance where a PR URL is present alongside a valid commit SHA.

### Why

ClawSweeper public comments can receive non-SHA provenance strings in `likelyOwners[].commits`. Before this change, those values were passed directly to the commit-link renderer, producing invalid GitHub commit URLs in visible comments.

Evidence from existing public ClawSweeper comments:

- https://github.com/openclaw/openclaw/issues/78671#issuecomment-4393034158 rendered a PR URL as `/commit/https://github.com/openclaw/openclaw/pull/76079`.
- https://github.com/openclaw/openclaw/issues/72080#issuecomment-4345196944 rendered `#23580` as `/commit/#23580`.
- https://github.com/openclaw/openclaw/issues/77514#issuecomment-4374218024 rendered `#73046` and `#727` as commit links.
- https://github.com/openclaw/openclaw/pull/71158#issuecomment-4340797017 rendered `PR #67960` as `/commit/PR #67960`.
- https://github.com/openclaw/openclaw/issues/75176#issuecomment-4354588642 rendered several `PR #...` entries as commit links.
- https://github.com/openclaw/openclaw/issues/68033#issuecomment-4353976939 rendered `PR #64713`, `PR #67642`, and `PR #67704` as commit links.

### Validation

- `pnpm run check`
- `/pre-ship --committed`: 0 MUST-FIX findings, no deferred findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ checkpoint, and status-only commits are intentionally omitted.
 
 ### Fixed
 
+- Ignored non-SHA likely-owner provenance values when rendering public commit
+  links, avoiding broken `/commit/...` URLs in review comments. Thanks @samzong.
 - Gave manual exact-item review dispatches their own concurrency group so
   targeted maintainer reviews no longer wait behind broad normal backfill runs.
 - Downgraded screenshot-only browser runtime proof so ClawSweeper no longer accepts "no visible console/CSP violation" screenshots as sufficient real behavior proof. Thanks @BunsDev.
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -3606,6 +3606,10 @@ function shortSha(sha: string): string {
   return sha.slice(0, 12);
 }
 
+function isCommitSha(value: string): boolean {
+  return /^[0-9a-f]{7,40}$/i.test(value.trim());
+}
+
 function releaseUrl(tag: string): string {
   return repoUrl(`/releases/tag/${encodeURIComponent(tag)}`);
 }
@@ -4095,7 +4099,8 @@ function likelyOwnerLine(owner: LikelyOwner): string {
   const role = owner.role.trim();
   const reason = sentence(owner.reason.trim() || "Related by repository history.");
   const commits = owner.commits
-    .filter(Boolean)
+    .map((commit) => commit.trim())
+    .filter(isCommitSha)
     .slice(0, 3)
     .map((commit) => linkedSha(commit))
     .join(", ");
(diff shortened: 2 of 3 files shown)


ds4psb-ai (commented):
## Independent code review

### Summary verdict
LGTM — small, surgical, evidence-backed fix; no blockers.

### Findings

**P0 / blocker**
- (none)

- (none)

- `src/clawsweeper.ts:3609` `isCommitSha` — the regex `/^[0-9a-f]{7,40}$/i` is correct, but the existing `reportLikelyOwners` parser at `src/clawsweeper.ts:4313-4319` already trims commits when reading them out of stored markdown. The new `.map(commit => commit.trim())` is therefore redundant for the persisted-then-rendered path, but is necessary for the freshly-decided path (object straight from Codex JSON). Worth a one-line code comment so future readers don't strip the trim on the assumption it's dead code.
- `src/clawsweeper.ts:4097-4101` — consider also dropping owners whose entire `commits` array filters to empty after the SHA gate (currently they still render with no `commits:` suffix, which is the correct/safe behavior — flagging only because the PR description focuses on URL noise and you may want symmetry with how empty `files` is handled). Not blocking.

### Test coverage
The single new test at `test/clawsweeper.test.ts:629-652` does cover the mixed-provenance case the PR body claims (one URL + one valid SHA, with whitespace padding on the SHA), and asserts both negatives (no `/commit/https:` substring) and positives (the well-formed link). Good shape. Two coverage gaps worth considering (P2):

1. A pure-noise input: `commits: ["#23580", "PR #67960"]` — should produce a line with no `commits:` suffix at all. This is the exact pattern from the issue 72080 and PR 71158 evidence URLs in the description.
2. Uppercase-hex SHA acceptance — the regex uses the `i` flag and your isCommitSha unit-tests aren't exposed; one round-trip case would lock that in.

Neither is blocking; the existing test catches the production regression class.

- **False negatives on real short SHAs**: regex floor of 7 matches GitHub's own short-SHA convention, no risk.
- **Truncated 6-char SHAs from older Codex output**: searched `src/clawsweeper.ts` for `shortSha(` callers — nothing emits a 6-char value back into a likelyOwner commits field, so the 7-char floor is safe.
- **Unicode / smart-quote variants of `PR #...`**: not produced by Codex; not in the evidence URLs.
- **Production evidence still broken**: spot-checked openclaw/openclaw#72080 (still renders `/commit/#23580` as a link) and openclaw/openclaw#71158 (still renders `/commit/PR` links) — confirms the fix isn't already obsoleted by something else.

Nice catch on the trim — many "filter Boolean" guards miss whitespace-only strings.

PR #68, opened 2026-05-12
perf: budget review prompt file patches

stainlu:
## Summary

- Bound total PR file patch text retained in ClawSweeper review prompts.
- Preserve hydrated PR file metadata and existing file-window omission markers.
- Add review telemetry for how many retained file patches were truncated.
- Include `dashboard/` in CI sparse checkout so the existing dashboard worker test can run on PRs.
- Add unit coverage for patch budgeting and omission marker preservation.

## Why

Large OpenClaw PRs can hydrate up to 80 PR files, and each file previously kept up to 2k chars of patch text. That can push roughly 160k chars of repeated diff hunks into every review prompt. This keeps the metadata breadth while bounding diff-token pressure.

The CI sparse checkout already included `test/`, but omitted tracked `dashboard/`, so `test/dashboard-worker.test.ts` failed to import `dashboard/worker.js` on GitHub. This branch includes the minimal workflow fix needed for CI parity.

## Validation

- `pnpm run build`
- `pnpm run format:check`
- `pnpm run test:unit`
- `git diff --check`

## Note

`pnpm run check` fails locally on two existing notifier repair tests under Node 22. The repo requires Node 24+, and I reproduced the same local Node 22 failures on clean `upstream/main`; this branch build, format, and unit surface are green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
           sparse-checkout: |
             .github
             config
+            dashboard
             docs
             instructions
             prompts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,9 @@ checkpoint, and status-only commits are intentionally omitted.
   instead of reserving an entire 35-70 shard lane for every planning or
   publishing background run, so saturated backlog runs keep using available
   Codex capacity.
+- Bounded the total pull request patch text kept in review prompts while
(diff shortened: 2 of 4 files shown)


steipete:
Closing this as no longer needed.

The patch-budgeting approach is useful in isolation, but we are moving the review path toward Codex inspecting the PR branch directly instead of carrying more prompt-hydration complexity in ClawSweeper. That should reduce the need for this extra truncation layer while keeping the core bot simpler.

Thanks for the focused performance work here.

PR #83, opened 2026-05-17
fix: align command status comment trust config

hxy91819:
## Problem

PR #75 fixed a real command-status targeting bug: ordinary event-item reviews and `/clawsweeper re-review` command runs share the same `clawsweeper_item` workflow, but only command-triggered re-reviews have a command status comment to update. When an ordinary PR/issue event had no `command_status_marker` or `status_comment_id`, the old parser could treat `--marker ""` as the string `"true"`, causing `update-command-status` to scan comments for `body.includes("true")` and potentially patch a human comment such as one containing `isError: true`.

That fix left two follow-up issues in the merged code:

- `update-command-status` hardcoded `clawsweeper` and `openclaw-clawsweeper` as trusted status comment authors, while the rest of the repair lane supports trusted bot overrides through `CLAWSWEEPER_TRUSTED_BOTS`.
- The direct env fallback for the status comment id read `CLAWSWEEPER_STATUS_COMMENT_ID`, but the workflow exports `STATUS_COMMENT_ID` and passes that value to `--status-comment-id`.

## Solution

This follow-up keeps PR #75's conservative behavior: `update-command-status` only updates a status comment that can be proven by exact comment id or full command marker, and the target must be authored by a trusted bot.

The changes are:

- Export the repair lane default trusted bot list from `src/repair/config.ts` so command status updates do not maintain a second copy of that policy.
- Parse `--trusted-bots` / `CLAWSWEEPER_TRUSTED_BOTS` in `src/repair/update-command-status.ts` and use `isAllowedMutationActor()` for the same `foo` / `foo[bot]` normalization used elsewhere in the repair lane.
- Align the env fallback with the workflow by reading `STATUS_COMMENT_ID` when `--status-comment-id` is not provided.
- Add regression tests for env-based status comment id parsing, custom trusted bot exact-id matching, custom trusted bot marker fallback, and the existing human-comment guard.

## Test plan

- `pnpm run build:repair && node --test test/repair/update-command-status.test.ts`
- `nvm use 24 && pnpm run check`

### DIFF
--- a/src/repair/config.ts
+++ b/src/repair/config.ts
@@ -12,7 +12,7 @@ import {
 export { DEFAULT_HEAD_PREFIX, DEFAULT_TARGET_REPO } from "./constants.js";
 
 const DEFAULT_ALLOWED_ASSOCIATIONS = ["OWNER", "MEMBER", "COLLABORATOR"];
-const DEFAULT_TRUSTED_BOTS = ["clawsweeper[bot]", "openclaw-clawsweeper[bot]"];
+export const DEFAULT_TRUSTED_BOTS = ["clawsweeper[bot]", "openclaw-clawsweeper[bot]"];
 
 export type CommentRouterConfig = {
   targetRepo: string;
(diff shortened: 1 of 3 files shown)


hxy91819:
@codex review

PR #168, opened 2026-05-22
fix(repair): sanitize non-github URLs out of result.json evidence

momothemage:
## Summary

Fixes the recurring `#<n> evidence contains non-GitHub external URL` deterministic-validator failure on automerge repairs (most recently observed on **PR #2353**, where Codex embedded a `https://vercel.com/...` deploy-preview link into `actions[].evidence`).

Until now, the strict allow-list in [`src/repair/review-results.ts`](src/repair/review-results.ts) (`evidenceHasExternalUrl`) was the **only** thing checking evidence URLs, and several writers of `result.json` had no symmetric sanitizer. Any time Codex (or even our own `failingCheckEvidence`) copied a third-party check URL — Vercel previews, Netlify, Codecov, raw.githubusercontent.com, gist.github.com, etc. — into evidence, automerge would block on a 100% deterministic gate that nobody could fix from inside the repair loop.

This PR introduces a single shared github.com allow-list and applies it on **every** `result.json` write path, while leaving the strict review gate untouched as a hard backstop.

> Note: there is already a vague entry in `CHANGELOG.md` 0.2.1 ("Fixed automerge repair evidence so third-party check detail URLs are summarized without tripping ClawSweeper's strict GitHub-only evidence validator."). That entry described the original partial fix (only `failingCheckEvidence` was hardened); this PR is the complete generalization across **all** evidence-bearing fields and writers, plus the matching `Fixed` entry.

## Why one more layer is needed

Today's flow:

```
runCodex → writes result.json → reviewResult() → if fail, repairResultIfNeeded()
                                       │
                                       └── evidenceHasExternalUrl(line) → BLOCK
```

The block is correct, but irrecoverable from the worker's perspective: Codex frequently re-emits the same external URL in the repair pass, so the loop just burns budget. The right place to fix this is at every **producer / writer** of `result.json`, not at the validator.

### New module — [`src/repair/url-safety.ts`](src/repair/url-safety.ts)

Single source of truth. Hostname allow-list = `{ "github.com" }`.

| Export | Purpose |
|---|---|
| `isGithubUrl(value)` | Strict check used by intake to reject dispatched `report_url` overrides. |
| `sanitizeCheckLink(value)` | Single-URL fields (e.g. a check `details_url`). Returns the URL if github.com, else `""`. |
| `sanitizeEvidenceText(value)` | Free-form string. Replaces every non-github `http(s)://...` token with `<external link>`. The placeholder intentionally contains no `http(s)://` so it cannot re-trigger `evidenceHasExternalUrl`. |
| `sanitizeEvidenceList(list)` | Array form, also stringifies non-string entries safely. |
| `sanitizeResultEvidence(result)` | In-place sanitize of the canonical evidence-bearing fields on `result.json` (see below). |

Fields cleaned by `sanitizeResultEvidence`:

- `actions[].evidence`
- `needs_human`
- `merge_preflight[].security_evidence`
- `merge_preflight[].comments_evidence`
- `merge_preflight[].bot_comments_evidence`
- `merge_preflight[].codex_review.evidence`

github.com URLs are preserved verbatim, so debugging is unaffected.

### Producer hardening — [`src/repair/deterministic-automerge-result.ts`](src/repair/deterministic-automerge-result.ts)

- `failingCheckEvidence` now uses the shared `sanitizeCheckLink`.
- The local `safeCheckDetails` helper is removed (dedup).
- For non-github URLs we fall back to `external check details on <hostname>` (a hostname **fragment** without `http(s)://`, so the strict gate is satisfied while the human still gets useful context).
- The full `evidence` array is wrapped in `sanitizeEvidenceList` as a final defense-in-depth.

### Writer hardening — [`src/repair/run-worker.ts`](src/repair/run-worker.ts)

Sanitize at every place `result.json` is written or modified:

| Site | Before write |
|---|---|
| Deterministic automerge result | `sanitizeResultEvidence(deterministicResult)` |
| Dry-run synthetic result | `sanitizeResultEvidence(dryResult)` |
| `writeBlockedResult(...)` | `sanitizeResultEvidence(result)` |
| After Codex direct-write | `sanitizeResultFile(resultPath)` |
| Inside `repairResultIfNeeded` (each retry) | `sanitizeResultFile(resultPath)` after the repair Codex returns |
| After `repairResultIfNeeded` returns | `sanitizeResultFile(resultPath)` |

`sanitizeResultFile` reads, parses, sanitizes, and re-writes the file. Parse failures are tolerated (treated as no-op) because malformed result.json is already handled elsewhere.

### Intake hardening — [`src/repair/commit-finding-intake.ts`](src/repair/commit-finding-intake.ts)

- `report_url` from dispatch input is now validated with `isGithubUrl`. If it does not point at `github.com`, we silently fall back to the canonical `https://github.com/<reportRepo>/blob/main/<reportPath>` default. This closes the obvious "attacker-controlled dispatch input could land an external URL in evidence" injection vector.
- The synthetic `result.json` is run through `sanitizeResultEvidence` before being written to disk.

### Tests — [`test/repair/url-safety.test.ts`](test/repair/url-safety.test.ts)

Nine `node:test` cases, all reading from the compiled `dist/repair/url-safety.js` (matching the project test convention):

1. `isGithubUrl` only accepts `github.com` (rejects `gist.github.com`, `raw.githubusercontent.com`, non-URLs, null).
2. `sanitizeCheckLink` keeps github.com URLs verbatim.
3. `sanitizeCheckLink` drops non-github URLs.
4. `sanitizeEvidenceText` replaces non-github URLs with the hostname-free placeholder.
5. Mixed-URL strings: github.com is preserved, sibling external URLs are scrubbed.
6. Idempotency.
7. `sanitizeEvidenceList` stringifies non-string entries safely.
8. `sanitizeResultEvidence` cleans `actions[].evidence` in place.
9. `sanitizeResultEvidence` cleans `needs_human`, `merge_preflight.{security,comments,bot_comments}_evidence`, and `merge_preflight.codex_review.evidence`.
10. Null/undefined safety.
11. Parity check: sanitized output never matches the same regex shape used by `evidenceHasExternalUrl`.

### CHANGELOG

Adds a precise entry to `0.2.1 - Unreleased` › `Fixed`.

## Explicit non-goals

- **`review-results.ts` is unchanged.** The strict `evidenceHasExternalUrl` gate stays as the hard backstop. If a new writer is ever added without going through `sanitizeResultEvidence`, this PR's fix degrades back to today's blocking behavior — which is the desired failure mode (visible, deterministic) rather than silent acceptance of external links in evidence.
- **No allow-list expansion.** Only `github.com` is allowed. Subdomains (`gist.`, `raw.`, `api.`) are deliberately excluded because the public review/PR comments rendered from this evidence are GitHub-only by contract.

## How to verify

1. Local: `pnpm run build:repair && node --test test/repair/url-safety.test.ts dist/repair/*.test.js` (requires Node ≥ 24 per `package.json` engines).
2. Replay PR through `repair:worker` in dry-run mode and confirm the resulting `result.json` no longer contains any `https://vercel.com/...` strings inside evidence arrays, and that `review:results` exits 0.
3. Spot-check a deterministic automerge run where one failing check's `link` is on a third-party host: the evidence line should read `Failing check: <name>:<state> (external check details on <hostname>)` instead of containing the raw URL.

## Risk

- Very low. The change is additive (new module + a sanitize call at known write points) and has no effect on already-clean evidence (github.com URLs are preserved verbatim, idempotent).
- The placeholder `<external link>` is purely informational; it deliberately does not contain a real URL, so downstream comment renderers cannot turn it into a clickable link.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ checkpoint, and status-only commits are intentionally omitted.
 ### Fixed
 
 - Restored UTF-8 emoji labels on the live dashboard after mojibake slipped into the Worker HTML template.
+- Sanitized non-`github.com` URLs out of repair worker `result.json` evidence (including `actions[].evidence`, `needs_human`, and every `merge_preflight` evidence list) before review so deploy-preview and other external links no longer trip the `evidence contains non-GitHub external URL` deterministic gate; deterministic automerge results, dry-run/blocked fallbacks, the Codex-written result, the result-repair retry, and synthetic commit-finding-intake results all share a single `src/repair/url-safety.ts` allow-list. The intake also rejects dispatched `report_url` overrides that are not on `github.com` and falls back to the canonical report path.
 - Kept scheduled target fanout covering public `steipete/*` repositories when the ClawSweeper GitHub App is not installed for that owner.
 - Reduced the shared Codex worker budget from 72 to 57 so background review, commit-review, repair, and issue-implementation lanes run about 20% fewer parallel workers.
 - Clarified re-review guidance so PR/issue authors and users with repository write access can request a fresh read-only review without a maintainer relay.
--- a/src/repair/commit-finding-intake.ts
+++ b/src/repair/commit-finding-intake.ts
@@ -20,6 +20,7 @@ import { readJsonFileIfExists as readJsonIfExists } from "./json-file.js";
 import { renderJobIntentFrontmatter } from "./job-intent.js";
 import { commitFindingPrTitle } from "./pr-title.js";
 import { escapeRegExp, slug } from "./text-utils.js";
+import { isGithubUrl, sanitizeResultEvidence } from "./url-safety.js";
 
 const args = parseArgs(process.argv.slice(2));
 const command = args._[0] ?? "prepare";
@@ -37,9 +38,9 @@ function prepare() {
     "report-path",
     stringArg("report_path", `records/${repoSlug(targetRepo)}/commits/${sha}.md`),
   );
-  const reportUrl =
-    stringArg("report-url", stringArg("report_url", "")) ||
-    `https://github.com/${reportRepo}/blob/main/${reportPath}`;
+  const defaultReportUrl = `https://github.com/${reportRepo}/blob/main/${reportPath}`;
+  const dispatchReportUrl = stringArg("report-url", stringArg("report_url", ""));
+  const reportUrl = isGithubUrl(dispatchReportUrl) ? dispatchReportUrl : defaultReportUrl;
   const active = truthy(enabled);
   const reportRead: CommitFindingReportReadResult = active
     ? readReport({ reportRepo, reportPath })
@@ -353,6 +354,7 @@ function writeSyntheticRun(context: LooseRecord) {
     `${JSON.stringify(result.fix_artifact, null, 2)}\n`,
     "utf8",
   );
+  sanitizeResultEvidence(result);
   fs.writeFileSync(
     path.join(context.runDir, "result.json"),
     `${JSON.stringify(result, null, 2)}\n`,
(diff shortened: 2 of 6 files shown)


momothemage:
@clawsweeper automerge

momothemage:
@clawsweeper approve

PR #182, opened 2026-05-24
fix: disable scheduled repair routing

guanbear:
## Summary
- stop the scheduled repair comment-router sweep so PR repair only runs from exact comment dispatches or explicit workflow dispatch
- stop scheduled repair self-heal retries; self-heal remains manually dispatchable
- update repair docs to remove the old scheduled repair variable guidance

## Verification
- pnpm run check
- pnpm run format:check
- git diff --check


guanbear:
Closing: this change targeted the wrong repository. The intended fix is for local OpenClaw scheduled tasks, not upstream ClawSweeper workflows.

PR #207, opened 2026-05-26
fix: run cluster intake daily

RomneyDa:
## Summary
- Change scheduled cluster repair intake from hourly to once per day at `8 8 * * *`.
- Remove the cluster intake cron from dashboard status and UI rendering so the cadence is not duplicated in dashboard code.
- Keep gitcrawl-store freshness docs focused on the store SHA duplicate guard while documenting the daily import limit.

## Verification
- `pnpm run build:all && node --test test/dashboard-worker.test.ts test/repair/gitcrawl-store.test.ts test/clawsweeper.test.ts`
- `pnpm run check`


RomneyDa:
@clawsweeper review

PR #208, opened 2026-05-26
fix: dispatch cluster intake after state publish

RomneyDa:
## Summary
- remove the post-publish git rebase from cluster intake dispatch so generated job files do not block dispatch
- read cluster intake dashboard markers from openclaw/clawsweeper-state@state, where intake publishes them
- add regression coverage for both the state marker location and dispatch workflow behavior

## Testing
- pnpm run check


RomneyDa:
@clawsweeper review

Issue #170, opened 2026-05-22
re-review workflow reports "Complete" but durable verdict marker not refreshed to new head SHA


solomonneas:
ClawSweeper's `@clawsweeper re-review` workflow posts a progress block that says `State: Complete` and "the durable review comment was updated, and the synced verdict was routed", but the durable review comment's verdict marker stays pinned to the SHA from the original review and never refreshes against the current PR head.

## Repro

Both observed on openclaw/openclaw#72295.

**Run 1**
- Re-review requested: 2026-05-12 16:40 UTC (https://github.com/openclaw/openclaw/pull/72295#issuecomment-4432728535)
- Workflow run: https://github.com/openclaw/clawsweeper/actions/runs/25748589643
- Progress block: State Complete, Updated 2026-05-12T16:48:19.455Z
- PR head at request: 1c826a861ac48a904cdaedd2339a11b798493022
- Durable review marker SHA after run: d2171647e1f54ae9f700b26438f757fd5dd3c916 (unchanged from original 2026-04-28 review)

**Run 2 (independent repro 10 days later)**
- Re-review requested: 2026-05-22 12:37 UTC (https://github.com/openclaw/openclaw/pull/72295#issuecomment-4518702669)
- Workflow run: https://github.com/openclaw/clawsweeper/actions/runs/26288252854
- Progress block: State Complete, Updated 2026-05-22T12:45:43.266Z
- PR head at request: 1c826a861ac48a904cdaedd2339a11b798493022
- Durable review marker SHA after run: d2171647e1f54ae9f700b26438f757fd5dd3c916 (still unchanged)

## Impact

Labels stay stuck where the original review left them (in this case `status: ⏳ waiting on author`), even when the author has pushed fixes addressing the original findings. Maintainers filter by these labels, so a stuck `waiting on author` effectively buries the PR. The author has no signal that the re-review didn't actually happen, since the progress block says it did.

## Likely cause

The progress-block update path and the durable-comment edit path appear to be decoupled. The status comment is being updated optimistically before the verdict edit succeeds, or the verdict edit silently no-ops when the new review run produces an identical verdict body, leaving the SHA marker stale.

## Suggested signal

Even if the re-review intentionally produces the same verdict text, the SHA marker in the durable comment should be bumped to the re-reviewed head SHA so authors and maintainers can tell the rerun actually happened.

Takhoffman:
@clawsweeper visualize

steipete:
[...]

PR #230, opened 2026-05-30
fix(repair): gate event re-review Complete on verdict marker freshness

openperf:
## Summary

- **Problem:** The `Mark re-review complete` step in `.github/workflows/sweep.yml` set `state="Complete"` from three step exit outcomes (`REVIEW_OUTCOME` / `PUBLISH_OUTCOME` / `ROUTE_OUTCOME` all `success`) only. It never verified that the durable `clawsweeper-verdict` marker actually advanced to the current PR head.

- **Root Cause:** `publish-event-result` calls `apply-decisions`, which syncs the durable verdict comment. On the two error branches where `apply-decisions` swallows the edit failure and exits 0 (GitHub auth rejection: `skipped_comment_auth`; locked conversation: `skipped_locked_conversation`), the publish step still reports `success` and the downstream gate had no signal that the comment was not updated. All other edit errors propagate as exceptions and already cause the publish step to fail, so the gate was already truthful in those cases.

- **Fix:** After a successful publish, `publish-event-result` calls `emitMarkerFreshness`, which fetches the live durable verdict marker body and the current PR head via `gh api` and writes `marker_fresh=1|0` to `$GITHUB_OUTPUT`. The gate reports `Complete` only on `marker_fresh=1`; on `marker_fresh=0` it reports `Incomplete` with guidance to re-run. Non-PR items (issues) vacuously pass (`marker_fresh=1`). Any error in `emitMarkerFreshness` is caught, logged, and defaults to `marker_fresh=0`; the publish step itself is never failed by this check.

- **What changed:**
  - `src/repair/comment-router-core.ts` — two pure exported helpers: `durableVerdictMarkerSha(commentBody)` parses the `sha=` attribute from the `clawsweeper-verdict` HTML marker using the existing `clawsweeperMarker` parser; `eventReviewMarkerFreshness({ itemKind, commentBody, currentHeadSha })` delegates to the existing `reviewedHeadShaBlockReason` to compare marker SHA against current head (non-PR items return `fresh: true`).
  - `src/repair/publish-event-result.ts` — adds `emitMarkerFreshness` / `verifyMarkerFreshness` / `latestVerdictCommentBody`; wired on both success paths in `publishEventResult`. Uses `ghJsonWithRetry` (4 attempts) for the PR head and `ghPagedWithRetry` (4 attempts) for the comment list; filters by trusted bot login (`clawsweeper` / `clawsweeper[bot]` / `openclaw-clawsweeper[bot]`) and `clawsweeper-verdict:` presence, sorts by `updated_at` descending.
  - `.github/workflows/sweep.yml` — `publish-event-result` step exposes `id: publish-event-result`; `Mark re-review complete` reads `${{ steps.publish-event-result.outputs.marker_fresh }}` and gates `Complete` on `marker_fresh=1`.
  - `test/repair/comment-router-core.test.ts` — 5 regression tests for the two new helpers (fresh match, stale old head, no marker, non-PR vacuous pass, SHA parse).

- **What did NOT change:**
  - `src/clawsweeper.ts` is untouched — no change to the review-generation or marker-write path.
  - No config surface, no plugin/SDK surface, no dependency changes.
  - The normal successful path is unchanged: when the marker advances the run still reports `Complete`.
  - All non-auth / non-locked edit errors already propagated as publish failures before this PR; that behavior is unchanged.

## Reproduction

1. An event re-review runs for a PR-targeted item; the review, publish, and route steps all exit 0, but the durable verdict comment edit was skipped (auth rejection or locked conversation).
2. **Before:** the gate reads only the three step outcomes and reports `state="Complete"` regardless of the comment edit outcome.
3. **After:** `publish-event-result` fetches the live verdict marker and current PR head; if they do not match, it writes `marker_fresh=0` and the gate reports `state="Incomplete"` with a detail directing operators to re-run once the head is stable.

## Real behavior proof

**Behavior addressed (gate truthfulness on swallowed comment-edit failure):** a successful publish that did not update the durable verdict comment (auth/locked) no longer falsely reports `Complete`. The helper logic is pinned by dedicated unit tests.

**Real environment tested (Node 24.9.0, node:test runner, production `comment-router-core.ts` helpers compiled via tsgo, tests importing directly from `dist/repair/comment-router-core.js`):** the five regression tests drive the actual `durableVerdictMarkerSha` and `eventReviewMarkerFreshness` production exports against real HTML marker strings and known head SHAs; no mocking of the parser or SHA comparison logic.

**Exact steps or command run after this patch:**

1. `pnpm run test:repair` (AFTER fix) — 403 pass / 0 fail
2. Remove `export` keyword from `durableVerdictMarkerSha` and `eventReviewMarkerFreshness` in `dist/repair/comment-router-core.js` (BEFORE simulation — helpers absent as on unpatched main)
3. Rerun `node --test test/repair/comment-router-core.test.ts dist/repair/comment-router-core.test.js` — import fails with `SyntaxError`, affected file reported as 1 fail
4. Restore exports, rerun — 103 pass / 0 fail

**Evidence after fix (verbatim node:test output for the 5 new regression tests):**

```
--- AFTER FIX ---
✔ durableVerdictMarkerSha extracts the head SHA from the verdict marker (0.190152ms)
✔ eventReviewMarkerFreshness treats non-PR items as fresh (0.197255ms)
✔ eventReviewMarkerFreshness passes when the marker matches the current head (0.171278ms)
✔ eventReviewMarkerFreshness flags a stale marker pinned to an old head (0.128476ms)
✔ eventReviewMarkerFreshness flags a PR with no verdict marker (2.375502ms)
  tests 103  pass 103  fail 0

--- BEFORE (exports removed — helpers absent as on unpatched main) ---
SyntaxError: The requested module '../../dist/repair/comment-router-core.js' does not
  provide an export named 'durableVerdictMarkerSha'
✖ test/repair/comment-router-core.test.ts
  tests 7  pass 6  fail 1
```

**Observed result after fix:**

- `emitMarkerFreshness` returns `marker_fresh=1` when the fetched durable comment's `sha=` attribute matches the current PR head → gate reports `Complete`.
- `emitMarkerFreshness` returns `marker_fresh=0` when the marker SHA does not match (auth/locked skip left it stale) → gate reports `Incomplete` with re-run guidance.
- Non-PR items (issues) skip the fetch entirely and always return `marker_fresh=1` (vacuously fresh; issues have no verdict marker).
- Any `gh api` error inside `emitMarkerFreshness` is caught, logged, and defaults to `marker_fresh=0`; the publish step itself does not fail.

**What was not tested:**

- End-to-end: a real re-review run where GitHub rejects the durable comment edit with auth error or locked-conversation and the gate observing `Incomplete`. The node:test suite pins the helper logic; the workflow gate change is structural and not exercised by the repair test suite.
- Full `pnpm run check` is green (all gates in CI pass on the pushed branch).

**Regression tests:**

- `durableVerdictMarkerSha extracts the head SHA from the verdict marker` — verifies the marker parser correctly extracts `sha=` from a well-formed `clawsweeper-verdict` marker body.
- `eventReviewMarkerFreshness passes when the marker matches the current head` — guards against false `Incomplete` when the marker is fresh.
- `eventReviewMarkerFreshness flags a stale marker pinned to an old head` — pins the core detection: mismatched SHA → `fresh: false`.
- `eventReviewMarkerFreshness flags a PR with no verdict marker` — pins the no-marker case (first review not yet synced).
- `eventReviewMarkerFreshness treats non-PR items as fresh` — pins the vacuous-pass for issues.

## Risk / Mitigation

- **Risk:** `emitMarkerFreshness` error (network, auth, parse) flips `marker_fresh=0`, causing a false `Incomplete` for a run that actually succeeded. **Mitigation:** the error is caught, logged to stdout, and the publish step completes normally; the operator sees the Incomplete state and can re-run — a conservative false negative rather than a silent false positive.
- **Risk:** transient GitHub propagation delay between comment edit and the freshness fetch reads a stale cached comment body. **Mitigation:** `ghPagedWithRetry` retries up to 4 times; the fetch runs after the publish push has already completed, giving propagation time. A transient miss produces `Incomplete`, which directs a re-run rather than masking the state.
- **Risk:** a post-review force-push moves the PR head between the review and the freshness check, causing a false `Incomplete` for a run that correctly reviewed the prior head. **Mitigation:** the `Incomplete` detail explicitly says "re-run once the head is stable," which is the correct operator action when the head has moved.

## Change Type

- [x] Bug fix

## Scope

- [x] Repair lane / event re-review gate (`sweep.yml`, `publish-event-result.ts`, `comment-router-core.ts`)

## Linked Issue / PR

None

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -428,6 +428,7 @@ jobs:
           REVIEW_OUTCOME: ${{ steps.review-exact-event-item.outcome }}
           PUBLISH_OUTCOME: ${{ steps.publish-event-result.outcome }}
           ROUTE_OUTCOME: ${{ steps.route-synced-verdict.outcome }}
+          MARKER_FRESH: ${{ steps.publish-event-result.outputs.marker_fresh }}
         run: |
           state="Failed"
           detail="The targeted re-review did not finish cleanly. Check the workflow run for details."
@@ -436,8 +437,13 @@ jobs:
             detail="A newer re-review for this item started before this run finished, so GitHub cancelled this older run. Check the latest ClawSweeper run for the current result."
           fi
           if [ "$REVIEW_OUTCOME" = "success" ] && [ "$PUBLISH_OUTCOME" = "success" ] && [ "$ROUTE_OUTCOME" = "success" ]; then
-            state="Complete"
-            detail="The targeted re-review finished, the durable review comment was updated, and the synced verdict was routed."
+            if [ "$MARKER_FRESH" = "1" ]; then
+              state="Complete"
+              detail="The targeted re-review finished, the durable review comment advanced to the current PR head, and the synced verdict was routed."
+            else
+              state="Incomplete"
+              detail="The targeted re-review ran, but the durable ClawSweeper verdict marker did not advance to the current PR head SHA, so labels may still reflect the previous review. Re-run the re-review once the head is stable."
+            fi
           fi
           pnpm run repair:update-command-status -- \
             --repo "$TARGET_REPO" \
--- a/src/repair/comment-router-core.ts
+++ b/src/repair/comment-router-core.ts
@@ -928,6 +928,31 @@ export function reviewedHeadShaBlockReason({
   return null;
 }
 
+export function durableVerdictMarkerSha(commentBody: JsonValue): string | null {
+  const marker = clawsweeperMarker(String(commentBody ?? ""), "verdict");
+  const sha = marker?.attrs?.sha;
+  if (typeof sha !== "string") return null;
+  const trimmed = sha.trim();
+  return trimmed && trimmed !== "unknown" ? trimmed : null;
+}
+
+export function eventReviewMarkerFreshness({
+  itemKind,
+  commentBody,
+  currentHeadSha,
+}: LooseRecord): { fresh: boolean; reason: string | null; markerSha: string | null } {
+  if (String(itemKind ?? "") !== "pull_request") {
+    return { fresh: true, reason: null, markerSha: null };
+  }
+  const markerSha = durableVerdictMarkerSha(commentBody);
+  const reason = reviewedHeadShaBlockReason({
+    expectedHeadSha: markerSha ?? "unknown",
+    currentHeadSha: currentHeadSha ?? undefined,
+    markerName: "verdict",
+  });
+  return { fresh: reason === null, reason, markerSha };
+}
+
 type AutoRepairDispatchEntry = {
   repo?: unknown;
   issue_number?: unknown;
--- a/src/repair/publish-event-result.ts
+++ b/src/repair/publish-event-result.ts
@@ -20,6 +20,14 @@ import {
   syncPublishPaths,
 } from "./git-publish.js";
 import { isJsonObject } from "./json-types.js";
+import { eventReviewMarkerFreshness } from "./comment-router-core.js";
+import { ghJsonWithRetry, ghPagedWithRetry } from "./github-cli.js";
+
+const CLAWSWEEPER_COMMENT_LOGINS = new Set([
+  "clawsweeper",
+  "clawsweeper[bot]",
+  "openclaw-clawsweeper[bot]",
+]);
 
 type ApplyAction = {
   action: string;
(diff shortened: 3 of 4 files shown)


openperf:
Closing: the fixed gap (gate reporting Complete when a durable comment edit is swallowed by an auth/locked-conversation error) is real but extremely narrow and rarely occurs in practice. The original motivation (#170) does not reproduce — live run logs confirm the marker correctly tracks the reviewed head on every re-review. Not worth the added API calls and complexity.

PR #233, opened 2026-05-30
Require routable ClawSweeper commands before fast ack

### LINKED ISSUE #232: Fast ack can queue inline @clawsweeper mentions that the router cannot parse
## Summary

A GitHub issue comment can receive the ClawSweeper fast-ack response even when the command router later cannot parse the comment as a command. This leaves a stale bot comment saying the command was queued, but no review/action run is started and the ack comment is never updated.

This is confusing for maintainers/trial maintainers because it looks like a permissions problem or a missing Actions run, when the actual issue is a parser mismatch.

## Example

On `openclaw/openclaw#87801`, this comment was posted by `brokemac79`:

```text
the closed PR 87835 was closed as already implemented by PR 87890 @clawsweeper re-review and if necessary close this issue
```

ClawSweeper replied:

```text
ClawSweeper picked this up.

Command router queued. I will update this comment with the next step.
```

But the matching `repair comment router` run did not see a routable command:

- source comment: `4583610644`
- status/ack comment: `4583610747`
- router run: https://github.com/openclaw/clawsweeper/actions/runs/26690832809
- router args included `--item-numbers 87801 --comment-ids 4583610644 --status-comment-id 4583610747 --execute`
- report showed `scanned_comments: 1`, `commands_seen: 0`, `actionable: 0`

## Likely cause

The fast-ack/webhook side uses a loose pattern that detects `@clawsweeper` anywhere after whitespace:

```ts
/(^|[ \t\r\n])@(?:clawsweeper|openclaw-clawsweeper)\b.../
```

The actual router parser is stricter. `parseCommand()` only accepts commands where the line itself is a command, for example:

```text
@clawsweeper re-review
```

or:

```text
/clawsweeper re-review
```

It does not parse prose with an inline mention in the middle of the sentence.

That means the webhook can fast-ack a comment that the router will later ignore.

## Expected behavior

Either of these would be OK, but the two layers should agree:

1. The fast-ack path should only ack comments that the router parser will accept, or
2. The router parser should support the same inline command shape that the fast-ack path accepts.

Given the risk of accidental inline mentions, option 1 seems safer: require command-like comments or command-like lines before posting the “queued” ack.

## Suggested fix

- In `src/repair/comment-webhook.ts`, reject issue comments when `parseCommand(body)` returns null, even if `COMMAND_PATTERN` matched.
- In `dashboard/worker.ts`, align the Cloudflare/dashboard fast-ack command detection with the router grammar as well, or factor the shared command-line detection so it cannot drift.
- Add tests for:
  - `@clawsweeper re-review` on its own line gets fast-acked and routed.
  - prose with inline `@clawsweeper re-review` does not get a queued ack unless the router will also route it.
  - a fast-ack comment is not left permanently saying “queued” for a non-routable command.

## Workaround

Use a command on its own line, for example:

```text
The closed PR 87835 was closed as already implemented by PR 87890.

@clawsweeper re-review
```

Any extra instruction/context should be placed after the command on following lines if the command supports it.

## Non-goals

This should not change repair, merge, or close permissions. It is only about making the fast-ack layer and command-router layer agree on what is actually routable.

brokemac79:
## Summary

Fixes openclaw/clawsweeper#232.

This makes the fast-ack paths reject comments that only mention `@clawsweeper` inline inside prose, unless the same comment is actually routable by the command parser.

Changes:

- gates the deterministic comment webhook on `parseCommand(...)` before posting the visible fast ack
- extracts the shared ClawSweeper command-line detector used by both the router parser and hosted dashboard worker
- replaces the hosted dashboard worker's loose mention regex with that shared parser-compatible detector
- keeps valid command forms working, including an own-line mention and the `@clawsweeper` then next-line `re-review` pattern used by issue authors
- adds regression coverage for inline mentions versus real command-line mentions
- normalizes a few workflow test reads so CRLF workflow files do not break unrelated assertions

## Testing

- `pnpm run build:all`
- `pnpm run lint`
- `node --test test\\repair\\comment-router-core.test.ts test\\repair\\comment-webhook.test.ts test\\dashboard-worker.test.ts`
- `node --test --test-name-pattern "sweep workflow_dispatch input count|sweep review continuations|repair workflows preserve" test\\clawsweeper.test.ts`
- `pnpm exec oxfmt --check src/repair/comment-command-text.ts src/repair/comment-router-core.ts dashboard/worker.ts`

Note: I also tried `pnpm run check` locally on Windows earlier. The relevant build/lint/regression parts pass, but the full suite still hits unrelated Windows/local-environment failures around spawning `codex` and POSIX file-mode assertions. A repository-wide `format:check` also reports existing unrelated formatting noise on this Windows checkout, while the touched files pass targeted format check.

### DIFF
--- a/dashboard/worker.ts
+++ b/dashboard/worker.ts
@@ -1,3 +1,8 @@
+import {
+  commandTextForClawSweeperFastAck,
+  isClawSweeperReReviewCommandText,
+} from "../src/repair/comment-command-text.ts";
+
 const ACTIVE_RUN_STATUSES = new Set(["queued", "in_progress", "waiting", "requested", "pending"]);
 const QUEUED_RUN_STATUSES = new Set(["queued", "waiting", "requested", "pending"]);
 type DashboardEnv = Record<string, unknown>;
@@ -23,8 +28,6 @@ const CLAWSWEEPER_REVIEW_REPO = "openclaw/clawsweeper";
 const CLAWSWEEPER_STATE_REPO = "openclaw/clawsweeper-state";
 const CLAWSWEEPER_STATE_REF = "state";
 const CLUSTER_REPAIR_INTAKE_WORKFLOW = "repair-cluster-intake.yml";
-const CLAWSWEEPER_COMMAND_PATTERN =
-  /(^|[ \t\r\n])@(?:clawsweeper|openclaw-clawsweeper)\b(?:\[bot\])?|(^|[ \t\r\n])\/(?:clawsweeper|review|re-review|rerun[ -]?review|status|explain|fix|build|implement|create[ -]?pr|fix[ -]?issue|autofix|auto[ -]?fix|automerge|auto[ -]?merge|approve|stop|autoclose)\b/i;
 const CLAWSWEEPER_ALLOWED_ASSOCIATIONS = new Set(["OWNER", "MEMBER", "COLLABORATOR"]);
 const CLAWSWEEPER_ISSUE_ITEM_ACTIONS = new Set([
   "opened",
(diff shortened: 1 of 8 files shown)


brokemac79:
@clawsweeper re-review

Updated to extract the shared command-line detector so the router parser and hosted dashboard worker no longer maintain separate command-shape regexes.

Takhoffman:
@clawsweeper automerge

PR #246, opened 2026-06-02
Gate proof nudges to 5am Central

Takhoffman:
## Summary
- add both UTC schedule candidates for 5:00 AM America/Chicago across daylight saving time
- gate scheduled proof-nudge execution to local hour 05 before checkout/token/setup work runs
- update the proof-nudge docs and workflow guard test for the timezone-aware schedule

## Real behavior proof
Local timezone proof for the two UTC schedule candidates:

```text
2026-07-01T10:00:00Z -> 05:00 CDT
2026-07-01T11:00:00Z -> 06:00 CDT
2026-12-01T10:00:00Z -> 04:00 CST
2026-12-01T11:00:00Z -> 05:00 CST
```

The workflow guard checks `TZ=America/Chicago date +%H` and only continues when the local hour is `05`, so the extra scheduled UTC event becomes a no-op in the opposite season.

## Validation
- node --test --test-name-pattern "proof nudge workflow" test/clawsweeper.test.ts
- pnpm run format:check
- node -e timezone mapping proof above

### DIFF
--- a/.github/workflows/proof-nudges.yml
+++ b/.github/workflows/proof-nudges.yml
@@ -31,6 +31,7 @@ on:
         default: ""
   schedule:
     - cron: "0 10 * * *"
+    - cron: "0 11 * * *"
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


Takhoffman:
@clawsweeper re-review

PR #254, opened 2026-06-04
[codex] Fix proof nudge schedule gate

Takhoffman:
## Summary

Fixes the proof-nudge scheduled gate so the DST cron candidate must also be running during the actual 5:00 AM America/Chicago hour before it can continue past the guard.

## Root cause

`main` has two UTC cron entries for the same intended local schedule: `0 10 * * *` for CDT and `0 11 * * *` for CST. That part is intentional because GitHub Actions schedules are UTC-only. The latest delayed-schedule fix removed the local-hour check and only matched the cron string to the current timezone abbreviation, so a delayed `0 10 * * *` run during CDT could still execute after 5 AM Central instead of skipping. The alternate cron also still appears in Actions as a short successful run because GitHub creates a run for every schedule entry even when our guard exits before checkout.

## Changes

- Restore the `local_hour == 05` guard while preserving the cron-string and CDT/CST checks.
- Update the proof-nudge workflow test to pin the local-hour guard.
- Document why there are two cron candidates and why the skipped one may still be visible in the Actions list.

## Validation

- `pnpm exec oxfmt --check .github/workflows/proof-nudges.yml docs/proof-nudges.md test/clawsweeper.test.ts`
- `pnpm run build && node --test test/clawsweeper.test.ts --test-name-pattern "proof nudge workflow is manual-first"` (Node still ran the full `test/clawsweeper.test.ts`; 359 tests passed)
- Crabbox direct AWS attempt `cbx_f4cec2927105` reached sync but was a raw unhydrated box without `pnpm`, so it did not validate repo tests. Follow-up Actions hydration was blocked by runner-registration permissions; unused lease `cbx_7fdde2040b76` was released.

## Schedule guard proof

I ran the exact shell boolean now present in `.github/workflows/proof-nudges.yml` against representative schedule inputs. This covers the intended CDT/CST pass cases plus delayed or wrong-zone skip cases:

```text
0 10 * * *   local_hour=05 local_zone=CDT => should_run=true
0 10 * * *   local_hour=06 local_zone=CDT => should_run=false
0 11 * * *   local_hour=05 local_zone=CST => should_run=true
0 11 * * *   local_hour=06 local_zone=CDT => should_run=false
0 11 * * *   local_hour=05 local_zone=CDT => should_run=false
```

The important regression case is the delayed CDT `0 10 * * *` candidate at local hour `06`, which now skips instead of continuing past the guard.

### DIFF
--- a/.github/workflows/proof-nudges.yml
+++ b/.github/workflows/proof-nudges.yml
@@ -60,15 +60,17 @@ jobs:
           PROOF_NUDGES_SCHEDULE_TZ: America/Chicago
           PROOF_NUDGES_EVENT_SCHEDULE: ${{ github.event.schedule }}
         run: |
+          local_hour="$(TZ="$PROOF_NUDGES_SCHEDULE_TZ" date +%H)"
           local_time="$(TZ="$PROOF_NUDGES_SCHEDULE_TZ" date '+%Y-%m-%d %H:%M:%S %Z')"
           local_zone="$(TZ="$PROOF_NUDGES_SCHEDULE_TZ" date +%Z)"
-          if { [ "$PROOF_NUDGES_EVENT_SCHEDULE" = "0 10 * * *" ] && [ "$local_zone" = "CDT" ]; } ||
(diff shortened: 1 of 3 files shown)


Takhoffman:
@clawsweeper re-review

PR #253, opened 2026-06-04
feat: add local-review command for offline branch review

anagnorisis2peripeteia:
## Summary

Adds a `local-review` command that reviews a local git branch without needing a GitHub PR. This enables pre-push code review: stage changes on a branch, run `local-review`, fix findings, then push and create the PR.

Self-contained: includes all prerequisites for local runs (CODEX_BIN, shell: win32, CLAWSWEEPER_CODEX_LOGIN_METHOD) applied to all 4 Codex subprocess launchers.

## How it works

The existing `review` command fetches PR metadata from GitHub (`fetchItem`, `collectItemContext`) and runs Codex against the checkout. `local-review` constructs the same `Item` and `ItemContext` from local git state:

- **Item**: repo from `git remote`, title from commit message, author from `git config`
- **ItemContext**: diff from `git diff <base>...HEAD`, changed files from `git diff --name-only`, commits from `git log`
- **PR body**: from `--pr-body-file` flag or commit message body
- **Review engine**: identical `runCodex()` call with the same prompt, schema, and decision output

No GitHub API calls are made. The decision is written to `~/.clawsweeper-local-reviews/local-review.json` and printed to stdout.

## Prerequisites included

This PR subsumes the changes from #250 (Windows spawn fix) and #251 (OAuth login method):

- `CODEX_BIN` env var for Windows `.cmd` wrapper compatibility
- `shell: true` on win32 for all Codex `spawnSync` calls
- `CLAWSWEEPER_CODEX_LOGIN_METHOD` env var for ChatGPT Pro OAuth login (centralized in `codex-env.ts` with validation)
- Applied to all 4 Codex subprocess launchers: `clawsweeper.ts` (review + assist), `commit-sweeper.ts`, `pr-close-coverage-proof.ts`

**Merge order**: if this lands first, close #250 and #251 as superseded. If either lands first, this PR needs a trivial rebase (identical changes on the same lines).

## Usage

```sh
# Basic: review current branch against main
node dist/clawsweeper.js local-review --target-dir /path/to/repo --repo owner/repo --base main

# With a PR body containing proof
node dist/clawsweeper.js local-review --target-dir /path/to/repo --repo owner/repo --pr-body-file body.md

# Auth: works with OPENAI_API_KEY or ChatGPT OAuth (codex login)
CLAWSWEEPER_CODEX_LOGIN_METHOD=chatgpt node dist/clawsweeper.js local-review ...
```

## After-fix proof

[...]

PR #251, opened 2026-06-03
feat: support CLAWSWEEPER_CODEX_LOGIN_METHOD for local OAuth runs

anagnorisis2peripeteia:
## Summary

- add `CLAWSWEEPER_CODEX_LOGIN_METHOD=api|chatgpt` for local Codex authentication selection
- keep `api` as the default for CI and production
- validate overrides fail-closed instead of passing arbitrary Codex config
- apply the login method consistently to review, assist, commit review, close-coverage proof, repair workers, and persistent app-server execution
- preserve the Windows-safe Codex launcher introduced on current main
- document the local OAuth path and add an Unreleased changelog entry

## Rewrite notes

The original patch changed only two direct review call sites. This rewrite covers every synchronous and repair Codex surface, including app-server mode, while retaining current main's Windows process launcher and timeout fixes.

Contributor credit is preserved in commit `4a6a58d109`:

`Co-authored-by: Cameron Beeley <cameron.beeley@gmail.com>`

## Validation

- `pnpm run check` on rebased head: passed
  - 500 unit tests
  - 549 repair tests
  - 1,049 combined coverage tests
  - coverage thresholds and format check passed
- focused Codex login and app-server tests: 4 passed
- final Codex autoreview on the rebased branch: no accepted/actionable findings
- contributor's Windows 11 OAuth proof remains applicable: a full local review completed with `CLAWSWEEPER_CODEX_LOGIN_METHOD=chatgpt`

## Behavior

Default behavior is unchanged. Local OAuth users may set:

```text
CLAWSWEEPER_CODEX_LOGIN_METHOD=chatgpt
```

Invalid non-empty values stop before Codex is launched. This PR does not enable automerge or alter generated-PR merge policy.

PR #250, opened 2026-06-03
fix: resolve Windows Codex spawn failures for local review runs

anagnorisis2peripeteia:
## Summary

- route review, assist, app-server, and repair Codex launches through one cross-platform process helper
- resolve and safely invoke Windows `.cmd`/`.bat` launchers, including global npm shims and paths with spaces
- terminate complete Codex process trees on timeout and wait for shutdown before returning
- add native Windows CI coverage for launcher, stdin, timeout, and app-server behavior

## Validation

- `pnpm run check`
- AWS Crabbox Windows `m7i.large`, lease `cbx_247406f20340`, run `run_9197b924d445`: build, repair build, and four focused process tests passed
- Codex autoreview on the rewritten exact head: no accepted/actionable findings

Rewritten on current `main` while preserving contributor credit.

PR #274, opened 2026-06-10
feat(local-review): add claude CLI (-p) review engine (alternative to codex)

anagnorisis2peripeteia:
## Summary

Adds a **Claude CLI (`-p`) review engine** to `local-review`, selected with `--engine claude`, as an alternative to the Codex engine. Codex stays the default, so existing runs are unchanged.

Unlike the Codex engine — which authenticates via Codex OAuth / `CLAWSWEEPER_CODEX_LOGIN_METHOD` (#251) and needs the Windows spawn handling in #250 — the Claude engine shells out to `claude -p` and relies on the CLI's own local auth, so it needs none of that login plumbing. The binary is configurable via `--claude-bin` / `CLAUDE_BIN`, so any `claude -p`-compatible CLI can be dropped in; `--claude-model` sets the model.

## ⚠️ Stacked on #253

This builds directly on **#253 (`feat/local-review-command`)** — the `local-review` command this extends — and is branched from it. Until #253 merges, this PR's diff includes #253's commit. **Merge #253 first;** the new content here is the `claude CLI (-p) review engine` + `harden(...)` commits.

## How it works

- Reuses the existing review prompt (`buildReviewPrompt`) and the `Decision` schema validation (`parseDecision`) unchanged — the engine only swaps the subprocess and the output recognition.
- The decision schema travels **in the prompt** rather than via a CLI flag: `claude`'s `--json-schema` is inline-only and the full schema (~36 KB) exceeds the platform argv limit, so it's appended to the prompt (stdin, unbounded) with an explicit output contract.
- The terminal result record is recognised the same way a claude-stream-json stream is parsed (handles both the `--output-format json` array form and the `stream-json` line form), preferring the CLI's native `structured_output` and falling back to the result text.

## Security hardening (addressing review feedback)

The security-relevant subprocess construction is extracted into `src/claude-engine.ts` and unit-tested (`test/local-review-claude.test.ts`, 7 tests):

- **Scrubbed environment.** The spawned binary is given an env via `claudeReviewEnv()` that **withholds GitHub/OpenAI/ClawSweeper-App credentials** (`GH_TOKEN`, `GITHUB_TOKEN`, `COMMIT_SWEEPER_TARGET_GH_TOKEN`, `CLAWSWEEPER_PROOF_INSPECTION_TOKEN`, `CLAWSWEEPER_APP_ID`, `CLAWSWEEPER_APP_PRIVATE_KEY`, `OPENAI_API_KEY`, `CODEX_API_KEY`) — important because `--claude-bin` is operator-configurable. The Claude CLI authenticates from its own local credentials.
- **No permission bypass; read-only tools.** Replaced `--permission-mode bypassPermissions` with an explicit `--allowedTools Read Grep Glob` allow-list — no `Edit`/`Write`/`Bash`/bypass. The dirty-checkout guard remains as defense-in-depth.

## Real behavior proof

Ran the **hardened** engine end-to-end against a one-line planted bug (`add()` rewritten to subtract). It produced a valid `Decision` that flagged the contract break and passed strict `parseDecision` validation (paths redacted):

```
$ local-review --engine claude --target-dir <repo> --base master --repo openclaw/openclaw
[local-review] repo=openclaw/openclaw base=master files=1 diff=170 chars
[local-review] title: fix: subtract in add()
[local-review] engine=claude bin=claude
[local-review] decision written to <redacted-path>
--- decision (excerpt) ---
{
  "decision": "keep_open",
  "overallCorrectness": "patch is incorrect",
  "overallConfidenceScore": 0.97,
  "summary": "This patch is incorrect: it changes `add()` to return `a - b`, inverting the function's established and named behavior. The base commit shows `add()` returning `a + b` was the original intended behavior, and nothing in the commit message, history, or repository suggests addition was ever broken. Merging this would make `add(2, 3)` return `-1` instead of `5` for every caller. A shell check of the one-line diff is enough to block this as written."
}
reviewFindings[0].title: add() now subtracts, breaking its named contract for all callers
securityReview.status: cleared
```

## Testing

- `tsgo` build clean; `node --test test/local-review-claude.test.ts` → 7/7 pass (read-only argv, scrubbed env, result-record parsing in both array and NDJSON forms, JSON extraction).
- End-to-end `local-review --engine claude` proof above.

### DIFF
--- a/src/claude-engine.ts
+++ b/src/claude-engine.ts
@@ -1,0 +1,127 @@
+// Pure, testable helpers for the Claude CLI (`-p`) review engine: building the
+// subprocess argv (read-only, no permission bypass) and recognising the
+// terminal result record in the CLI's output. Kept separate from clawsweeper.ts
+// so the security-relevant command construction and the output parsing are
+// unit-tested in isolation.
+
+// Read-only inspection tools the reviewer is permitted to use. This is an
+// explicit allow-list — NOT a permission bypass — so the spawned binary can
+// never invoke Edit/Write/Bash/etc. against the target checkout. A local review
+// has the full diff and context in the prompt, so read-only file inspection is
+// all the engine needs.
+export const CLAUDE_REVIEW_READONLY_TOOLS = ["Read", "Grep", "Glob"] as const;
+
+// Credentials the configurable Claude binary must never receive: GitHub tokens,
+// the ClawSweeper GitHub App identity/key, and OpenAI/Codex keys. The Claude CLI
+// authenticates from its own local credentials, so none of these are needed.
+export const SCRUBBED_CREDENTIAL_ENV_KEYS = [
+  "GH_TOKEN",
+  "GITHUB_TOKEN",
+  "COMMIT_SWEEPER_TARGET_GH_TOKEN",
+  "CLAWSWEEPER_PROOF_INSPECTION_TOKEN",
+  "CLAWSWEEPER_APP_ID",
+  "CLAWSWEEPER_APP_PRIVATE_KEY",
+  "OPENAI_API_KEY",
+  "CODEX_API_KEY",
+] as const;
+
+// Build the environment for the Claude review subprocess, scrubbing the
+// credentials above. The binary is operator-configurable via --claude-bin, so
+// this boundary matters even though Codex's path scrubs the same set.
+export function claudeReviewEnv(baseEnv: NodeJS.ProcessEnv = process.env): NodeJS.ProcessEnv {
+  const env: NodeJS.ProcessEnv = { ...baseEnv };
+  for (const key of SCRUBBED_CREDENTIAL_ENV_KEYS) {
+    delete env[key];
+  }
+  env.GIT_OPTIONAL_LOCKS = "0";
+  return env;
+}
+
+export type ClaudeReviewArgsOptions = {
+  proofScratchDir: string;
+  model?: string;
+};
+
+// Build the argv for a headless, read-only `claude -p` review. Deliberately
+// uses an `--allowedTools` allow-list and NO `--permission-mode bypassPermissions`,
+// so the engine cannot mutate the checkout via tools.
+export function buildClaudeReviewArgs(options: ClaudeReviewArgsOptions): string[] {
+  const args = [
+    "-p",
+    "--output-format",
+    "json",
+    "--allowedTools",
+    ...CLAUDE_REVIEW_READONLY_TOOLS,
+    "--add-dir",
+    options.proofScratchDir,
+  ];
+  if (options.model) {
+    args.push("--model", options.model);
+  }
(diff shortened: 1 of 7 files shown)


anagnorisis2peripeteia:
@clawsweeper re-review — addressed the P1s: scrubbed subprocess env (no GitHub/OpenAI/app creds reach the configurable binary), replaced `bypassPermissions` with a read-only `--allowedTools Read Grep Glob` allow-list, extracted the argv/env construction + parser into `src/claude-engine.ts` with unit tests, and added redacted real `--engine claude` terminal output to the PR body.

anagnorisis2peripeteia:
@clawsweeper re-review — addressed the P2 correctness finding (conf 0.96): reverted the shared-parser `maintainerComment` relaxation, so `parseMantisRecommendation` is strict `requireString` again for **all** engines (preserves the "recommended ⇒ needs a command" invariant). Moved the Claude-only tolerance into the engine path: `normalizeClaudeDecisionShape` defaults `maintainerComment` to `""` only when `status !== "recommended"`; a `recommended` result with no command still fails strict validation and is fixed by the bounded correction retry. Added regression coverage (strict-required for both statuses; normalize-only-when-not-recommended) and an operator-facing `--claude-bin` trust-boundary note. Live e2e: first-try pass on the common `not_recommended` case. The #253 stack/rebase and explicit trust-boundary acceptance remain maintainer gates. (ac82b03e62)

jason-allen-oneal:
Review note from local harness testing:

Validated the #274 branch locally on Linux:

```sh
pnpm install --frozen-lockfile
pnpm run build
node --test test/local-review-claude.test.ts
pnpm run test:unit
```

Results:

- Claude engine unit tests passed: 11/11
- Full unit lane passed: 432/432

I did not live-test `--engine claude` because this machine does not currently have a `claude` CLI installed. The implementation is cleanly layered on top of #253, keeps Codex as the default engine, and adds useful safeguards around read-only tool access, env scrubbing, dirty-checkout detection, schema pruning, and bounded schema correction.

Recommended merge order: land #253 first, then rebase/merge #274.

steipete:
Closing this branch because it depends on the local-review architecture closed in https://github.com/openclaw/clawsweeper/pull/253 and adds a provider credential boundary that is not safe to land.

The configurable Claude executable receives the provider API key, repository contents, and prompt; the path bypasses the current bounded process/retry infrastructure; staged review is rejected while still advertised; concurrent runs overwrite shared item-0 artifacts; and unsupported-repository fallback remains ineffective. There is also no maintainer decision to add a second review provider to this system.

A future alternate-provider design would need a provider-neutral committed-range review, unique temporary state, a fixed trusted executable boundary, minimal allowlisted environment, shared process capture/retry handling, and identical final schema validation. Thanks @anagnorisis2peripeteia for exploring this direction.

PR #276, opened 2026-06-10
fix(ci): bump setup-pi profile pin to my-pi-actions 0.1.11

valkyriweb:
0.1.10 — and every published `my-pi-*` profile before it — fails pi extension load with `Cannot find module '@valkyriweb/pi-turn-source'`: the my-pi release manifest builder skipped bundled members' workspace deps, and npm does not resolve a bundled package's missing deps. Fixed in [my-pi#154](https://github.com/valkyriweb/my-pi/pull/154); **0.1.11 is the first loadable published version**. One-line default bump; no workflow currently consumes this action (it serves the future pr-review-gate), so this is forward-prep, not a live fix.


valkyriweb:
Opened against the wrong repo (fork tooling slip) — apologies for the noise. This belongs on the valkyriweb fork.

Issue #282, opened 2026-06-12
ClawSweeper re-review fails with Codex transient_transport on moderate fork PRs (3/3 retries, then falls back to off-meta tidepool)


wangwllu:
## Symptom

`@clawsweeper re-review` on a fork PR consistently fails with `reason=transient_transport` from Codex retries, then falls back to a placeholder review:
- `decision=keep_open confidence=low action=kept_open`
- `review_summary: "Review failed before ClawSweeper could summarize the requested change."`
- routed verdict label: `🌊 off-meta tidepool / [P1] Review did not complete (retryable codex transport failure)`

The run itself reports `conclusion: success` (all GitHub Actions steps succeed), so the failure is only visible by reading step 13 "Review exact event item" logs.

## Repro

PR: openclaw/openclaw#92181 (fork PR, `+135 src / +103 tests`, 4 files, body 4243 B, diff 21460 B — moderate size).

Three independent `@clawsweeper re-review` attempts:

| Time (UTC) | Run | Outcome |
|---|---|---|
| 2026-06-11 14:45Z | https://github.com/openclaw/clawsweeper/actions/runs/27355240940 | Failed (transient_transport) |
| 2026-06-11 16:16Z | (status comment shows same pattern) | Failed (transient_transport) |
| 2026-06-12 02:10Z | https://github.com/openclaw/clawsweeper/actions/runs/27389971002 | Failed (transient_transport) |

## Run 27389971002 step 13 log excerpt

```
[review] 2026-06-12T02:11:27Z shard=0/1 selected=1 scanned_pages=0
[review] 2026-06-12T02:11:27Z shard=0/1 start #92181 (1/1)
[review] 2026-06-12T02:11:35Z shard=0/1 start-comment=existing #92181
[review] 2026-06-12T02:12:28Z codex-retry #92181 attempt=2/3 delay_ms=15000 reason=transient_transport
[review] 2026-06-12T02:13:30Z codex-retry #92181 attempt=3/3 delay_ms=30000 reason=transient_transport
[review] 2026-06-12T02:14:10Z shard=0/1 done #92181 (1/1) decision=keep_open confidence=low action=kept_open
[review] 2026-06-12T02:14:10Z shard=0/1 complete reviewed=1
Error: Codex failed for 1 item; review artifacts were written and the workflow recovery lane can requeue the planned set.
ELIFECYCLE  Command failed with exit code 1.
```

[...]

Issue #272, opened 2026-06-09
Adaptive codex timeout based on PR size and media proofs — large PRs with videos consistently time out


scotthuang:
Problem                                                                                                                                                                                                            
                                                          
The default codex_timeout_ms of 600,000ms (10 minutes) is too short for large PRs, especially those with video proofs. External contributors have no way to override this timeout — the repository_dispatch payload
 in dispatchItemReview() never includes codex_timeout_ms, so it always falls through to the hard default.
                                                                                                                                                                                                                   
Concrete example                                                                                                                                                                                                   

PR #91093 (https://github.com/openclaw/openclaw/pull/91093) in openclaw/openclaw:                                                                                                                                  
                                                          
┌───────────────┬────────────────────────────────────────┐                                                                                                                                                         
│    Metric     │                 Value                  │
├───────────────┼────────────────────────────────────────┤                                                                                                                                                         
│ Files changed │ 71                                     │
├───────────────┼────────────────────────────────────────┤                                                                                                                                                         
│ Lines added   │ +4,176                                 │                                                                                                                                                         
├───────────────┼────────────────────────────────────────┤                                                                                                                                                         
│ Commits       │ 18                                     │                                                                                                                                                         
├───────────────┼────────────────────────────────────────┤                                                                                                                                                         
│ Video proofs  │ 2 (.mov files in PR body)              │
├───────────────┼────────────────────────────────────────┤                                                                                                                                                         
│ Codex result  │ spawnSync codex ETIMEDOUT (repeatedly) │
└───────────────┴────────────────────────────────────────┘                                                                                                                                                         
                                                          
This PR is also tagged size: XL.                                                                                                                                                                                   
                                                          
Why this happens                                                                                                                                                                                                   
                                                          
1. Large diff context: 71 files × up to 2,000 chars per file patch = ~140KB of diff alone, plus PR body, 18 commits, timeline events, review comments, and the 760-line static prompt template.                    
2. Video preprocessing eats into the wall-clock budget: prepareMediaProofArtifacts() runs before codex and for each video:
  - curl download (up to 90s per video via --max-time 90)                                                                                                                                                          
  - ffprobe metadata extraction                                                                                                                                                                                    
  - ffmpeg contact sheet generation                                                                                                                                                                                
                                                                                                                                                                                                                   
Two videos could consume up to 180s just in download, plus ffmpeg processing time — all before codex even starts. This time counts against the same outer shell timeout in sweep.yml.                              
[...]

Issue #266, opened 2026-06-07
ClawSweeper command status comment edits can re-enter activity handling and churn review labels


ooiuuii:
# ClawSweeper command status comment edits can re-enter activity handling and churn review labels

## Summary

ClawSweeper command/status comments can be edited by `clawsweeper[bot]` while they still contain command markers and command-like text such as `@clawsweeper re-review`. Those bot-authored status comment edits appear to re-enter the GitHub activity lane as command-like activity, contributing to repeated `ClawSweeper Dispatch` runs, cancellations, and visible label/rating churn on a PR that had no new push and no new manual command.

This looks like the same class of self-comment inertness bug fixed for proof-nudge comments in openclaw/clawsweeper#243, but for `clawsweeper-command-status` / `clawsweeper-command-ack` / progress marker comments.

## Real incident

Observed on openclaw/openclaw#90328 (`Expose model picker agent runtimes`) on 2026-06-07.

After the final manual `@clawsweeper re-review` request, the PR still showed repeated ClawSweeper activity without new author pushes or new manual re-review comments.

Representative `ClawSweeper Dispatch` runs for #90328, all from `issue_comment` events:

```text
27098166267  2026-06-07T16:25:24Z  completed success
27098147987  2026-06-07T16:24:41Z  completed success
27098142690  2026-06-07T16:24:26Z  completed cancelled
27098047591  2026-06-07T16:20:34Z  completed cancelled
27098028747  2026-06-07T16:19:51Z  completed cancelled
27098027944  2026-06-07T16:19:49Z  completed cancelled
27098027269  2026-06-07T16:19:47Z  completed cancelled
27098026607  2026-06-07T16:19:45Z  completed cancelled
27098023797  2026-06-07T16:19:38Z  completed success
```

The same window had visible label churn by `clawsweeper[bot]`:

```text
2026-06-07T16:02:13Z  unlabeled  proof: sufficient
2026-06-07T16:02:15Z  unlabeled  proof: 📸 screenshot
2026-06-07T16:02:16Z  unlabeled  rating: 🦞 diamond lobster
2026-06-07T16:02:18Z  labeled    rating: 🌊 off-meta tidepool
2026-06-07T16:02:19Z  unlabeled  status: 👀 ready for maintainer look
2026-06-07T16:11:11Z  labeled    proof: sufficient
2026-06-07T16:11:12Z  labeled    proof: 📸 screenshot
[...]

Issue #252, opened 2026-06-03
clawsweeper re-review fails for repos with default branch of 'master' instead of 'main'


kmahone:
See this run:
https://github.com/openclaw/clawsweeper/actions/runs/26900824457/job/79352427381
>Run set -euo pipefail
>fatal: couldn't find remote ref main
>Warning: Cached target repository fetch failed; rebuilding cache.
>Cloning into bare repository 'openclaw-windows-node-cache.git'...
>fatal: Remote branch main not found in upstream origin
>Error: Process completed with exit code 128.

This was triggered from this PR:
https://github.com/openclaw/openclaw-windows-node/pull/671

The openclaw-windows-node repo has a default branch of 'master' and does not have a 'main' branch.

The initial review from clawsweeper worked correctly. But the re-review requests failed.

steipete:
Fixed by #269, merged as 985df9447491662f8738e1d1a87828ce8477844d.

The target repository's default branch is now preserved through planning, re-review dispatch, normal sweep continuation, and failed-shard recovery. Full `pnpm check`, focused regression tests, and a live non-default-branch target proof passed.

PR #283, opened 2026-06-12
Recover transport-exhausted re-reviews with a bounded lower-effort fallback (#282)

yetval:
## Problem

`@clawsweeper re-review` on PR openclaw/openclaw#92181 consistently exhausts the Codex transient-transport retry budget (`attempt 2/3`, `3/3`, `reason=transient_transport`) and then falls back to a low-confidence placeholder:

- `decision=keep_open confidence=low action=kept_open`
- routed verdict: `off-meta tidepool / [P1] Review did not complete (retryable codex transport failure)`

The high-reasoning-effort path produces long thinking streams; when the backend/proxy closes the stream early, every same-effort retry hits the same failure and the contributor has no path forward. See #282.

## Fix

`runCodex` now, after the transient-transport retries are exhausted, runs **one bounded lower-reasoning-effort fallback pass** when enough of the timeout budget remains:

- Effort step-down: `high`/`medium` -> `low`, `low` -> `minimal` (`minimal`/unknown -> no fallback).
- The fallback runs a single attempt and shares the original `--codex-timeout-ms` budget, so a re-review never becomes a stalled job.
- A successful fallback returns a real review, with:
  - degraded mode disclosed in the summary,
  - the original transport diagnostic preserved as evidence (`degraded review mode`, `original codex transport failure`),
  - `high` confidence capped to `medium`, so a degraded review can never trigger an auto-close (close actions require `confidence === "high"`).
- If the fallback also fails, the thrown `CodexReviewError` preserves the original transport classification, so the placeholder review still reads as a transport failure (capacity/network) rather than whatever the fallback surfaced last.

The minimum remaining-time budget for attempting the fallback is configurable via `CLAWSWEEPER_CODEX_FALLBACK_MIN_BUDGET_MS` (default `120000`).

## Real behavior proof

Driving the real `runCodex` code path end-to-end through the built `dist/` with a stubbed `codex` that reproduces #282 (stream closes at `high` effort, succeeds at lower effort), `CLAWSWEEPER_CODEX_REVIEW_ATTEMPTS=3`:

### BEFORE (origin/main @ 2e4282b, unpatched) — reproduces #282
```
[review] codex-retry #92181 attempt=2/3 delay_ms=5 reason=transient_transport
[review] codex-retry #92181 attempt=3/3 delay_ms=10 reason=transient_transport
reasoning-efforts-seen: high, high, high
codex-invocations: 3
{
  "threw": true,
  "errorHead": "Codex review failed for #92181 with exit 1.",
  "placeholderDecision": "keep_open",
  "placeholderConfidence": "low",
  "placeholderSummary": "Codex review failed: retryable codex transport failure (network) (exit 1).",
  "placeholderChangeSummary": "Review failed before ClawSweeper could summarize the requested change."
}
```

### AFTER (patched) — bounded lower-effort fallback
```
[review] codex-retry #92181 attempt=2/3 delay_ms=5 reason=transient_transport
[review] codex-retry #92181 attempt=3/3 delay_ms=10 reason=transient_transport
[review] codex-fallback #92181 reason=transient_transport from_effort=high to_effort=low remaining_ms=599702
reasoning-efforts-seen: high, high, high, low
codex-invocations: 4
{
  "threw": false,
  "decision": "close",
  "confidence": "medium",
  "summaryHead": "Degraded review: Codex hit a retryable codex transport failure (network) at high reasoning effort, so ClawSweeper completed this review with...",
  "changeSummary": "Adds the retry-classification helper plus regression coverage.",
  "evidenceLabels": ["degraded review mode", "original codex transport failure", "implementation"]
}
```

The unpatched build dead-ends at the placeholder; the patched build runs exactly one extra `low`-effort pass, returns a real verdict, discloses degraded mode, and caps the fixture's `high` confidence to `medium`.

## Tests

Added to `test/clawsweeper.test.ts` (all pass under `node --test ... --test-name-pattern 'Codex|transient|retry|fallback|failure'`):

- `lowerCodexReasoningEffort steps down one tier and stops at minimal`
- `runCodex completes via a lower-effort fallback after transport exhaustion`
- `runCodex keeps the transport classification when the fallback also fails`
- `runCodex skips the lower-effort fallback when the time budget is too small`

Verified locally: `pnpm run build`, `pnpm run lint:src`, `pnpm run lint:scripts`, `pnpm run check:active-surface`, `pnpm run check:limits`, `pnpm run format:check`, and the targeted test pattern (399 pass / 0 fail).

## Related issues (triage cross-reference)

`transient_transport` has no earlier hits in the tracker, so this failure mode is new. Distinct-but-adjacent issues:

- **#272** (Adaptive codex timeout for large/video PRs): same Codex-call seam, but #272 is about actually *hitting* the 10-minute `--codex-timeout-ms` on large/video PRs. Here every attempt returns `transient_transport` after roughly 2:46 of wall-clock and never reaches the timeout; the PR is moderate (~21 KB diff, no video proof), so #272's size-based mitigations would not trigger. This fix is self-contained in `runCodex` and does not depend on #272's dispatch-payload work; if #272 surfaces `codex_*` knobs in `dispatchItemReview()`, an opt-in flag for this fallback could live there later, but it is not required.
- **#170** (re-review reports "Complete" but durable marker not refreshed): same `re-review` workflow surface, different leg (verdict-sync path vs. this review-execution path).
- **#252** (re-review fails on `master` default branch): sibling re-review failure category, unrelated to Codex transport (repo fetch).
- **#266** (command-status edits re-enter activity handling and churn labels): text-search collision on "off-meta tidepool" only; underlying bug is label churn, not review failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ checkpoint, and status-only commits are intentionally omitted.
 - Compacted completed ClawSweeper-generated replacement branches to one reviewed commit before publication, removing transient checkpoint and review-repair noise while preserving contributor branch history.
 - Skip optional ClawSweeper label additions when an issue or pull request already has GitHub's 100-label maximum, so one saturated item cannot abort a comment-sync batch.
 - Served stale dashboard status immediately while coalescing a background refresh, bounded job-detail fanout, and cached and parallelized historical GitHub lookups to reduce cold-load latency, diagnostic timeouts, and API usage.
+- Recover transport-exhausted reviews with one bounded lower-effort fallback while preserving the original failure classification when recovery also fails. Thanks @yetval. (#283)
 - Preserved records written by concurrent workers during generated-state publish races while retaining deliberate item-to-closed moves and plan cleanup.
 - Raised and unified Codex review timeouts at 20 minutes, including exact event reviews, so high-context reviews do not fall back at the previous 10-minute ceiling.
 - Deferred workflow utility CLI execution until module initialization completes, preventing apply preselection from crashing on close-action constants.
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -1040,6 +1040,7 @@ const DEFAULT_CODEX_MODEL = PUBLIC_CODEX_MODEL;
 const DEFAULT_REASONING_EFFORT = "high";
 const DEFAULT_SERVICE_TIER = "";
 const DEFAULT_REVIEW_CODEX_TIMEOUT_MS = 1_200_000;
+const DEFAULT_CODEX_FALLBACK_MIN_BUDGET_MS = 120_000;
 const REVIEW_POLICY_VERSION = "2026-05-30-policy-v19";
 const REVIEW_ITEM_PROMPT_PATH = join(ROOT, "prompts", "review-item.md");
 const CLAWSWEEPER_DECISION_SCHEMA_PATH = join(ROOT, "schema", "clawsweeper-decision.schema.json");
@@ -6423,6 +6424,61 @@ function codexFailureReason(detail: string, errorCode?: string | null): string {
   return "codex execution failed";
 }
 
+function codexFallbackMinBudgetMs(): number {
+  const configured = Number(process.env.CLAWSWEEPER_CODEX_FALLBACK_MIN_BUDGET_MS?.trim());
+  return Number.isFinite(configured) && configured > 0
+    ? configured
+    : DEFAULT_CODEX_FALLBACK_MIN_BUDGET_MS;
+}
+
+export function lowerCodexReasoningEffort(effort: string): string | null {
+  switch (effort.trim().toLowerCase()) {
+    case "high":
+    case "medium":
+      return "low";
+    case "low":
+      return "minimal";
+    default:
+      return null;
+  }
+}
+
+function annotateDegradedReview(
+  decision: Decision,
+  options: { item: Item; fromEffort: string; toEffort: string; transportError: CodexReviewError },
+): Decision {
+  const reason = codexFailureReason(
+    `${options.transportError.message}\n${options.transportError.stderr}\n${options.transportError.stdout}`,
+  );
+  const disclosure =
+    `Degraded review: Codex hit a ${reason} at ${options.fromEffort} reasoning effort, so ClawSweeper ` +
+    `completed this review with a single lower-effort (${options.toEffort}) fallback pass. ` +
+    `Treat the verdict as best-effort and re-run for a full-fidelity review once transport recovers.`;
+  const transportDetail =
+    trimMiddle(
+      redactInternalCodexModel(
+        `${options.transportError.message}\n${options.transportError.stderr}`.trim(),
+      ),
+      2000,
+    ) || "No transport detail captured.";
+  return {
+    ...decision,
+    confidence: decision.confidence === "high" ? "medium" : decision.confidence,
+    summary: `${disclosure}\n\n${decision.summary}`,
+    evidence: [
+      evidenceEntry({
+        label: "degraded review mode",
+        detail: `${options.fromEffort} → ${options.toEffort} reasoning effort fallback after ${reason}.`,
+      }),
+      evidenceEntry({
+        label: "original codex transport failure",
+        detail: transportDetail,
+      }),
+      ...decision.evidence,
+    ],
+  };
+}
+
 function codexFailureDecision(
   status: number | null,
   detail: string,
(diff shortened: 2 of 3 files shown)


wangwllu:
Adding two pieces of supplementary evidence I gathered while investigating #282 from the affected PR's side, in case they're useful for the maintainer review of this fallback approach.

### 1. The failure was Codex-classified as `capacity`, not a generic transport flake

The clawsweeper review comment on `openclaw/openclaw#92181` already shows:

> **failure reason:** retryable codex transport failure (capacity)

This is the capacity branch in `codexFailureReason()` at `src/clawsweeper.ts:6357` (matches `rate limit reached / TPM / RPM / 429 / overloaded / temporarily unavailable / try again in Ns`). Step-down to lower reasoning effort is the right shape for that class — capacity rejection is about per-request token weight, which is exactly what `high → low` reduces.

### 2. The rejection was PR-specific, not a fleet-wide capacity event

I checked the 14:35Z-15:00Z window for `Review event item` runs in clawsweeper. Aside from `#92181`, every other event-item run in that window completed successfully on the same backend:

| Time (UTC) | Run | Result |
|---|---|---|
| 14:48:25 | `Review event item openclaw/openclaw#92181` | failed (this issue) |
| 14:50:47 | `Review event item steipete/CodexBar#1442` | success |
| 14:53:31 | `Review event item steipete/CodexBar#1436` | success |
| 14:57:34 | `Review event item steipete/CodexBar#1433` | success |
| 14:57:41 | `Review event item openclaw/openclaw#91632` | success (same target_repo) |
| 14:58:52 | `Review event item openclaw/openclaw#92196` | success (same target_repo) |
| 14:59:49 | `Review event item steipete/CodexBar#1434` | success |

So the rejection wasn't a transient global capacity dip — `steipete/CodexBar#1442` succeeded ~2 minutes later, and `openclaw/openclaw#91632` (same target repo) succeeded ~9 minutes later. Each `#92181` retry sent the same large prompt and got rejected at admission, which means **a longer in-process backoff alone would not have helped**; only reducing per-request token weight does. That's a positive corroboration that this PR's lower-effort fallback (rather than just longer backoff) is the right minimal fix.

### Timing detail

For completeness, the three exhausted attempts on [run 27355240940](https://github.com/openclaw/clawsweeper/actions/runs/27355240940) died this fast:

```
attempt 1: 14:48:25 → 14:48:37   12s
attempt 2: 14:48:52 → 14:49:23   31s
attempt 3: 14:49:53 → 14:50:04   11s
```

Each died well below the 600s `--codex-timeout-ms` budget and before model thinking — so the original timeout/large-reasoning hypothesis is ruled out, and the budget-share-with-fallback approach in this PR fits cleanly within the unused remainder.

(Posted #282 evidence at openclaw/clawsweeper#282 (comment 4692572669) before noticing this PR was already up — flagging here so maintainers don't have to cross-reference. Happy to help test once merged on `openclaw/openclaw#92181`.)

yetval:
Context from #282: the OP narrowed the root cause to a **capacity** (TPM admission) rejection, not a network flake or timeout. The three retries each died in 11-31s, far under the 600s budget, and other reviews on the same backend in the same window succeeded, so it is a per-request token-cost problem specific to that PR, not a fleet-wide dip. Their conclusion: backoff alone cannot help, since each retry resends the same oversized request; only reducing per-request token cost does.

That validates the mechanism in this PR. The fallback here is exactly a per-request token-cost reduction (`high -> low` reasoning effort), and it already triggers on the classified capacity case (`isRetryableCodexTransportError` matches `rate limit reached / TPM / RPM / 429 / overloaded`).

One design choice worth confirming, since #282 also floats putting the downgrade in the `retry-failed-reviews` cron lane (#260):

- This PR does the downgrade **in-process**, right after the retries exhaust, so the same run recovers and the contributor is unblocked immediately.
- The cron-lane variant keeps first-pass at `high` and only the daily recovery lane pays the lower cost, so the PR stays blocked until the next tick (08:13Z / 20:13Z).

The in-process path is strictly faster for the contributor, and the two are not mutually exclusive (the cron lane can still re-attempt at lower effort if an in-process fallback also fails). If maintainers prefer to scope the downgrade to the recovery lane only, this is a small change to make. Flagging here to avoid a duplicate PR for the same fix.

ooiuuii:
Adding another affected-case data point: I'm seeing the same failure mode on openclaw/openclaw#92251 and openclaw/openclaw#90610. The ClawSweeper re-review path hit transient_transport / Codex transport exhaustion, then the durable review/labels ended up as off-meta rather than reflecting the PR proof/status. So this does not seem isolated to #92181.

#283's bounded lower-effort fallback looks like the right shape to unblock these re-reviews without adding unbounded retry behavior.

PR #307, opened 2026-06-16
Separate human-only PR intake signals

Jhacarreiro:
## Summary
- classify ClawSweeper status/rating/merge-risk labels as metadata signals rather than automatic repair blockers
- emit `requires_human` results for PRs that only have human/workflow metadata and no repairable blocker
- avoid creating automatic repair jobs for comment-only or label-only cases when `proof: sufficient` is already present
- add an optional watch-silence fingerprint config so known watchlisted PRs stay quiet until their stable state changes

## Rationale
Repair-only intake should enqueue jobs only when there is an objective repairable blocker, such as failed checks, dirty/blocked merge state, changes requested, or unresolved actionable review threads. Stale labels and human workflow status should be reported separately for maintainer attention instead of creating repair work.

## Verification
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm run build:repair`
- `node --test test/repair/pr-repair-intake.test.ts`

### DIFF
--- a/src/repair/pr-repair-intake.ts
+++ b/src/repair/pr-repair-intake.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 import fs from "node:fs";
 import path from "node:path";
+import crypto from "node:crypto";
 import type { JsonValue, LooseRecord } from "./json-types.js";
 import { ghJson } from "./github-cli.js";
 import { renderJobIntentFrontmatter } from "./job-intent.js";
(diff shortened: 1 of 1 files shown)


steipete:
Closing rather than repairing this branch. It combines signal classification with watch-silence state, has no regression coverage for either path, and is superseded operationally by narrower intake work. The useful classification idea can return as a focused, tested change. Thank you for the proposal.

Jhacarreiro:
Closing this branch for now instead of trying to prove it with local operational runs.

If the human-only intake separation is still needed, it should come back as a small, model-agnostic PR with clean checks and upstream-clean tests/proof. Any adapter or non-Codex-specific behavior must not be part of ClawSweeper.

PR #312, opened 2026-06-17
fix(review): requeue capacity-blocked exact reviews

### LINKED ISSUE #310: Exact-review capacity queue times out before issue/PR reviews run
## Summary

Exact issue/PR review runs are currently capped by the exact-review capacity semaphore, but overflow work waits inside already-started GitHub Actions runs and then fails after the 20 minute capacity timeout. Under burst load this leaves issues/PRs without a ClawSweeper review comment or labels.

This is separate from #304. PR #304 addresses the case where Codex already produced a decision but `decision.rootCauseCluster` made the decision schema-invalid. This issue covers runs that never reach the review/comment/label stage because they time out or stall while waiting for exact-review capacity.

## Evidence

Current main has an exact-review cap:

- `config/automation-limits.json` sets `lanes.exact_review.max_concurrent` to `4`.
- `.github/workflows/sweep.yml` runs `pnpm run repair:codex-capacity` with `--timeout-ms 1200000`.
- `src/repair/codex-capacity.ts` admits only runs whose rank is inside the configured exact-review limit.

Live behavior observed on June 17, 2026:

- 75 active exact-review runs were present.
- Recent `Review event item ...` failures lasted around 20-22 minutes.
- Example: https://github.com/openclaw/clawsweeper/actions/runs/27673696915 for `openclaw/openclaw#94004` failed at the capacity step with:

```text
timed out waiting for Codex capacity: exact rank 6/37, 4 exact-review slots configured
```

A sampled admitted run waited about 17 minutes at `Wait for exact-review Codex capacity`, then reached Codex and completed review. Later-ranked runs remained stuck at the capacity step.

Additional live evidence from the re-review of #311:

- Run https://github.com/openclaw/clawsweeper/actions/runs/27678181750/job/81858579747 entered `Wait for exact-review Codex capacity`.
- The capacity helper repeatedly logged normal rank updates, then hit GitHub API installation rate limiting while polling Actions state:

```text
Could not verify Codex capacity: Command failed: gh api --method GET repos/openclaw/clawsweeper/actions/workflows/sweep.yml/runs?per_page=100&event=repository_dispatch&status=in_progress&page=1 --jq .workflow_runs
API rate limit exceeded for installation ... (HTTP 403)
```

- A live snapshot at the same time showed this run ranked 4 of 53 exact-review waiters with an exact limit of 4. In other words, it should have been admissible, but the semaphore could not reliably verify capacity because its polling hit the GitHub App installation rate limit.

That points to a second failure mode inside the same capacity gate: many waiting runs can create a polling stampede against the Actions API. Once the app installation is rate-limited, even an otherwise-admissible run can remain stuck in the capacity wait step.

## Impact

When the queue is deeper than the 20 minute wait window, ClawSweeper fails before it can publish a review comment or apply advisory labels. From the target repository this looks like ClawSweeper never ran.

The API-rate-limit behavior makes the queue less reliable under exactly the conditions where it is needed most: high burst load. It can prevent slot admission even when a run is within the configured `4` exact-review slots.

## Expected behavior

Capacity overflow should not be terminal or lossy. One of these should happen:

- exact-review work is durably queued before a worker run starts;
- capacity timeouts automatically requeue/backfill the same item;
- the timeout is treated as a retryable scheduler condition rather than a failed review;
- the exact-review lane has enough admission/backpressure to avoid accumulating doomed waiters; or
- capacity admission uses a shared/low-rate signal instead of every waiting run polling the Actions API independently.

The fix should include visible telemetry so maintainers can distinguish `waiting for capacity`, `reviewing`, `timed out but requeued`, `rate-limited while waiting`, and `permanently failed`.

brokemac79:
## Summary
- Reduce exact-review capacity polling to running workflow jobs only, with a lower polling cadence to avoid GitHub App rate-limit stampedes during bursts.
- Add a bounded capacity retry path before Codex starts: timed-out or rate-limited exact-review waiters mark the command as `Retry scheduled` and redispatch the same item once by default.
- Use a separate ClawSweeper App dispatch token for the retry path, so recovery does not depend on the same Actions token that may have been exhausted by capacity polling.
- Keep retry metadata nested under one `client_payload.capacity_retry` key so repository dispatches stay within GitHub's 10-property payload limit.
- Document the exact-review queue behavior and add focused unit/workflow assertions.

Fixes #310
Refs #304

## Verification
- `pnpm install --frozen-lockfile`
- `pnpm run build:all`
- `node --test test\repair\codex-capacity.test.ts`
- `node --test --test-name-pattern "sweep workflow admits exact event reviews" test\clawsweeper.test.ts`
- `pnpm run format:check`
- `pnpm run check:limits`
- `pnpm run lint:repair`
- `git diff --check` (CRLF warning only)
- Custom Node workflow assertion for the capacity retry/re-dispatch token path
- `codex -c service_tier=fast review --base origin/main` found and I fixed the retry-token issue; the final rerun reported no actionable correctness issues.

## Notes
- `node --test test\clawsweeper.test.ts --test-name-pattern "exact-review|capacity|event review|workflow"` is currently blocked in this Windows checkout by unrelated baseline failures around `/usr/bin/git`, `/bin/bash`, local Codex config, and unrelated assertions, so the workflow behavior is covered by focused assertions instead.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -285,6 +285,17 @@ jobs:
           permission-issues: write
           permission-pull-requests: write
 
+      - name: Create ClawSweeper dispatch token
+        id: clawsweeper-dispatch-token
+        continue-on-error: true
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        with:
+          client-id: ${{ env.CLAWSWEEPER_APP_CLIENT_ID }}
+          private-key: ${{ secrets.CLAWSWEEPER_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: ${{ github.event.repository.name }}
+          permission-contents: write
+
       - name: React to target item review start
         continue-on-error: true
         env:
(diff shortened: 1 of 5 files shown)


brokemac79:
@clawsweeper review

brokemac79:
@clawsweeper review

PR #331, opened 2026-06-19
Use public inventory fallback for local fanout

Jhacarreiro:
## Summary

Use the public inventory token fallback for local target fanout runs when an owner-specific inventory token is absent.

GitHub Actions still skips owners without minted inventory tokens. Outside Actions, local operators can safely use the public inventory env path instead of forcing empty `GH_TOKEN`/`GITHUB_TOKEN` values.

## Changes

- Return `publicInventoryEnv()` for non-Actions runs without owner-specific inventory tokens.
- Add a focused regression test that preserves the Actions skip behavior while asserting the local fallback path.

## Validation

- `corepack pnpm run build:repair`
- `node --test test/repair/target-fanout.test.ts`

### DIFF
--- a/src/repair/target-fanout.ts
+++ b/src/repair/target-fanout.ts
@@ -304,7 +304,7 @@ function inventoryEnv(owner: string): NodeJS.ProcessEnv | null {
   if (token === PUBLIC_INVENTORY_TOKEN) return publicInventoryEnv();
   if (token) return { GH_TOKEN: token, GITHUB_TOKEN: token };
   if (process.env.GITHUB_ACTIONS === "true") return null;
-  return { GH_TOKEN: "", GITHUB_TOKEN: "" };
+  return publicInventoryEnv();
 }
 
 function publicInventoryEnv(): NodeJS.ProcessEnv {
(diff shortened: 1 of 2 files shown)


Jhacarreiro:
## Validation evidence

Local validation run against this branch:

```text
corepack pnpm run build:repair
node --test test/repair/target-fanout.test.ts
```

Result:

```text
build:repair: passed
target-fanout.test.ts: 7/7 tests passed
```

Behavior covered by the new regression test:

```text
target fanout falls back to public inventory env outside Actions
```

The test verifies that non-Actions runs without an owner-specific inventory token use:

```text
return publicInventoryEnv();
```

while preserving the existing GitHub Actions behavior:

```text
if (process.env.GITHUB_ACTIONS === "true") return null;
```

This keeps Actions-safe skip behavior intact and only improves local/operator fanout behavior.

steipete:
Landed after maintainer verification.

Proof:
- Structured Codex autoreview against `origin/main`: clean, no accepted/actionable findings.
- `pnpm run build:repair && node --test test/repair/target-fanout.test.ts`: 7 passed, 0 failed.
- Real local inventory run with Actions and all inventory/dispatch/GitHub token environment variables unset successfully listed 68 eligible public `openclaw` repositories.
- GitHub CI and CodeQL were green on the reviewed head.

Landed commit: `826996f3ed496d73557e7ec2cb12801bd0fe5644`

PR #336, opened 2026-06-19
Enforce repair contract before checkpointing

Jhacarreiro:
## Summary

Enforce the repair contract before creating checkpoint commits.

The executor now checks the real changed-file set against the fix artifact's likely files / validation-derived contract before checkpointing. This catches repairs that touch no relevant files or only unrelated lockfiles.

## Scope

Model-agnostic ClawSweeper executor behavior only. No model backend, OpenAI-compatible, adapter, provider-specific code, or review/result normalization.

## Validation evidence

```text
corepack pnpm run build:repair
node --test test/repair/execute-fix-artifact-source.test.ts
```

Result:

```text
build:repair: passed
execute-fix-artifact-source.test.ts: 7/7 tests passed
```

The new regression covers:

```text
repair contract enforces likely file touch before checkpointing
```

It also verifies the git porcelain path parser uses `line.slice(3)` instead of trimming before offset parsing.

### DIFF
--- a/src/repair/execute-fix-artifact.ts
+++ b/src/repair/execute-fix-artifact.ts
@@ -1971,11 +1971,20 @@ function editValidatePrepareMerge({
   if (shouldRunCodexEdit) {
     for (let attempt = 1; attempt <= maxEditAttempts; attempt += 1) {
       const headBeforeAttempt = currentHead(targetDir);
+      const effectiveFallbackReason =
+        rebaseResult?.status === "conflicts"
+          ? [
+              "Resolve only the current rebase conflicts before doing any feature repair.",
+              "Do not broaden the patch while conflict markers remain.",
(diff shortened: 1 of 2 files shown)


steipete:
Closing as submitted. The branch fails `pnpm check`, duplicates conflict-prompt behavior already landed in https://github.com/openclaw/clawsweeper/pull/332, and guards only the first of four checkpoint paths, so later review/validation checkpoints bypass the proposed contract. The fail-closed boundary also has no real worker proof and could block valid repairs when `likely_files` is incomplete.

A future version should start from current `main`, centralize every checkpoint behind one contract-enforcing helper, exclude the already-landed conflict work, and include focused bypass tests plus a real repair-worker run. Thanks for exploring the contract boundary.

PR #350, opened 2026-06-23
perf(scheduler): raise ClawSweeper worker budget

vincentkoc:
## What Problem This Solves

ClawSweeper background review can fall hours behind while the live fleet is healthy because the checked-in worker budget and derived lane caps are too conservative for the current queue.

## Why This Change Was Made

Raise the single global budget from 32 to 48 and regenerate the derived workflow, dashboard, and operator-document values. The existing reserve and priority-lane rules stay unchanged.

## User Impact

Normal review rises to 33 shards, hot intake to 16, commit review pages to 2, and repair defaults to 19 while preserving the existing reserve policy.

## Evidence

- `pnpm run check:limits` passes.
- Live dashboard sample showed 24 active Codex jobs at 75% of the old 32-slot budget with zero recent failures.
- Main still reports `workers.max=32`; this branch carries the intended `48` budget.

AI-assisted: yes.


steipete:
Closing as superseded by the current scheduler capacity model.

Current `main` sets `workers.max` to 128 in `config/automation-limits.json`, landed through https://github.com/openclaw/clawsweeper/pull/392. This draft proposes the older 48-worker regime, conflicts across the generated capacity surfaces, and would reduce the current worker budget if revived as written.

Thanks for the capacity work and the operational groundwork here.

PR #370, opened 2026-06-25
feat: opt-in cadence to skip re-reviewing unchanged pull requests

saariuslystoned:
We run ClawSweeper's review brain on a metered subscription and noticed open pull requests get re-reviewed about once a day even when nothing changed since the last review, reproducing an identical verdict and re-spending review budget each time.

reviewCadenceMs already drops to hourly when there is activity since the last review and uses the daily hot-window cadence for new pull requests, but a pull request past its hot window with no activity still falls to the daily branch. This adds an optional CLAWSWEEPER_STALE_PULL_REQUEST_REVIEW_DAYS that extends only that branch (for example 7 for weekly). Unset, non-numeric, or sub-daily values keep the current daily behavior, and the hourly-activity and daily hot-window branches are untouched, so changed and brand-new pull requests are still reviewed promptly. Behavior is unchanged unless an operator opts in.

- src: stalePullRequestReviewCadenceMs() helper, read in reviewCadenceMs
- test: scheduler-policy cadence cases (default daily, extended weekly, invalid/sub-daily fallback) through the exported shouldReviewItem
- docs: scheduler.md cadence section

Release note (CHANGELOG.md left untouched -- it is release-owned per AGENTS.md): adds optional CLAWSWEEPER_STALE_PULL_REQUEST_REVIEW_DAYS to extend the re-review cadence for open PRs past their hot window with no activity since last review.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -1092,6 +1092,7 @@ jobs:
           MAX_PAGES: ${{ steps.mode.outputs.max_pages }}
           MIN_ACTIVE_SHARDS: ${{ steps.mode.outputs.min_active_shards }}
           MIN_BACKFILL_REVIEW_AGE_MINUTES: ${{ steps.mode.outputs.min_backfill_review_age_minutes }}
+          CLAWSWEEPER_STALE_PULL_REQUEST_REVIEW_DAYS: ${{ vars.CLAWSWEEPER_STALE_PULL_REQUEST_REVIEW_DAYS || '' }}
           TARGET_REPO: ${{ steps.target.outputs.target_repo }}
           GH_TOKEN: ${{ steps.target-read-token.outputs.token || github.token }}
         run: |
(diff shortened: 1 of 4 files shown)


saariuslystoned:
Thanks for the review! Addressed the workflow-wiring finding: CLAWSWEEPER_STALE_PULL_REQUEST_REVIEW_DAYS is now wired into .github/workflows/sweep.yml (the scheduled planner select step + the
  dashboard audit step), and cadenceBucketForReview reuses the same helper so the dashboard matches the planner. Docs updated; per AGENTS.md both src/clawsweeper.ts and sweep.yml are updated. Default
  behavior unchanged unless the repo variable is set.
  
  Real behavior proof (built dist/, Node 24):
  
  node --test test/scheduler-policy.test.ts → 13 tests, pass 13, fail 0.
  
  Cadence flip via shouldReviewItem from dist/, stale PR with no activity:
  unset (default daily)   reviewed ~26h ago -> due: true     # unchanged
  =7 (weekly)             reviewed ~26h ago -> due: false    # skipped, not re-reviewed
  =7 (weekly)             reviewed ~8d  ago -> due: true      # due after the window
  pnpm run check passes locally.
  
  @clawsweeper re-review

steipete:
Closing this stale branch rather than carrying its known planner bypass forward.

The review found that active-floor backfill can still select an unchanged pull request before the configured stale-review cadence expires, so the proposed setting does not reliably deliver its advertised budget reduction. The branch now conflicts with current scheduler policy, cannot be updated by maintainers, and lacks production planner-path proof. A future replacement should apply one canonical cadence decision to both normal selection and floor backfill, with an end-to-end planner regression.

Thanks for identifying the repeated-review cost and for iterating on the workflow wiring.

PR #584, opened 2026-07-14
fix(state): keep immutable ledger publication live on hot branches

vincentkoc:
## What Problem This Solves

Fixes an issue where action-ledger receipt publication could exceed its workflow deadline when the shared state branch was changing continuously. The immutable publisher repeatedly rebuilt candidates through the full state worktree after each lost race, making otherwise valid receipts fail on a hot branch.

## Why This Change Was Made

Rebuild each immutable candidate from the observed remote tree in a temporary Git index, overlay only missing frozen source blobs, and publish the resulting commit with an exact-OID force-with-lease. The real index and worktree remain untouched between CAS retries; collision and equivalent-replay behavior remain fail-closed and idempotent.

## User Impact

Operators can expect review, receipt, and action-ledger events to converge under sustained state-branch contention instead of timing out on repeated reset, checkout, stage, and commit work.

## Evidence

- Production failures before this change: https://github.com/openclaw/clawsweeper/actions/runs/29372680279 and https://github.com/openclaw/clawsweeper/actions/runs/29372680763.
- `node --test test/repair/git-publish.test.ts`: 67/67 passed.
- The strengthened mutable-storm regression forces 12 consecutive remote advances and asserts retries perform zero checkout, one initial stage/commit, two total resets, and at least 13 object-level `commit-tree` candidates.
- `./node_modules/.bin/tsc -p tsconfig.repair.json` passed.
- Repair source/test oxlint and oxfmt checks passed.
- Live receipt replay and fresh dual-dispatch canary will run after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ checkpoint, and status-only commits are intentionally omitted.
 
 ### Changed
 
+- Rebuilt immutable action-ledger race candidates directly from Git trees and
+  temporary indexes, avoiding repeated state-worktree reset, checkout, stage,
+  and commit work so hot state branches can converge within their deadline.
 - Bounded immutable action-ledger publishers' priority yield to exclusive state
   leases so a valid long-lived lease cannot starve a shorter immutable
   publication deadline; exclusive publishers still rebuild around branch races.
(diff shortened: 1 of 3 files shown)


vincentkoc:
Closing as part of the July 14 runaway-session cleanup. This campaign branch is superseded by the full recovery rollback in https://github.com/openclaw/clawsweeper/pull/586; preserve any still-useful idea as a fresh, independently scoped change after recovery.

PR #585, opened 2026-07-14
fix(state): roll back saturated immutable canary

joshavant:
## Summary

- return GitHub activity ledger publication and receipt replay to exclusive state publication
- keep the repair-native immutable publisher available, but do not expand it to exact reviews
- document the live canary outcome and lock the workflow expectation to the exclusive path

## Live evidence

Two production publishers on the old 90-second budget lost four consecutive state races and failed. After #583 raised the immutable budget to 180 seconds and 128 attempts, a manual replay of the still-missing durable receipt bundle lost nine consecutive races and failed after 191 seconds / 286 git processes. The state branch advanced every few seconds while each immutable rebuild took roughly 17-19 seconds, so another deadline increase would consume runner time without creating a credible CAS window.

Failed post-merge canary: https://github.com/openclaw/clawsweeper/actions/runs/29374066539

Exact-review immutable publication was never enabled. No receipt is dropped: the source artifact remains eligible for the existing replay workflow, and will be replayed after this rollback reaches main.

## Validation

- `pnpm run format`
- `pnpm run check:active-surface`
- `pnpm run check:limits`
- `pnpm run build:repair`
- `node --test test/repair/dispatch-action-receipts.test.ts dist/repair/action-event-publisher.test.js` (26/26)

This PR is a single standalone commit directly on current `main`; it does not depend on another open PR.

### DIFF
--- a/.github/workflows/github-activity-receipt-replay.yml
+++ b/.github/workflows/github-activity-receipt-replay.yml
@@ -181,7 +181,6 @@ jobs:
 
       - name: Replay GitHub activity dispatch action ledger
         env:
-          CLAWSWEEPER_ACTION_LEDGER_IMMUTABLE_PUBLISH: "1"
           PUBLISH_TOKEN: ${{ steps.replay-state-token.outputs.token }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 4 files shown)


vincentkoc:
Thanks @joshavant — the saturated immutable-canary diagnosis and live evidence here are correct and useful. The canonical recovery at https://github.com/openclaw/clawsweeper/pull/586 has to go broader: it removes the immutable publisher and replay stack itself because those components were introduced by the runaway July 14 campaign, so this narrower branch cannot land independently without preserving the originating failure surface. Closing as superseded while retaining this PR as the evidence record.

PR #589, opened 2026-07-15
fix(review): stop requiring routine PR rebases

hxy91819:
## Summary

- stop treating an ordinary behind branch as a contributor blocker or recommending a routine rebase
- reserve rebase/resolve guidance for conflicts, unavailable merge trees, or concrete merge-result integration failures
- add a non-blocking maintainer-facing Base freshness metric when the merge base is at least 7 days old
- avoid daily review-cache churn by invalidating only when freshness state or merge-base identity changes

## Motivation

High-throughput repositories can advance by thousands of commits per day. Commit-count drift alone therefore creates noisy, repeated contributor work without proving that a PR is unsafe to merge. This keeps contributor action tied to concrete integration evidence while retaining a simple signal for genuinely old branches.

## Impact

Only PRs whose GitHub mergeable state is behind trigger the compare lookup. A stale result adds review context and a non-blocking metric for maintainers or merge automation; it does not lower the PR rating or add contributor next steps. Compare failures fail soft.

## Validation

- Node 24.15.0: build passed
- Node 24.15.0: 49 focused review prompt/content-cache tests passed
- formatting, TypeScript compilation, and lint passed as part of `pnpm run check`
- full `pnpm run check` did not complete because Codex runner tests reached the real Codex API and received 401 Unauthorized
- autoreview was attempted with the bundled helper, but the same Codex API 401 prevented a review result

### DIFF
--- a/prompts/review-item.md
+++ b/prompts/review-item.md
@@ -234,13 +234,19 @@ or absent. Use it only when the PR diff changes automation behavior or
 plausibly causes CI, automerge, proof capture, label sync, or related automation
 to fail after merge.
 Do not treat a branch being behind the current base as proof that merging the
-PR will delete current-base-only files or commits. When GitHub reports the PR as
-mergeable or clean and the only concern is stale base drift, describe it as
-needing rebase or review refresh in `risks`, `workReason`, or `bestSolution`,
-but leave `reviewFindings` and `mergeRiskLabels` focused on defects or risks
-that survive the actual three-way merge result. Use deletion/drop wording for
-current-base behavior only when a merge result, merge ref, conflict, or concrete
-patch evidence shows that the merged PR would remove or regress it.
+PR will delete current-base-only files or commits. Ordinary `behind` state is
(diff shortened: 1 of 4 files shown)


steipete:
Maintainer decision (relayed from @steipete): **policy adopted.** Behind-only branches stop being a contributor blocker; rebase/resolve guidance is reserved for conflicts, unavailable merge trees, and concrete merge-result integration failures, with the non-blocking base-freshness note at 7+ days as proposed. Please move this out of draft and rebase onto current `main` — the review prompt and cache have shifted since Jul 15. Happy to land it once CI is green.

steipete:
Closing during maintainer cleanup: whether routine PR rebases should stop being required is a policy decision, and a draft PR is the wrong vehicle for it — the change would sit here unmergeable until someone rules, which is what has happened since June. If you want to pursue it, the fastest route is a short issue laying out the tradeoff (stale-branch review risk vs contributor friction) so the maintainers can decide; the implementation is small once the policy is settled. Thanks for the nudge on the friction — it is a real cost. 🦞

PR #591, opened 2026-07-15
test(repair): execute containment worker from isolated cwd

hxy91819:
## Summary

- execute the built containment worker from a temporary isolated working directory
- verify the worker loads through its absolute entry path and returns the expected structured result
- clean up the temporary workspace even when the assertion fails

## Why

The regression coverage added in https://github.com/openclaw/clawsweeper/pull/590 prevents the workflow from returning to a relative worker path, but it checks the workflow text. This adds behavior-level proof that the worker entrypoint can actually load after the child process changes to the isolated work directory. The prior relative invocation would fail this scenario with `MODULE_NOT_FOUND`.

## Impact

Test-only change. Production containment and security boundaries are unchanged.

## Validation

- focused containment test: 7 passed
- `pnpm run check`: passed
  - unit, repair, changed coverage, and full coverage suites: 0 failures
  - environment-dependent namespace/Landlock cases skipped on this host as designed
- autoreview (`--mode uncommitted`): clean, no actionable findings
- `git diff --check`
- `pnpm run format:check`


steipete:
Merged — thanks @hxy91819! Good behavior-level proof to back the workflow-text pin from #590.

PR #610, opened 2026-07-15
fix(queue): unblock exact review publication

brokemac79:
## Summary

- give exact-review artifact publication its own serial GitHub Actions concurrency lane instead of placing it behind every bulk state publisher
- recover both new and already-persisted unclaimed publication dispatches after 15 minutes, with stale lease tuples rejected safely
- use the existing five-minute dashboard Worker cron as a guaranteed post-deploy queue heartbeat, and correct the false future/zero-age dispatch telemetry

This fixes the confirmed path where a review completed successfully but its final ClawSweeper comment and labels could not be published. It deliberately does **not** bulk-rewrite every pending review row; the separate stale-pending recovery/capacity follow-up remains necessary if the producer backlog does not drain normally.

## Incident evidence

At `2026-07-15T18:00:34Z`, the live queue API reported:

- `pending=1961`, `dispatching=1`, `leased=15`
- oldest pending item `2026-07-15T11:01:02Z`
- 48 of 64 review slots shown as available
- a false `oldest_dispatching_at=2026-07-22T13:10:22Z` with age zero
- `next_wake_at=2026-07-15T20:05:52Z`

The false future timestamp came from deriving the dispatch start by subtracting the normal six-minute lease from the seven-day publication lease introduced with #592. It made a multi-hour stuck handoff look healthy, but it was telemetry fallout rather than the cause.

The publication failure itself is directly visible in Actions:

- [run 29418496158](https://github.com/openclaw/clawsweeper/actions/runs/29418496158), `Review event item openclaw/openclaw#107676@publish:29409874087:1`, was created at `13:16:24Z` and was still pending at `18:00Z`
- its only relevant job, [87362614333](https://github.com/openclaw/clawsweeper/actions/runs/29418496158/job/87362614333), had no runner and had never completed
- an earlier exact publisher, [job 87336341784](https://github.com/openclaw/clawsweeper/actions/runs/29410554983/job/87336341784), was created at `11:07:50Z` but did not start until `13:14:06Z`
- a representative bulk publisher, [job 87350685065](https://github.com/openclaw/clawsweeper/actions/runs/29414724394/job/87350685065), was created at `12:22:22Z` and did not start until `16:10:32Z`

[...]

Issue #574, opened 2026-07-14
normalizeGitHubActor: not idempotent — strips only one trailing [bot]


anagnorisis2peripeteia:
`normalizeGitHubActor(login)` in `src/repair/comment-router-utils.ts:185` strips a single trailing `[bot]` via `/\[bot\]$/i`, so a doubled suffix leaves a residual that a second pass strips again.

Repro:
```
normalizeGitHubActor("x[bot][bot]") === "x[bot]"
normalizeGitHubActor("x[bot]")      === "x"        // => not idempotent
```
A normalizer should be idempotent. Loop/global-strip the suffix (though a doubled [bot] is rare in practice). Related: `stripEnvPrefix` in `repair/validation-command-utils.ts` has the same single-pass shape.

brokemac79:
Thanks for the report. We?ve confirmed this is separate from the recent ClawSweeper review/publication stability incident.

We will review this carefully with its linked implementation once the immediate stability work is complete. Because it intersects with actor normalisation and mutation-authorisation boundaries, we need to assess the final behaviour deliberately rather than rush a decision during recovery.

Please do not treat this as a request for further changes at present.

brokemac79:
@clawsweeper re-review

steipete:
Reopening: https://github.com/openclaw/clawsweeper/pull/576 was deliberately narrowed before merge to the `compactText` fix for https://github.com/openclaw/clawsweeper/issues/573 only. This issue's implementation was removed from the final diff, so the merge did not fix it. GitHub nevertheless applied the PR's older closing relationship at merge time.

No new contributor changes requested. This restores the pre-merge open state for separate maintainer triage.

PR #616, opened 2026-07-16
fix: run exact publishers on current main

brokemac79:
## Summary

Repairs queued exact-review publication without reducing the queue's existing
parallel publisher capacity.

- Validate the immutable artifact against its producer SHA, but run the
  publisher from current `main` so repaired publication logic is used.
- Recognize legacy artifacts with no durable review-lease tuple and safely
  enqueue one fresh exact review instead of attempting a stale target mutation.
- Remove the exact publisher's second action-ledger state publication, which
  amplified shared `clawsweeper-state` contention after the actual review
  publication had already completed.

## Incident evidence

This is deliberately not a claim that #592 was the original incident cause.

- `openclaw/openclaw#107787`'s review run
  [29408820027](https://github.com/openclaw/clawsweeper/actions/runs/29408820027)
  started at `2026-07-15T10:38:19Z`, 26 seconds **before** #592 merged at
  `10:38:45Z`. It created the full review comment at `10:46:50Z`, then its
  publisher reported `RoutableSyncPublishRaceError` while committing the
  durable result. That prevented the deferred verdict router from applying the
  expected target labels. The comment is visible at
  [#107787 comment 4979684095](https://github.com/openclaw/openclaw/issues/107787#issuecomment-4979684095).
  This proves the shared-state publication failure predates #592.
- #592 separately split exact review generation from publishing and checked the
  publisher out at the artifact's historical `source_sha`. A queued artifact
  therefore runs code from before later publisher repairs. #593 corrected the
  bundle-command spelling but did not change that historical checkout.
- #612 added the correct current-main behavior for legacy tuple-less artifacts:
  requeue a fresh review instead of applying an artifact that cannot prove the
  visible review lease. That behavior cannot help an artifact whose publisher
  is still executing a pre-#612 checkout. A representative failed publisher is
  [run 29460181512](https://github.com/openclaw/clawsweeper/actions/runs/29460181512)
  for [issue #574](https://github.com/openclaw/clawsweeper/issues/574), which
  reported that the live durable tuple existed while its local report had no
  durable lease identity.
- The exact action ledger is observability data. Keeping it as a second
  best-effort state write makes the shared-state race worse; it is not required
  to publish the verdict or route labels.

Related: #592, #593, #610, #611, #612, #614.

## Implementation and safety

The artifact validator still binds the producer SHA and target identity. Only
the implementation checkout changes to `main`. Legacy tuple-less artifacts do
not receive state/target publication credentials and do not mutate their target;
they dispatch one signed fresh exact review using the existing source-drift
route. Normal exact publication keeps its existing 24-slot capacity and broad
publication stays target-scoped.

This is an immediate repair for historical artifact execution and avoidable
state-write pressure. It does not claim to fully solve the older shared-state
race demonstrated by #107787. A separate bounded state-write critical-section
repair remains warranted.

## Validation

```text
pnpm run build:all
pnpm exec node --test --test-name-pattern 'exact event review hands immutable artifacts|publish workflow dispatches immediate apply|trusted comment router owns command ledger capacity retries|every action-ledger publication authenticates' test/sweep-workflow.test.ts test/clawsweeper-action-ledger.test.ts test/repair/command-ops-action-ledger.test.ts
pnpm exec node -e "const fs=require('node:fs'); const YAML=require('yaml'); YAML.parse(fs.readFileSync('.github/workflows/sweep.yml','utf8')); console.log('sweep.yml parsed')"
git diff --check
codex review --uncommitted
git fetch origin main --prune
codex review --base origin/main
```

Focused proof: 6 passing tests; TypeScript build and workflow YAML parse pass.
Both Codex review passes were clean. `actionlint` is not installed on this host.
The full workflow suite has five established native-Windows baseline failures
because WSL cannot execute `/bin/bash`; the touched tests pass and a Linux
Crabbox build completed.

## Risk and rollout

This touches a GitHub Actions workflow, so GitHub App/OAuth workflow scope is
the deployment risk. Roll out by observing the next legacy and current exact
publisher runs: legacy artifacts should complete by requeueing once, while
normal artifacts should publish/routably defer without the extra ledger commit.
No bulk backlog replay is included.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -1344,7 +1344,9 @@ jobs:
       - uses: actions/checkout@v7
         if: ${{ steps.publication-context.outputs.claimed == 'true' }}
         with:
-          ref: ${{ steps.publication-context.outputs.source_sha }}
+          # The artifact validator binds the producer SHA. The publisher itself
+          # must run the current repaired implementation, not historical code.
+          ref: main
           filter: blob:none
           fetch-depth: 0
           persist-credentials: false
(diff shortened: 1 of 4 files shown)


brokemac79:
@clawsweeper review

PR #645, opened 2026-07-17
Show stale review status during refresh

hxy91819:
## Summary

- project an existing durable ClawSweeper review into a visible stale/in-progress state after the exact-head review lease is confirmed
- collapse the previous review while stripping actionable verdict, action, security, version, and identity markers
- replace abandoned or expired refreshes with a tuple-bound interrupted state
- preserve prior-review digest, context, and review-history continuity for cache revalidation and the next publication
- ignore canonical refresh projections in the repair router

## Root cause

ClawSweeper coordinates a new review with a separate transient lease comment, but the older durable review comment remains unchanged higher in the timeline. That makes the previous verdict look current while a new review is already running and forces maintainers to inspect comment history or find the transient status manually.

## Safety

- the dedicated lease comment remains the authoritative coordination record; the durable projection is presentation-only and best effort
- projection writes happen only after lease election and use the live comment ETag with `If-Match`, so a stale read cannot overwrite a concurrently published verdict or newer projection
- publication remains fenced to the matching refreshing lease tuple; interrupted tuples cannot publish
- old automation markers are removed, including for comments near GitHub's size limit, and the repair router explicitly ignores validated projections
- malformed projections and displaced-body digest mismatches fail closed

## Validation

- `pnpm run check`
- focused review status, history, rendering, and repair-router tests (180 passing)
- `gitleaks protect --staged --redact --no-banner` (no leaks found)
- `autoreview --mode local` (clean after addressing three actionable findings)

### DIFF
--- a/docs/pr-review-comments.md
+++ b/docs/pr-review-comments.md
@@ -27,10 +27,16 @@ Each synced comment includes the durable identity marker:
 ClawSweeper edits that comment in place instead of posting repeated comments.
 Report front matter stores the synced comment id, URL, hash, and sync time.
 
-When review starts and no ClawSweeper-owned comment exists yet, the review
-shard posts a short status placeholder with the same durable identity marker.
-The placeholder is intentionally light and crustacean-friendly, then the final
-review sync edits that exact comment in place.
+The review shard uses a separate transient status comment as its coordination
+lease. When a durable review already exists, acquiring that lease also edits
+the durable comment to show that a fresh review is in progress, marks the old
+review stale, and collapses it as previous context. The status projection has
+no active verdict or action markers, so downstream automation cannot act on the
+displaced review. A matching apply run replaces the projection with the new
+completed review; an abandoned or expired lease marks the refresh interrupted.
+
+When no durable review exists yet, the transient lease comment remains the only
+start status until the first completed review is published.
 
 For a PR that needs work, the visible comment starts with:
 
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -187,6 +187,11 @@ import {
   type ReviewHistoryCycle,
   type ReviewHistoryLedger,
 } from "./review-history.js";
+import {
+  parseDurableReviewStatusProjection,
+  renderDurableReviewRefreshProjection,
+  renderInterruptedDurableReviewProjection,
+} from "./review-comment-status.js";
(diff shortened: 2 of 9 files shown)


hxy91819:
## Live validation result: partial pass, blocking failure

I exercised this branch against disposable test PR https://github.com/openclaw/clawsweeper/pull/646 using `workflow_dispatch` with `--ref feature/durable-review-status`.

### What passed

- A baseline review created durable comment https://github.com/openclaw/clawsweeper/pull/646#issuecomment-5002397642.
- A second review run successfully replaced the final durable review in place: the comment id stayed `5002397642`, while `updated_at` moved from `2026-07-17T10:51:30Z` to `2026-07-17T11:09:30Z` and the visible reviewed time moved to `11:04 UTC`.
- Evidence run: https://github.com/openclaw/clawsweeper/actions/runs/29575130444.

### What failed

- The durable comment never entered `clawsweeper-review-status:refreshing` while lease comment `5002510417` was active.
- The review-shard log shows the fresh projection write was rejected:

  ```text
  [review] could not project durable review refresh for #646:
  gh api ... --method PATCH -H If-Match: W/"53562..." ...
  gh: Bad Request (HTTP 400)
  ```

- A controlled 30-second Codex timeout created lease comment `5002756860`, then removed it when the review failed, but the durable comment entered neither `refreshing` nor `interrupted`.
- Interruption evidence run: https://github.com/openclaw/clawsweeper/actions/runs/29576709251. Its log repeats the same HTTP 400 on the fresh projection write before the expected Codex failure.
- I also tested a temporary user-owned probe comment directly. Both GitHub's weak ETag (`W/"..."`) and the same value without `W/` produced HTTP 400 on issue-comment PATCH. The probe comment was deleted afterward.

### Conclusion

GitHub issue-comment PATCH does not accept this `If-Match` usage. The current implementation therefore rejects fresh projection writes, not only stale writes. Final publication still replaces the durable comment, but the maintainer-facing refreshing/interrupted states do not work, and the requested stale-conditional-write proof cannot pass.

This live proof should be treated as blocking for the current implementation. The concurrency design needs to stop relying on HTTP `If-Match` for issue-comment PATCH, then the same three-path validation should be rerun on PR 646.

hxy91819:
Superseded by https://github.com/openclaw/clawsweeper/pull/674, which includes this durable review-status behavior and its tests together with the per-item shard architecture. Closing this draft to keep one active implementation path.

PR #649, opened 2026-07-17
fix(validation): stop misreading pnpm lockfile integrity as install URLs

zhangguiping-xydt:
Related: https://github.com/openclaw/openclaw/pull/108974

## What Problem This Solves

Fixes an issue where ClawSweeper automerge repairs would stop before validation when a valid pnpm lockfile integrity value contained double slashes.

## Why This Change Was Made

The dependency-install guard now recognizes protocol-relative URLs only when they begin at a URL boundary. This retains the fail-closed network policy for actual install destinations while avoiding false positives from opaque SHA-512 integrity data.

## User Impact

Operators can rerun affected automerge repairs without valid pnpm lockfiles being rejected as unapproved external dependency hosts. Actual non-registry install URLs remain blocked.

## Evidence

- `pnpm run build` and `pnpm run build:repair` passed.
- Focused regression test passed: `dependency setup permits pnpm integrity strings containing double slashes`.
- Focused network-policy tests passed, including semicolon-prefixed HTTPS, HTTPS after a non-boundary character, and `git://` destinations that remain rejected.
- Any explicit `<scheme>://` URL remains detectable at any position; only ambiguous bare protocol-relative `//` tokens require a metadata boundary.
- The regression fixture uses the actual `@mistralai/mistralai@2.4.0` integrity shape that blocked https://github.com/openclaw/openclaw/pull/108974.

### Paired real-behavior proof

Positive path, using the PR's production `prepareTargetToolchain` entry against the exact affected OpenClaw head and lockfile:

```text
PROOF_TARGET=https://github.com/openclaw/openclaw/pull/108974
PROOF_HEAD=34a3001388bb99fb4a041a73aad98631c4557634
PROOF_ENTRY=prepareTargetToolchain
PROOF_INTEGRITY_PRESENT=true
Scope: all 170 workspace projects
Verifying lockfile against supply-chain policies (1410 entries)
PROOF_INSTALL_CHILD_COMPLETED=true
PROOF_TOP_LEVEL_NODE_MODULES_ENTRIES=851
PROOF_MISTRAL_PACKAGE_INSTALLED=true
FORMER_DESTINATION_ERROR_PRESENT=false
```

The isolated pnpm child completed and `prepareTargetToolchain` advanced into its post-install checkout-identity verification. That progression is possible only after the lockfile network guard, supply-chain policy stage, and dependency install succeed. The former false positive did not recur:

```text
target dependency install destination is not approved:
https://ogr2wvngtd7to6tpp5poq==
```

Negative path, through the same production dependency-install guard with a committed `bun.lock` containing `@git://evil.example/payload.git`:

```json
{
  "status": "HOSTILE_URI_REJECTED",
  "message": "target dependency install destination is not approved: null"
}
```

This proves the intended pair: the affected integrity value now proceeds through a real dependency install, while an explicit hostile URI remains fail-closed.

### Remaining limitation

This is not presented as a completed repair-worker run. After the successful install, the local run exceeded its checkout-identity deadline while repeatedly traversing hoisted workspace `node_modules` links. That later timeout is independent of URL classification and occurred only after the behavior changed by this PR had completed successfully.

### DIFF
--- a/src/repair/target-validation.ts
+++ b/src/repair/target-validation.ts
@@ -915,12 +915,12 @@ function assertTrackedPatchDependency(
 }
 
 function assertApprovedInstallMetadataDestinations(text: string, registryOrigin: string) {
-  const networkTokens = [
-    ...text.matchAll(
-      /(?:^|[\s"'`<>{}\x5b\x5d(),;:=])((?:https?:\/\/|\/\/|git\+[^:\s]+:\/\/|ssh:\/\/)[^\s"'`<>{}\x5b\x5d,]+)/gim,
-    ),
+  const explicitNetworkTokens =
+    text.match(/(?:https?:\/\/|git\+[^:\s]+:\/\/|ssh:\/\/)[^\s"'`<>{}\x5b\x5d,]+/gi) ?? [];
+  const protocolRelativeNetworkTokens = [
+    ...text.matchAll(/(?:^|[\s"'`<>{}\x5b\x5d(),;=])(\/\/[^\s"'`<>{}\x5b\x5d,]+)/gim),
   ].flatMap((match) => (match[1] ? [match[1]] : []));
-  for (const token of networkTokens) {
+  for (const token of [...explicitNetworkTokens, ...protocolRelativeNetworkTokens]) {
     assertApprovedInstallUrl(token.replace(/[);]+$/, ""), registryOrigin);
   }
   if (/(?:^|[\s"'`])(?:git@|github:|gitlab:|bitbucket:)/im.test(text)) {
--- a/src/repair/target-validation.ts
+++ b/src/repair/target-validation.ts
@@ -916,7 +916,7 @@ function assertTrackedPatchDependency(
 
 function assertApprovedInstallMetadataDestinations(text: string, registryOrigin: string) {
   const explicitNetworkTokens =
-    text.match(/(?:https?:\/\/|git\+[^:\s]+:\/\/|ssh:\/\/)[^\s"'`<>{}\x5b\x5d,]+/gi) ?? [];
+    text.match(/[A-Za-z][A-Za-z0-9+.-]*:\/\/[^\s"'`<>{}\x5b\x5d,]+/g) ?? [];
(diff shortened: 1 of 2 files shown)


zhangguiping-xydt:
@steipete Could you take a look when you have a moment?

**TL;DR:** False positive in the dependency-install network guard. The bare `//` matcher was matching `//` inside pnpm SHA-512 integrity values (e.g. `...X//OgR2w...==`), treating opaque lockfile data as a protocol-relative install URL and failing closed on valid pnpm lockfiles. This blocked the automerge repair on https://github.com/openclaw/openclaw/pull/108974.

**Fix shape (narrow):**
- Explicit URI schemes (`https://`, `git://`, `git+ssh://`, `ssh://`, `ftp://`, …) keep unrestricted detection at any position — old behavior preserved.
- Only bare protocol-relative `//` tokens now require a metadata boundary, so SHA-512 integrity data no longer triggers the guard.

**Verification:**
- `pnpm run build` and `pnpm run build:repair` pass.
- Regression test with the exact `@mistralai/mistralai@2.4.0` SHA-512 integrity shape from #108974 now proceeds; previously the same fixture threw `destination is not approved`.
- Hostile-destination tests still fail closed: added `;https://evil.example/...`, `@https://evil.example/...`, `@git://evil.example/...` to the rejection table.

**ClawSweeper status (head `fddcb94415`):**
- Security: cleared — no concrete bypass remains in the final diff.
- Patch quality: 🐚 platinum hermit.
- All code-level P1 findings from the prior 5 review cycles are resolved.
- Remaining blocker is real-behavior proof only: my local run against the #108974 head gets past the former validation failure but stops at Corepack fetching `https://registry.npmjs.org//pnpm/11.2.2` from my network. If you'd like full repair-worker logs, I can re-run on a credentialed box (Crabbox/testbox) — happy to do that if you prefer Option 2 (accept test-backed risk) is not your call.

Mirror of ClawSweeper's latest review: https://github.com/openclaw/clawsweeper/pull/649#issuecomment-5004144536 (scroll to current verdict).

PR #670, opened 2026-07-18
fix(validation): disable implicit pnpm installs

hxy91819:
## Summary

- disable pnpm's `verifyDepsBeforeRun` behavior for validation commands
- keep the existing lifecycle-hook suppression and workspace filter fail-closed behavior
- update focused command-boundary assertions

## Root cause

Dependency setup and validation intentionally run with separate disposable homes and pnpm stores. With pnpm 11.2.2, `verifyDepsBeforeRun` treats that profile/store change as a stale install and can run an implicit `pnpm install` before `pnpm check:changed`.

That hidden install rewrites runtime metadata such as `node_modules/.modules.yaml` and `.pnpm-workspace-state-v1.json`. ClawSweeper then correctly rejects validation because the checkout identity changed; on a full OpenClaw checkout, pnpm may also purge/reinstall the dependency tree or attempt registry access.

Dependency installation is already a separate reviewed phase. Validation should execute only the requested script, so the production command now passes `--config.verify-deps-before-run=false` alongside `--config.enable-pre-post-scripts=false`.

## Validation

- OpenClaw-shaped container E2E: happy path passed with different setup/validation homes and stores
- OpenClaw-shaped historical CI regression scenario passed
- `node --test --test-name-pattern='package validation execution suppresses lifecycle hooks and implicit installs' test/repair/target-validation.test.ts`
- `pnpm --config.verify-deps-before-run=false run build`
- `pnpm --config.verify-deps-before-run=false run build:repair`
- `pnpm --config.verify-deps-before-run=false run lint`
- focused `oxfmt --check`
- isolated autoreview: no findings, correctness confidence 0.97

### DIFF
--- a/src/repair/validation-command-utils.ts
+++ b/src/repair/validation-command-utils.ts
@@ -712,9 +712,13 @@ export function validationCommandForExecution(parts: readonly string[]): string[
     return [...envPrefix, "npm", "--ignore-scripts", ...commandParts.slice(1)];
   }
   if (invocation.executable === "pnpm") {
+    // pnpm can otherwise run an implicit install when its disposable validation
+    // profile uses a different store. Dependency setup is a separate reviewed
+    // phase, so validation must execute the requested script without mutation.
     return [
       ...envPrefix,
       "pnpm",
+      "--config.verify-deps-before-run=false",
       "--config.enable-pre-post-scripts=false",
       ...commandParts.slice(1),
     ];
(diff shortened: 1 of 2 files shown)


hxy91819:
@clawsweeper re-review

Rebased current head: `bc3496b8ed620afe27b2ca3d42d68b892c7e667c`. The production-call-site invariant has been checked: all four validation entry points invoke `prepareTargetToolchain()` before validation, so `--config.verify-deps-before-run=false` removes mutation from validation without skipping dependency setup.

PR #720, opened 2026-07-20
fix(publish): prevent state races from timing out result delivery

brokemac79:
Related: #712
Related: #714
Related: #716

## What Problem This Solves

Fixes an issue where completed issue and PR reviews remain queued when concurrent exact-result publishers race to update the shared state branch.

The incident began immediately after #712 merged at 2026-07-20 06:04:34Z. That change made a shallow publisher hydrate Git history when `merge-base` could not see common ancestry. Under the normal 32-writer publication load, losing writers repeatedly entered `--deepen` / `--unshallow`; many timed out after 60 seconds and held their publication lease through the failed attempt. The queue was still at 215 publication items with all 32 slots active at 2026-07-20 13:21Z.

This is not a label-only or dashboard-only issue: the failure occurs in the exact-result publication step before the durable result can finish its comment/label/application lifecycle.

## Why This Change Was Made

Exact event publication commits one known record tuple. This change gives that path a dedicated `pushSingleRecordTupleCommit` entry point which:

- captures the source commit and derives the one allowed tuple key from the staged publication paths;
- opts only that bounded path into #716's remote-head rebuild when shallow history hides the merge base;
- avoids history hydration in the hot two-writer race; and
- keeps the existing tuple arbitration, including preservation of a newer concurrent remote tuple.

All generic `pushCommit` callers retain the existing #714/#716 hydration behavior. This does not change queue capacity, add retries, reset state, alter gates, dispatch Actions, or weaken reconciliation safety.

## User Impact

Issue and PR result publishers that lose a state-branch push race can reconcile immediately on the fetched remote head instead of spending up to 60 seconds hydrating history and failing. This should release publication leases promptly and let the backlog drain while preserving both independent record tuples and the newer winner for same-tuple races.

GitHub API rate limiting remains a separate possible failure mode; this patch intentionally addresses the dominant Git fetch timeout signature rather than masking it with capacity or retry changes.

### Production causal proof

- Representative post-#716 failure: [workflow run 29741488075](https://github.com/openclaw/clawsweeper/actions/runs/29741488075), job [88349031217](https://github.com/openclaw/clawsweeper/actions/runs/29741488075/job/88349031217), on default-branch SHA `846b73c666c01b1c29d8a65a93549fca5742b404`.
- The job entered `Publish event result and apply safe close` at 2026-07-20 12:17:06Z and failed after 9m24s. Its log shows rejected state pushes, no visible common base, #712's shallow-recovery helper, `fatal: shallow file has changed`, and `GitCommandTimeoutError: git fetch timed out after 60000ms` before retryable completion.
- A read-only sample of 20 completed publisher failures during the incident found 18 with this direct 60-second Git fetch timeout signature and 2 with GitHub rate-limit failures.
- `git log -S mergeBaseWithShallowRecovery` traces the failing helper to #712 (`25135198f921b062b6d276931ca794d0222129c6`). Timing alone is not the proof: the runtime stack and exact failing command both enter that helper.
- #716 correctly added a bounded remote-head rebuild, but the exact event publisher did not provide its source commit and tuple bound, so it still had to hydrate first. This PR supplies that missing bound instead of reverting #714/#716's correctness work.

### Real behavior proof

Environment: Crabbox `local-container`, Docker image `mcr.microsoft.com/playwright:v1.60.0-noble`.

Focused lease: `cbx_00f30e0daaa8` (`pearl-shrimp`). The Docker run built `tsconfig.repair.json`, ran the exact two-writer regression plus the existing shallow/unrelated-history guards, checked the production call-site contract, and ran scoped oxlint, oxfmt, and `git diff --check`.

```text
tests 4
pass 4
fail 0
Found 0 warnings and 0 errors.
All matched files use the correct format.
exitCode: 0
```

The new integration case creates a depth-1 publisher and a second writer against a real bare Git remote. The second writer wins the first push; a shim fails the test if `git fetch --deepen` is attempted. The publisher then rebuilds on the fetched remote head, pushes successfully, remains shallow, and the remote contains both the publisher's tuple and the concurrent writer's tuple.

Broad lease: `cbx_53da04c5a2c5` (`golden-krill`). The Docker run rebuilt all three TypeScript targets and ran the complete repair suite:

```text
tests 987
pass 980
fail 0
skipped 7
duration_ms 48006.822504
Found 0 warnings and 0 errors.
All matched files use the correct format.
exitCode: 0
```

The broad Crabbox setup materialized untouched files from Git's LF index because raw Windows-to-Linux sync otherwise presents repository-wide CRLF fixture noise. The initial raw-sync failure was entirely in workflow text/contained `pnpm` harness expectations; after Linux checkout normalization and a Corepack `pnpm` shim, the unchanged full suite passed without exclusions.

Not proven pre-merge: production backlog drain. The deterministic test proves the exact failed two-writer lifecycle and removes the 60-second hydration command from that bounded path; live drain must be observed after deployment.

### Review gates

- Dirty patch: `codex-review --mode local --full-access` — clean, no accepted/actionable findings.
- Committed branch vs actual base: `codex-review --mode branch --base origin/main --full-access` — clean, no accepted/actionable findings.
- The branch review also exercised the same-tuple shallow race and confirmed the newer remote tuple remains the winner.
- `git diff --check` passed on the host and on the scoped Linux patch.

## Risks And Rollout

The behavior change is opt-in and limited to exact single-record result publication. A guard rejects zero or multiple tuple identities before enabling the bounded rebuild. Generic checkpoint, ledger, and other reconciliation callers are unchanged.

The main rollout signal is the publication queue: the active-lease timeout signature should disappear first, then leased duration and ready depth should fall. A remaining GitHub rate-limit signature would be separate evidence rather than a reason to broaden this patch.

No automerge is requested. A maintainer should make the explicit merge decision after normal CI and review-stage evidence.

## Attribution

This repair builds on Peter Steinberger's state recovery and remote-head reconciliation work in #712, #714, and #716. It preserves those correctness changes while narrowing the exact event publisher's contention path.

### DIFF
--- a/src/repair/git-publish.ts
+++ b/src/repair/git-publish.ts
@@ -1402,6 +1402,7 @@ export function pushCommit(options: {
   rebaseStrategy?: RebaseStrategy;
   reconciliationSourceCommit?: string;
   reconciliationTupleKeys?: ReadonlySet<string>;
+  boundedRemoteHeadRebuild?: boolean;
 }): boolean {
   const remote = options.remote ?? "origin";
   const branch = options.branch ?? publishDefaultBranch();
@@ -1434,6 +1435,7 @@ export function pushCommit(options: {
           branch,
           options.reconciliationSourceCommit,
           options.reconciliationTupleKeys,
+          options.boundedRemoteHeadRebuild,
         )
       ) {
         return false;
@@ -1481,6 +1483,30 @@ export function pushCommit(options: {
   return spawnGit(["push", remote, `HEAD:${branch}`]).status === 0;
 }
 
+export function pushSingleRecordTupleCommit(options: {
+  paths: readonly string[];
+  remote?: string;
+  branch?: string;
+  pushAttempts?: number;
+}): boolean {
+  const reconciliationTupleKeys = recordTupleKeysForPaths(options.paths);
+  if (reconciliationTupleKeys.size !== 1) {
+    throw new Error(
+      `Single-record reconciliation requires exactly one tuple; received ${reconciliationTupleKeys.size}`,
+    );
+  }
+  const reconciliationSourceCommit = runGit(["rev-parse", "HEAD"], { quiet: true }).trim();
+  return pushCommit({
+    ...(options.remote !== undefined ? { remote: options.remote } : {}),
+    ...(options.branch !== undefined ? { branch: options.branch } : {}),
+    ...(options.pushAttempts !== undefined ? { pushAttempts: options.pushAttempts } : {}),
+    rebaseStrategy: "reconcile-records",
+    reconciliationSourceCommit,
+    reconciliationTupleKeys,
+    boundedRemoteHeadRebuild: true,
+  });
+}
+
 function reconciliationChangesOverlap(
   remoteRef: string,
   reconciliationSourceCommit?: string,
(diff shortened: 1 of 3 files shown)


brokemac79:
@clawsweeper review

PR #726, opened 2026-07-20
fix(dashboard): bound Bay terminal records to tide

brokemac79:
## Summary

Stop OpenClaw Bay from replaying terminal outcomes from earlier tide windows. The terminal pools now retain only records admitted after the last completed wash, so old completed/failed entries cannot reappear after the bounded recent-event dedupe rolls over.

## Problem

The public Bay snapshot was showing historical terminal rows even though the visible tide was recent. At 2026-07-20T21:35Z, the live `bay-demo` status contained success and failure outcomes approximately 238–316 hours old while `last_tide_at` was 2026-07-20T20:13:57Z.

The previous projection relied on a bounded `seen_events` list. Once that list rolled over, an old immutable completed GitHub run could be admitted into the terminal buffer again.

## Implementation

- Persist a terminal-window watermark when a tide washes a batch.
- Persist the event IDs on the exact watermark timestamp, so a replay of the washed boundary event is still rejected while a distinct event completed in the same second is retained.
- Filter stored and incoming terminal records to the current tide window before the normal bounded buffer logic.
- Safely migrate existing state that has no stored watermark: retain its recent tail, discard historical stale records with a one-hour bootstrap cutoff, and then switch to the exact persistent watermark on subsequent updates.

`last_tide_at` remains the physical tide timestamp shown to users. The new internal watermark is the completion time of the last washed item; this preserves terminal items that arrive later in the same collection batch instead of accidentally discarding them.

## API impact

No new GitHub API, GraphQL, or dashboard refresh calls are added. This is a local state-projection and persistence fix over the terminal evidence already collected for OpenClaw Bay.

## Validation

```text
pnpm run build:dashboard
pnpm run lint:dashboard
node --test test/dashboard-worker.test.ts
git diff --check
```

- Build and dashboard lint passed.
- Focused dashboard worker suite passed: 156/156.
- Live public-status simulation reduced a stale terminal snapshot from 18 records to 4 current-window records (14 historical replays removed).
- Added coverage for stale replay rejection, a valid same-second boundary event, bounded dedupe rollover, two tides in one merge, migration of legacy state, and preservation of a legitimate legacy tail.

## Review closeout

```text
codex review --uncommitted
codex review --base origin/main
```

Both reviews completed clean. Earlier accepted review findings on post-tide tails, legacy-state tails, and same-second watermark boundaries were fixed and the focused proof was rerun after each change.

## Rollout / caveat

On the first state merge after deployment, old Bay state without the new watermark uses a conservative one-hour bootstrap. This deliberately clears historic terminal rows that were already stale; a genuinely current legacy tail is preserved. Thereafter, terminal eligibility is exact and tide-window based.

After deployment, verify that Bay terminal entries are newer than the current tide window, that current outcomes continue to enter the pools, and that a tide clears only its threshold batch without allowing washed rows to return.

### DIFF
--- a/dashboard/worker.ts
+++ b/dashboard/worker.ts
@@ -82,6 +82,7 @@ const BAY_TIDE_THRESHOLD = 20;
 const BAY_SEEN_EVENT_LIMIT = 256;
 const BAY_WASH_VISIBLE_MS = 60_000;
 const BAY_TIMING_WINDOW_MS = 60 * 60 * 1000;
+const BAY_INITIAL_TERMINAL_LOOKBACK_MS = BAY_TIMING_WINDOW_MS;
 const BAY_TIMING_MAX_SAMPLE_MS = 24 * 60 * 60 * 1000;
 const BAY_JOURNEY_LIMIT = 100;
 const BAY_JOURNEY_TTL_SECONDS = 24 * 60 * 60;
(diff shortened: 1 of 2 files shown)


brokemac79:
@clawsweeper review

PR #762, opened 2026-07-22
docs: record batching rollout status

hxy91819:
## Summary

- record the first three successful live size-2 publication commits and their two accepted acknowledgements
- distinguish successful batch behavior from the still-blocked formal telemetry and capacity gates
- record the current no-go production baseline and the remaining `2 -> 4 -> 8` rollout sequence
- make batch size 8 the explicit final rollout target while preserving rollback and evidence requirements

## Why

The previous plan still described the first size-2 behavior proof as entirely pending and said coordinator configuration did not yet exist in production. Live runs now prove the Git and queue behavior, but the dashboard remains stale and throughput is below arrivals. The plan needs to preserve that distinction so a successful commit is not mistaken for authorization to increase batch size.

## Validation

- `CI=true pnpm exec oxfmt --check docs/state-publication-batching-plan.md`
- `git diff --check`

This is a documentation-only change. No live apply, close, workflow cancellation, DLQ replay, or production configuration change was performed.

### DIFF
--- a/docs/state-publication-batching-plan.md
+++ b/docs/state-publication-batching-plan.md
@@ -1,12 +1,22 @@
 # State publication batching plan
 
-**Status:** PR 1 through PR 4, the rollout hotfix, the repository-wide FIFO
-state-writer coordinator, and the fence identity hotfix are merged. Production
-remains fixed at batch size 2 and a 60-second maximum wait. Coordinator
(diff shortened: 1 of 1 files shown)


hxy91819:
Superseded by merged https://github.com/openclaw/clawsweeper/pull/764, which carries this rollout-plan update together with the implementation. Future rollout PRs will continue updating the same document.

PR #775, opened 2026-07-22
Re-advance state publication batches to eight

hxy91819:
## Summary

- raise the dashboard's effective publication claim cap from 4 to 8
- retain the workflow publisher ask of 32 introduced by #773 as a candidate scan allowance
- clamp actual batch ownership to the dashboard's configured lease size
- persist the effective lease cap so retries, manifests, and writer telemetry remain consistent across deploys and rollbacks
- retain the 60-second wait and logical durable-item admission serialization from #772
- record the duplicate-item incident, rollback, recovery proof, and executable re-advance gates in the rollout plan

## Why re-advance

Size 4 passed its safety gate but could not drain arrivals. The first size-8 rollout exposed two separate production signals. The reproducible failure was caused by distinct queue events targeting the same durable item path; #772 now admits at most one event per `targetRepo#itemNumber` in a batch and leaves the later event pending.

Post-hotfix recovery run 29891989901 selected four distinct items, published one state commit, and reported `materialized=4`, `accepted=4`, `retryable=0`, and `released=0`. The independent writer-identity signal did not recur. Earlier run 29891630162 also showed that one event each for the previously colliding items remained consumable in a later eight-item commit. The queue remains the effective size boundary: the workflow may ask to scan 32 candidates, while the dashboard leases at most 8. This PR makes that separation explicit so the larger ask cannot create an unsupported 32-member commit, and records 8—not 32—as the configured writer size used by fullness telemetry. The recorded cap is immutable for an active lease, including same-ID retries across configuration changes.

## Validation

- focused cap/migration/protocol/telemetry run: 40 tests passed
- fixed container: Node 24.18.0, Git 2.47.3, pnpm 11.10.0; full `pnpm run check` passed with 2,538 passed, 0 failed, and 8 skipped
- local committed-range ClawSweeper review on final head: complete, high confidence, no findings, security cleared

## Rollout gate

Merge only after CI and CodeQL pass. After merge, require dashboard deployment and smoke before observing production. Keep size 8 only after one eight-distinct-item commit has eight accepted outcomes, zero retryable outcomes, no contention or open-DLQ growth, and two consecutive five-minute windows show resolved at least arrivals plus falling pending depth and oldest age.


hxy91819:
Superseded by #779, which extracted the minimal lease-cap and rolling-deploy safety fixes onto current main and passed the required checks before merge.

Issue #738, opened 2026-07-21
RFC: single-writer state materialization via the queue DO (retire git-lease concurrency)


steipete:
## Problem

`clawsweeper-state` is a git repo used as a high-contention multi-writer database: every run does fetch→merge→commit→push against one branch. We built database concurrency control on top of git refs — publish lease, writer admission (#724/#727), priority intent (#737), sweep-status/router-ledger merge arbitration, checkpoint chunking, shallow-recovery ladders (#712/#714/#716/#721/#730). The Jul 19–21 outage was this architecture failing five different ways: 30GB history, unavailable objects, orphaned roots, app-token expiry under hour-long publishes, lease starvation of the apply lane. History compaction (2026-07-20, backup at `backup/pre-compact-2026-07-20`) bought headroom; growth will reproduce it.

## Proposal

Writers stop touching git; the queue Durable Object becomes the single coordination point; git becomes a single-writer materialized view.

1. **WP-1 — DO state-append endpoint.** `POST /internal/state/append` (HMAC via existing webhook-secret pattern, idempotent by `delivery_id`), rows in DO SQLite as a bounded un-materialized window with backpressure shed to the existing artifact/reconcile fallback when the window is full.
2. **WP-2 — Materializer.** DO-alarm-driven dispatch (or 5-min schedule) drains pending rows into ONE fast-forward commit to `clawsweeper-state` per cycle; acks rows after push. Single writer ⇒ no lease, no merges, no contention. Proof/ledger entries batch through the same commit; DO holds only the un-materialized window (SQLite size cap from #519 stays respected).
3. **WP-3 — Cutover + deletion.** Writers cut over lane-by-lane (sweep statuses → router ledger → apply proofs), then delete the lease/admission/priority/merge machinery from `src/repair/git-publish.ts`.
4. **Guardrails (any outcome):** scheduled state-repo compaction with backup refs, scratch-branch janitor, repo-size line in the reconcile sweep.

## Coordination

The writer-coordination work (#724/#727) and this proposal touch the same subsystem. This issue is the convergence point — please flag conflicts here before landing further coordination changes. Implementation is proceeding on maintainer authorization; PRs will reference this issue.

steipete:
[...]

PR #811, opened 2026-07-23
fix(materializer): append recovery action events

vincentkoc:
Related: #738

## What Problem This Solves

Resolves a problem where each state materializer cycle could finish its bulk append drain and then immediately perform another direct Git commit for its recovery action-event ledger, adding avoidable contention to the state repository.

## Why This Change Was Made

The recovery action-event publisher now uses the existing immutable `apply_proof` append path. The normal queue failure fallback remains available, and the change introduces no new append kind, SQLite schema, or protocol surface.

## User Impact

Successful materializer cycles no longer follow the main drain with a second routine state-repository writer, reducing contention while preserving recovery evidence.

## Evidence

- State-writer workflow tests: 10/10 passed.
- Immutable ledger focused tests: 2/2 passed.
- Added a repository-wide invariant proving all seven `publish-action-event-paths` workflow surfaces enable state append.
- Formatting, lint, `git diff --check`, and structured autoreview passed with no actionable findings.

### DIFF
--- a/.github/workflows/state-materializer.yml
+++ b/.github/workflows/state-materializer.yml
@@ -107,6 +107,7 @@ jobs:
         if: ${{ always() }}
         env:
           CLAWSWEEPER_MODEL_RECOVERY_ENABLED: "0"
+          CLAWSWEEPER_STATE_APPEND_ENABLED: "1"
           CLAWSWEEPER_STATE_REPO_TOKEN: ${{ steps.state-token.outputs.token }}
           CLAWSWEEPER_WEBHOOK_SECRET: ${{ secrets.CLAWSWEEPER_WEBHOOK_SECRET }}
           QUEUE_URL: ${{ vars.CLAWSWEEPER_EXACT_REVIEW_QUEUE_URL || 'https://clawsweeper.openclaw.ai' }}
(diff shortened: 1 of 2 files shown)


vincentkoc:
Addressing the ClawSweeper rank-up move: the release-owned changelog edit is removed, and fresh autoreview is clean (`0.96`). I am not dispatching an extra branch-specific materializer while the production single-flight drain has one active run and two queued runs; that would add pressure to the exact path under recovery. The existing append failure fallback is unchanged and covered by the immutable publisher tests. I will verify the first post-merge materializer cycle publishes its recovery action event through the append window and preserves evidence.

vincentkoc:
Post-merge live proof on `main`:

- exact head: `9d8bd8d4d0863e4203edfe4486859fc76c4dd0ee`
- materializer run: https://github.com/openclaw/clawsweeper/actions/runs/30016558192
- result: success
- `Materialize queued state`: success
- `Finalize materializer recovery action ledger`: success
- `Publish materializer recovery action events`: success

The append window fell from roughly 28.4k rows at incident start to 136 rows during this run. The recovery publisher completed immediately after materialization instead of entering the old Git-writer recovery path.

PR #812, opened 2026-07-23
fix(queue): backpressure reviews on publication health

vincentkoc:
Related: #738

## What Problem This Solves

Fixes an issue where the exact-review queue reported `idle` and continued admitting background reviews while durable publication was critically backlogged. This let review production compound result-delivery debt.

## Why This Change Was Made

Publication health now elevates the public pressure signal: degraded publication produces soft pressure and critical publication produces hard pressure. Existing review-pressure severity still wins, and priority review lanes plus lease semantics are unchanged.

## User Impact

Background review admission yields while durable results are falling behind, allowing publication to recover without disabling interactive or priority review work. Operators also see the publication cause instead of a misleading idle badge.

## Evidence

- Main, repair, and dashboard TypeScript builds passed.
- 19 focused tests passed.
- Touched-file lint and `git diff --check` passed.
- Structured autoreview passed with no actionable findings after one severity-ordering fix.
- The unrelated dashboard type-aware lint baseline still has the same 32 failures on `origin/main`.

### DIFF
--- a/dashboard/exact-review-health.ts
+++ b/dashboard/exact-review-health.ts
@@ -53,12 +53,18 @@ export type ExactReviewPressureSummary = {
     | "no_admissible_backlog"
     | "dispatcher_inactive"
     | "handoff_unknown"
-    | "capacity_full_with_backlog";
+    | "capacity_full_with_backlog"
+    | "publication_degraded"
+    | "publication_critical";
   capacity: number;
   active: number;
   pending: number;
   ready_pending: number;
   admissible_pending: number;
+};
+
+export type ExactReviewPublicationHealth = {
+  status?: "idle" | "healthy" | "degraded" | "critical";
 };
 
 const PHASES: ExactReviewPhase[] = ["pending", "dispatching", "leased"];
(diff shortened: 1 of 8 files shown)


vincentkoc:
Resolved the exact ClawSweeper finding from the prior head: critical publication health now preserves an already-saturated review-pressure reason instead of overwriting it. Added a combined saturated+critical regression; `node --test test/exact-review-health.test.ts` passes 11/11, formatting and diff checks pass, and fresh autoreview is clean (`0.92`). The release-owned changelog entry is also removed.

vincentkoc:
Rank-up proof: the admission-path suites pass 89/89 on the rebased exact head. They explicitly prove critical publication health cannot leave pressure idle and that queue pressure scales background lanes while priority lanes remain unchanged. This covers the final ClawSweeper operational question without changing lease behavior.

vincentkoc:
Maintainer decision on the final ClawSweeper question: accept the conservative throttle. Live production currently has critical publication health, ~2,049 pending publication items, and effectively flat net drain even after the append window fell from ~28k to ~1.1k rows. The exact-head admission suites pass 89/89 and prove priority lanes remain unchanged. This is the intended incident control; I will verify the deployed dashboard reports `publication_critical` and background admission yields after merge.

PR #813, opened 2026-07-23
fix(repair): remove router Git writer pressure

vincentkoc:
Related: #738

## What Problem This Solves

Fixes an issue where every distinct `@clawsweeper` command entered the serialized state-repository writer even when routing completed in seconds and no repair job changed. During the incident, one router finished command processing in about 20 seconds, then waited over 22 minutes at writer queue position 33 before cancellation.

## Why This Change Was Made

Distinct maintainer commands remain independent and lossless. The durable processed-command ledger continues through the append/materializer path; the invocation-only latest scan report is no longer Git-published; and `jobs/` enters the Git writer only when a command actually changed durable repair work. The materializer retires the old tracked latest report on its next router batch.

## User Impact

Routine status and re-review commands no longer contend for the state Git writer after routing, while commands that genuinely create or update repair jobs retain durable publication.

## Evidence

- Live failure evidence: run 30010373110 routed successfully, then stalled in the state writer until cancellation.
- 75 focused tests passed.
- Main and repair TypeScript builds passed.
- Targeted formatting, lint, `git diff --check`, and structured autoreview passed with no actionable findings.
- No SQLite schema or append protocol change.

### DIFF
--- a/.github/workflows/repair-comment-router.yml
+++ b/.github/workflows/repair-comment-router.yml
@@ -292,12 +292,17 @@ jobs:
             echo "Exact terminal comment version already recorded; skipping state publication."
             exit 0
           fi
-          pnpm run repair:publish-main -- \
+          publish_args=(
             --message "chore: record ClawSweeper comment routing" \
             --path results/comment-router.json \
-            --path results/comment-router-latest.json \
-            --path jobs \
             --rebase-strategy theirs
+          )
+          # The latest scan report is run-local. Only commands that actually
+          # changed repair jobs need the serialized Git state writer.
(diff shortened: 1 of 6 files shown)


vincentkoc:
Addressing the ClawSweeper rank-up move: the release-owned changelog edit is removed, focused workflow/materializer coverage exercises both unchanged and changed `jobs/` paths, and fresh autoreview is clean (`0.87`). I am not creating a synthetic repair job in production solely to prove the changed-jobs branch while publication is critical. After merge I will verify the next no-change router run bypasses the Git writer and inspect the next real job-changing command for durable `jobs/` publication.

PR #595, opened 2026-07-15
fix(review): screenshot proof is available to sandboxed reviewers

goutamadwant:
Closes #594

## What Problem This Solves

Fixes an issue where contributors supplying screenshot-only real-behavior proof would have that evidence treated as unviewable by ClawSweeper's network-disabled review sandbox.

## Why This Change Was Made

ClawSweeper now hydrates linked PNG, JPEG, WebP, and GIF proof through the existing bounded media download path. Downloaded images are exposed directly for local inspection, while videos retain their existing ffprobe and contact-sheet processing. Local-range reviews continue skipping all host media downloads.

## User Impact

Screenshot-only evidence can now be assessed on its merits instead of being structurally classified as insufficient solely because the reviewer cannot access the network. Existing video proof behavior is unchanged.

## Evidence

- Before the fix, the focused regression could not discover or hydrate a PNG proof URL, while the identical MOV URL was recognized.
- `pnpm run build && node --test test/review-prompt-policy.test.ts`: 31/31 passed. Coverage includes image discovery, image artifact preparation, no video-tool invocation for images, mixed image/video ordering, preserved video-only hook behavior, and existing ffprobe/contact-sheet processing.
- Real production-helper proof on macOS 26.0.1 with Node 24.16.0: served the repository's tracked PNG over a localhost HTTP server and ran the built media preparation helper with the real curl command. Result: `status=prepared`, `kind=image`, `downloaded=true`, `bytes=1478360`, `metadataCreated=false`, `contactSheetCreated=false`, `manifestArtifacts=1`; the server recorded an HTTP 200 GET.
- `pnpm run check`: passed, including build, lint, 1,180 unit tests with 0 failures, repair tests, changed and full coverage suites, and formatting.
- Limitation: no live GitHub-hosted ClawSweeper review decision was run; validation covers the production media hydration boundary, generated manifest, prompt guidance, and full local test suite.

Disclosure: AI was used to understand the codebase and review the fix.

Issue #594, opened 2026-07-15
Screenshot proof is never hydrated, so realBehaviorProof is always "insufficient" for image-only evidence


anagnorisis2peripeteia:
I ran into this while using `Bash`: The review prompt says screenshots are valid real-behavior proof and that the reviewer may open linked artifacts from the provided scratch directory. But prepareMediaProofArtifacts() hydrates video URLs only, so linked images are never downloaded into proofScratchDir, and the reviewer runs sandbox read-only where sandboxPolicy() hardcodes networkAccess:false - so it cannot fetch them itself either. The reviewer therefore cannot view screenshot proof and, following the prompt's own rule ("insufficient when the evidence is ... unviewable"), rates realBehaviorProof insufficient with needsContributorAction=true. Video-proven PRs are unaffected. This is not local-only: .github/workflows/sweep.yml runs the review job with --codex-sandbox read-only.

I expected Linked screenshots should be hydrated into proofScratchDir (already passed to codex via --add-dir) so the read-only reviewer can open them locally, exactly as linked videos are today - and screenshot-proven PRs should be assessable on their merits.
Instead, Image URLs are filtered out before hydration, so nothing is downloaded. The sandboxed reviewer has neither a local copy nor network access, cannot view the evidence, and rates realBehaviorProof=insufficient / needsContributorAction=true no matter how good the screenshots are.

I traced this to proofVideoUrlsFromContext() keeps only URLs ending in VIDEO_PROOF_EXTENSIONS, so prepareMediaProofArtifacts() returns early for image-only proof and writes nothing to proofScratchDir. Since the reviewer runs sandbox read-only and sandboxPolicy() hardcodes networkAccess:false for that mode, the model has neither a local copy nor network, making screenshot proof structurally unviewable.
I verified that with Red-green on a clean clone of upstream main at c17589a56c, using the project's own exported test hook and test helpers: a .png proof URL returns [] from proofVideoUrlsFromContextForTest (RED - never hydrated) while an otherwise identical .mov URL returns the URL (GREEN control - hydrated). Only the extension differs. Result: 2 pass / 0 fail.

Reproduction: Clean clone of openclaw/clawsweeper at c17589a56c, then pnpm install && pnpm run build. Add test/zz-repro.test.ts mirroring the existing video assertion in test/review-prompt-policy.test.ts:274, changing only the file extension, and run: node --test test/zz-repro.test.ts

import assert from "node:assert/strict";
import test from "node:test";
[...]

PR #814, opened 2026-07-23
fix(review): hydrate .avif/.bmp screenshot proof and cover the failed-download path

anagnorisis2peripeteia:
Applies the two inline review suggestions left on #595 that weren't picked up before merge:

1. Add `.avif` and `.bmp` to `IMAGE_PROOF_EXTENSIONS` — without them those formats are never collected/hydrated for sandboxed reviewers, so the #594 symptom persists for `.avif`/`.bmp`.
2. Add a failure-path test so a regression that marks a failed screenshot download as `prepared` can't slip through (mirrors the existing success test; the impl at `src/clawsweeper.ts:9816` emits `status:"failed"` + `"download failed: …"`).

Stacked on #595 — its `IMAGE_PROOF_EXTENSIONS` is the base for change 1, so this PR currently also shows #595's diff. Once #595 merges I'll rebase onto `main` so it shows only these two changes.

Related: #595, #594


steipete:
Merged — thanks @anagnorisis2peripeteia for circling back on the #595 suggestions; the failure-path test closes a real regression window. (I resolved the post-#595 stack conflicts on your branch to get it green.)

PR #822, opened 2026-07-24
fix(publication): prepare batches concurrently

vincentkoc:


vincentkoc:
Maintainer decision: accept the bounded production envelope of four concurrent publication batches, eight items per batch, and four-way preparation.

Rank-up moves:

- Owner confirmation: accepted. Four batches is the intended incident-recovery cap; the queue remains the admission authority and the final state-writer mutation stays serialized.
- Overlapping live evidence before merge: deferred to the immediate post-merge deployment proof because the new admission cap cannot produce overlapping production leases until this Worker revision is deployed. After deploy I will verify four disjoint active batch leases, successful batch acknowledgements, no new dead letters, and serialized state-writer completion. Any duplicate ownership, stranded lease, contention regression, or DLQ growth is a rollback condition.

Current capacity evidence shows the bottleneck is the single publication batch, not GitHub or Blacksmith capacity. The change is bounded at four and clamped in code.

PR #824, opened 2026-07-24
improve(publication): raise concurrent batch owners to eight

vincentkoc:
## What Problem This Solves

Reduces the remaining exact-review publication backlog faster after the four-owner rollout proved stable but still left roughly 1,500 stale publication rows draining over several hours.

## Why This Change Was Made

Raises only the isolated publication preparation-owner cap from four to eight. Each owner remains bounded to eight items, final mutations still cross the single state-writer coordinator, and the cap remains independently reversible in Worker configuration.

## User Impact

Maintainers should see the historical publication backlog drain closer to the current arrival rate without increasing per-batch mutation size or changing review behavior.

## Evidence

- Four-owner live cohort: consecutive overlapping batches completed successfully with zero retries, zero released items, and zero new dead letters.
- Queue at July 24, 2026 03:47 UTC: 1,517 pending; 136/hour 15-minute net drain; four leases / 32 active items; DLQ unchanged at 614.
- `node --test test/state-writer-workflow.test.ts` (10/10)
- focused `oxfmt --check` and `git diff --check` pass
- autoreview clean, no accepted/actionable findings

The broad local `pnpm run check` was not run because it attempted to reconcile the shared worktree install; CI owns that gate.


vincentkoc:
Maintainer decision: keep the per-claim scan at 50 and raise only concurrent owners to eight. The current exact-review finding conflates two independent limits.

Why the scan does not cap total active items at 50:
- `claimPublicationBatch` computes `leaseSize = min(requestedSize, configuredSize)`, so each dispatched workflow leases at most 8 items even though it scans up to 50 candidates.
- Every later claim excludes item keys already owned by active batches.
- `maxConcurrentBatches` independently caps the number of leased workflows; raising it to 8 permits eight disjoint 8-item leases.
- The queue alarm dispatches one workflow, then refills another slot after the claim consumes the reservation; the 50-row scan is repeated per workflow, not shared globally.
- Live proof with the same 50-row scan already reached four simultaneous leases / 32 active items, confirming the model.

Therefore eight owners can reach 64 active items across eight independent claims without changing the per-claim 50-candidate owner-homogeneity scan. I am rejecting the suggested 64 scan change as unnecessary extra admission work. Post-deploy proof will verify eight leases, retry rate, DLQ delta, and writer health before keeping the ramp.

PR #835, opened 2026-07-24
fix(review): escape model prose before it reaches report sections

yetval:
## Summary

A contributor could get an unproven external PR routed into the automated repair lane by putting a fake `## Real Behavior Proof` block in the PR body. When the review model echoed that text into the report summary, the report parser read the forged block instead of the real one and the PR's proof verdict flipped from blocked to sufficient.

This patch escapes model-authored prose on the way into the durable report artifact, so injected heading-shaped text can no longer create a section boundary that downstream parsers will trust.

## Root cause

`sectionValue` in `src/clawsweeper.ts` is a single non-global regex match:

```ts
function sectionValue(markdown: string, heading: string): string {
  const match = markdown.match(
    new RegExp(`(?:^|\\n)## ${heading}\\n\\n([\\s\\S]*?)(?=\\n## |\\n?$)`),
  );
  return match?.[1]?.trim() ?? "";
}
```

The first `\n## <heading>\n\n` anywhere in the report wins. `markdownFor` emits `## Summary` eleven sections before `## Real Behavior Proof`, so any `## Real Behavior Proof` block sitting inside the summary body is read first.

`parseDecision` sanitized only two of its model-authored fields, `systemContext` and `architectureDiagram`. Every other prose field, including `summary`, was a raw `requireString` and reached the report verbatim.

The consequences chained: `reportRealBehaviorProof` returned `sufficient`, the `proof: sufficient` label was applied, `realBehaviorProofBlocksMerge` returned false, `derivedPrRating` raised the proof tier, and `reviewAutomationMarkersFromReport` emitted `clawsweeper-verdict:needs-changes` plus `clawsweeper-action:fix-required` instead of `clawsweeper-verdict:needs-human`. The forgery works bare, inside a fence, inside `<details>`, and inside an HTML comment.

## Fix

1. Moved `OWNED_REVIEW_SECTION_HEADINGS` and `neutralizeOwnedSectionSpoofing` verbatim from `src/clawsweeper.ts` into the zero-import leaf `src/clawsweeper-text.ts` so `src/decision-packets.ts` can use them without importing `clawsweeper.ts`.
2. Added `requireReportText` and `requireReportTextArray` next to `requireString`.
3. Applied them to every model-authored prose field that lands in a report section body: `parseDecision` (summary, changeSummary, systemContext, risks, bestSolution, reproductionAssessment, solutionAssessment, visionFitReason, visionFitEvidence, closeComment, workReason, workPrompt), and the sub-parsers for evidence, likely owners, review findings, security review and its concerns, real behavior proof, PR rating, Telegram visible proof, Mantis recommendation reason, feature showcase, root-cause cluster, and AGENTS.md policy status. `parseMaintainerDecision` and its option and owner parsers in `src/decision-packets.ts` get the same treatment.
4. Left code-span and fence fields raw, because the escaper rewrites `<` to `&lt;` and that renders literally inside backticks: `evidence[].command`, `mantisRecommendation.maintainerComment`, `mergeRiskOptions[].automergeInstruction`, `workLikelyFiles`, `workValidation`, `workClusterRefs`, `reviewFinding.file`, `securityConcern.file`, and `architectureDiagram`, which keeps its existing `sanitizeArchitectureDiagram` allowlist. Each renderer was read before the field was classified.
5. Made front matter win over section lines in `reportRealBehaviorProof` and `reportPrRating`. Write-side escaping cannot repair reports already published in `openclaw/clawsweeper-state`, and those are re-read every sweep cycle. The renderer emits front matter and section body from the same `Decision`, so the two can only disagree when the section has been tampered with.

Write-side escaping is already this codebase's load-bearing defense. `extractMarkdownSection` in `src/repair/comment-router-core.ts` is safe in production only because its input, the rendered review comment, has `neutralizeOwnedSectionSpoofing` applied on the way in. The gap was that the durable report artifact never got the same treatment. This patch extends an existing pattern one layer earlier rather than adding a second, differently-shaped defense.

Hardening the reader was considered and rejected. Delegating `sectionValue` to `extractMarkdownSection` does not fix the primary attack, because that helper also takes the first top-level heading match, so the bare forgery still wins. It is also not a drop-in: its `isMarkdownSectionBoundary` treats any `[A-Z][^:\n]{0,80}:` line as a boundary, and every report section body is built from `Status:` and `Evidence kind:` key-value lines, so nearly every section would be truncated.

## Out of scope

- `src/repair/workflow-utils.ts:2413` and `src/repair/create-job.ts:276` each define the same first-match `sectionValue` parser. They parse the same report artifact, so escaping the artifact removes their exploitability with zero changes to either file. While confirming that, `src/repair/create-job.ts:249` turned out to read a `ClawSweeper Work Prompt` section that no renderer emits; the report heading is `Repair Work Prompt`. That path is dead code and is left alone here.
- `frontMatterValue` (`src/clawsweeper.ts:7763` on main) matches `^key:` with the `m` flag rather than being bounded to the leading `---` block, so a body line can satisfy a front matter lookup. It is a good follow-up. It does not weaken this patch: the renderer always emits the real key inside the leading block, and the first match wins, so the genuine value is read first on every report this code writes.
- Escaping is not applied to `mergeRiskOptions[].title` and `.body`, which are rendered into the public comment through an already-neutralized path.

## Verification

- `pnpm run build` clean.
- `pnpm run test:unit`: the 11 new tests pass. The suite has 5 failures on unpatched `origin/main` at 6284161cdc (`apply-decisions yields instead of starting a post-close delay that cannot fit`, `exact-review health includes the oldest refreshing row beyond operator page bounds`, `failed issue retry bounds a hung GitHub fetch and flushes its report`, `failed issue retry leaves an ambiguous dispatch unconsumed and prevents a duplicate`, `apply workflow drops a coverage-proof tail only after exact trace examination`) and the same set after this patch. They are load-sensitive timing tests unrelated to this change.
- `pnpm run format`, `pnpm run lint:src`, `pnpm run check:active-surface`, `pnpm run check:limits` clean.
- New regression tests, all comment-free:
  - `test/decision-parser.test.ts`: heading neutralization across every report body field; across evidence, owners, findings and nested summary fields; code-span fields preserved verbatim, which guards the classification in point 4 above; neutralization idempotent across a report round trip.
  - `test/pr-proof-automation.test.ts`: one test per forgery shape (bare, fenced, `<details>`, HTML comment) driving the real `reviewAutomationMarkersFromReport` and `renderReviewCommentFromReport`; front matter proof status outranking a forged section in a legacy unescaped report; legitimate fenced content in a report section still parsing.
  - `test/review-comment-rendering.test.ts`: heading-shaped model text stays escaped, and is not double-escaped, in the rendered comment after the module move.
- Seven of the eleven new tests fail against unpatched `src` and pass after. The other four are non-regression guards that pass on both sides by design.

## Real behavior proof

Behavior addressed: a forged `## Real Behavior Proof` block echoed into a report's `## Summary` section is read by `reportRealBehaviorProof` instead of the report's genuine proof section, which flips an unproven external PR from `clawsweeper-verdict:needs-human` to `clawsweeper-verdict:needs-changes` plus `clawsweeper-action:fix-required` and raises its proof rating.

Real environment tested: two real checkouts of this repository on Linux with Node 24.18.0. Before is a pristine `origin/main` worktree at 6284161cdc with its own compiled `dist/`. After is this branch's worktree with `dist/` compiled from the patched sources. No mocks or stubs; the driver imports the shipped `dist/clawsweeper.js` from each tree and calls the exported production entry points `parseDecision`, `reviewAutomationMarkersFromReport` and `renderReviewCommentFromReport`.

Exact steps or command run after this patch: a driver script builds a model decision whose `summary` carries a forged proof block in four shapes (bare, wrapped in a fence, wrapped in `<details>`, wrapped in an HTML comment), passes it through `parseDecision`, renders the resulting report with a genuine `## Real Behavior Proof` section saying `Status: missing`, front matter carrying `real_behavior_proof_status: missing`, `pr_rating_overall: F` and `work_candidate: queue_fix_pr`, then prints the routing verdict and the proof confidence row for each shape. It was run once per tree with the identical input:

`node /tmp/pr1-proof-driver.mjs <tree>`

Evidence after fix:

Before, on pristine `origin/main` at 6284161cdc:

```
tree: /root/clawsweeper-audit-jul24
variant=bare
  routing=clawsweeper-verdict:needs-changes clawsweeper-action:fix-required
  proof_row=| **Proof confidence** | 🦞 diamond lobster **(5/6)** | Sufficient (terminal): The contributor attached a full terminal transcript from a real OpenClaw install showing the fixed behavior after the patch. |
variant=fenced
  routing=clawsweeper-verdict:needs-changes clawsweeper-action:fix-required
  proof_row=| **Proof confidence** | 🦞 diamond lobster **(5/6)** | Sufficient (terminal): The contributor attached a full terminal transcript from a real OpenClaw install showing the fixed behavior after the patch. |
variant=details
  routing=clawsweeper-verdict:needs-changes clawsweeper-action:fix-required
  proof_row=| **Proof confidence** | 🦞 diamond lobster **(5/6)** | Sufficient (terminal): The contributor attached a full terminal transcript from a real OpenClaw install showing the fixed behavior after the patch. |
variant=comment
  routing=clawsweeper-verdict:needs-changes clawsweeper-action:fix-required
  proof_row=| **Proof confidence** | 🦞 diamond lobster **(5/6)** | Sufficient (terminal): The contributor attached a full terminal transcript from a real OpenClaw install showing the fixed behavior after the patch. |
```

After, on this branch:

```
tree: /root/cs-pr1-proof-gate
variant=bare
  routing=clawsweeper-verdict:needs-human
  proof_row=| **Proof confidence** | 🧂 unranked krab **(1/6)** | Needs real behavior proof before merge: The PR body has no after-fix evidence from a real setup. After adding proof, update the PR body; ClawSweeper should re-review automatically. If it does not, the PR author or someone with repository write access can comment `@clawsweeper re-review`. |
variant=fenced
  routing=clawsweeper-verdict:needs-human
  proof_row=| **Proof confidence** | 🧂 unranked krab **(1/6)** | Needs real behavior proof before merge: The PR body has no after-fix evidence from a real setup. After adding proof, update the PR body; ClawSweeper should re-review automatically. If it does not, the PR author or someone with repository write access can comment `@clawsweeper re-review`. |
variant=details
  routing=clawsweeper-verdict:needs-human
  proof_row=| **Proof confidence** | 🧂 unranked krab **(1/6)** | Needs real behavior proof before merge: The PR body has no after-fix evidence from a real setup. After adding proof, update the PR body; ClawSweeper should re-review automatically. If it does not, the PR author or someone with repository write access can comment `@clawsweeper re-review`. |
variant=comment
  routing=clawsweeper-verdict:needs-human
  proof_row=| **Proof confidence** | 🧂 unranked krab **(1/6)** | Needs real behavior proof before merge: The PR body has no after-fix evidence from a real setup. After adding proof, update the PR body; ClawSweeper should re-review automatically. If it does not, the PR author or someone with repository write access can comment `@clawsweeper re-review`. |
```

Observed result after fix: all four forgery shapes now route to `clawsweeper-verdict:needs-human` with no `clawsweeper-action:fix-required` marker, and the proof confidence row reports the genuine missing proof at `🧂 unranked krab (1/6)` instead of the forged `🦞 diamond lobster (5/6)`. The rendered comment shows the injected block escaped in the summary body, for example `&lt;!--` and `\## Real Behavior Proof`, so it is visible to a human reviewer but inert to the parser. Before the patch every shape produced the same wrong result, which shows the fence, `<details>` and HTML-comment wrappers were not defenses.

What was not tested: no live GitHub API calls, so the label write and PR comment update paths were exercised only through the renderer, not against a real repository. Reports already published in `openclaw/clawsweeper-state` were not replayed; the front matter precedence change is what covers them and it was verified only on a synthetic legacy report. The repair worker was not run end to end, so the effect on an already-queued job was not observed. Non-Linux platforms were not exercised.

### DIFF
--- a/src/clawsweeper-text.ts
+++ b/src/clawsweeper-text.ts
@@ -1,3 +1,69 @@
+export const OWNED_REVIEW_SECTION_HEADINGS = new Set([
+  "summary",
+  "what this changes",
+  "merge readiness",
+  "review scores",
+  "verification",
+  "how this fits together",
+  "decision needed",
+  "before merge",
+  "next step",
+  "next step before merge",
+  "automerge follow-up",
+  "autofix follow-up",
+  "findings",
+  "review findings",
+  "security",
+  "label changes",
+]);
+
+// Model-generated text is rendered above renderer-owned sections such as
+// "## Before merge", and downstream routing extracts those sections from the first
+// matching Markdown heading. Escape heading-shaped lines in model text so injected
+// content can never spoof a renderer-owned section boundary.
+export function neutralizeOwnedSectionSpoofing(value: string): string {
+  // GitHub normalizes CRLF and bare CR to line endings, so normalize first or a
+  // bare-CR line break could smuggle a heading past the per-line checks.
+  return value
+    .replace(/\r\n?/g, "\n")
+    .split("\n")
+    .map((line) => {
+      // Strip blockquote/list container prefixes so nested heading constructs are
+      // neutralized too.
+      // CommonMark accepts blockquotes without a following space and ordered lists
+      // with either "1." or "1)".
+      const containerPrefix =
+        line.match(/^[ \t]*(?:(?:>|(?:[-*+]|\d+[.)])[ \t])[ \t]*)*/)?.[0] ?? "";
+      // Escape every raw HTML delimiter (renderer-emitted <br> excepted) so inline
+      // tags and comment openers cannot restructure or hide trusted sections;
+      // &lt; renders identically to a literal <.
+      const content = line.slice(containerPrefix.length).replace(/<(?!br\s*\/?>)/gi, "&lt;");
+      const trimmed = content.trim();
+      if (/^#{1,6}\s+\S/.test(trimmed)) {
+        return `${containerPrefix}${content.replace("#", "\\#")}`;
+      }
+      if (/^\*\*[^*\n]+\*\*:?\s*$/.test(trimmed)) {
+        return `${containerPrefix}${content.replace("**", "\\*\\*")}`;
+      }
+      if (/^(?:```|~~~)/.test(trimmed)) {
+        return `${containerPrefix}${content.replace(/[`~]/, "\\$&")}`;
+      }
+      // A run of = or - alone on a line is a Setext underline that would promote the
+      // previous line to a heading.
+      if (/^(?:=+|-+)[ \t]*$/.test(trimmed)) {
+        return `${containerPrefix}${content.replace(/[=-]/, "\\$&")}`;
+      }
+      if (
+        trimmed.endsWith(":") &&
+        OWNED_REVIEW_SECTION_HEADINGS.has(trimmed.slice(0, -1).trim().toLowerCase())
+      ) {
+        return `${containerPrefix}${content.trimEnd().slice(0, -1)}&#58;`;
+      }
+      return `${containerPrefix}${content}`;
+    })
+    .join("\n");
+}
+
 export function escapeRegExp(value: string): string {
   return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -168,7 +168,13 @@ import {
   stringArg,
   type Args,
 } from "./clawsweeper-args.js";
-import { escapeRegExp, safeOutputTail, trimMiddle, truncateText } from "./clawsweeper-text.js";
+import {
+  escapeRegExp,
+  neutralizeOwnedSectionSpoofing,
+  safeOutputTail,
+  trimMiddle,
+  truncateText,
+} from "./clawsweeper-text.js";
 import {
   emptyMaintainerDecision,
   maintainerDecisionBlocksClose,
@@ -3422,6 +3428,14 @@ function requireStringArray(value: unknown, path: string): string[] {
   return value.map((entry, index) => requireString(entry, `${path}[${index}]`));
 }
 
+function requireReportText(value: unknown, path: string): string {
+  return neutralizeOwnedSectionSpoofing(requireString(value, path));
+}
+
+function requireReportTextArray(value: unknown, path: string): string[] {
+  return requireStringArray(value, path).map(neutralizeOwnedSectionSpoofing);
+}
+
 function requireEnumArray<T extends string>(value: unknown, allowed: Set<T>, path: string): T[] {
   return requireStringArray(value, path).map((entry, index) =>
     requireEnum(entry, allowed, `${path}[${index}]`),
@@ -3604,8 +3618,8 @@ function parseEvidence(value: unknown, path: string): Evidence {
   const record = requireRecord(value, path);
   rejectUnexpectedKeys(record, EVIDENCE_SCHEMA_KEYS, path);
   return {
-    label: requireString(record.label, `${path}.label`),
-    detail: requireString(record.detail, `${path}.detail`),
+    label: requireReportText(record.label, `${path}.label`),
+    detail: requireReportText(record.detail, `${path}.detail`),
     file: requireNullableString(record.file, `${path}.file`),
     line: requireNullableInteger(record.line, `${path}.line`),
     command: requireNullableString(record.command, `${path}.command`),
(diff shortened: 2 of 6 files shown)


yetval:
@codex review

brokemac79:
@yetval  - can i ask if you can try and hold off creating new PRs on clawsweeper at the moment.  Please monitor discord channels.  We hope to have things in a stable state in the next 24 hours or so.  

@Takhoffman - please try and refrain from merging any PRs if possible.  Message me on discord when you get a chance. I have messaged you, but you seem offline.  

In summary - the clawsweeper issues in the last few days have had various AI agents proclaim to 'fix' things and haven't .  About 80% have probably made things worse, or not fixed underlying issues and adding more noise is creating extra code that isn't necessary or helping right now.  

cc: @vincentkoc

yetval:
@brokemac79 My bad thats on me. Ill hold off on the PR's, sorry for the noise

brokemac79:
> @brokemac79 My bad thats on me. Ill hold off on the PR's, sorry for the noise

no problem - if possible, please set this one to draft, otherwise  a maintainers clanker may pick it up and just merge. :-)

yetval:
@clawsweeper re-run

yetval:
Closing this per the request to hold off on contributor PRs while things stabilize. The underlying defect and everything learned here (root cause, repro against production entry points, the legacy-reader gap flagged in review, and adjacent surfaces) is now captured as issue #951 so it is not lost. The branch yetval:fix/report-section-spoofing (head 7e6f81f27872) remains available if a maintainer wants to take it over or reference it.

PR #837, opened 2026-07-24
fix(review-placeholder): rank recovery discovery before the check budget

yetval:
## Summary

The review-placeholder recovery sweep is the only backstop for the "ClawSweeper status: review started." comments that ClawSweeper posts on every lease acquisition and never edits. Right now that backstop cannot see most of what it is supposed to clean up: openclaw/openclaw currently carries roughly 1,200 stranded placeholders across 586 open pull requests, 616 open issues, and 3,288 closed items, and a sample of 60 recently reviewed pull requests turned up exactly one leftover placeholder on every single one.

This PR fixes three independent defects in `src/review-placeholder-recovery.ts`:

- discovery applied the check budget before any ranking, so only the newest 20 matches were ever reachable;
- placeholder detection looked at a single comment per item, so an edited durable verdict hid the placeholder permanently;
- the job failure predicate could not fire on the shape the sweep actually produces, so a standing backlog reported success every 15 minutes.

## Root cause

**Discovery truncated before it ranked.** `searchCandidates` queried `/search/issues` with `sort=updated&order=desc` and broke out of the item loop at `added >= maximumChecks` (default 20). The oldest-first ordering landed by #816 (`orphanedCandidates.sort((a, b) => a.createdAtMs - b.createdAtMs)`) runs later and only over the survivors of that truncation. #816 moved the `enqueued >= maximumRecoveries` break out of the inspection loop and fixed the recoveries budget of 5; it did not touch discovery. Because the break happened in the middle of page 1, `SEARCH_MAX_PAGES = 2` was dead at default settings and the oldest stranded placeholders were structurally unreachable.

**Only one comment per item was inspected.** `fetchLatestBotComment` returned `latestClawSweeperBotComment(comments)`, which picks the single newest ClawSweeper comment ranked by `commentCreatedAtMs`, that is `max(created_at, updated_at)`. `upsertReviewComment` PATCHes the durable verdict comment in place, so a verdict written before the placeholder and edited after it outranks the placeholder forever. The item still consumed a check slot (`checked += 1` runs before inspection) and was then dropped.

**The failure predicate never fired.** `reviewPlaceholderRecoveryFailureReason` required `orphaned > 0 && errors > 0 && enqueued + cleaned === 0`, a three-way AND. The dominant real shape is orphaned small, errors zero, at least one enqueue, which returns null. The summary also had no input describing truncation or backlog: `checked` was never consulted, and the `total_count` that every `/search/issues` response already returns was discarded.

## Fix

- `searchCandidates` now queries `order=asc`, accumulates whole pages in the `ghPagedLimit` loop shape (`page <= SEARCH_MAX_PAGES && found.length < maximumChecks`, no mid-page break), ranks the accumulated candidates by item `updated_at` ascending with a stable index tiebreak, and only then fills the per-state check budget. Each state class keeps its own budget as before.
- `selectReviewPlaceholderComment(number, comments)` replaces `latestClawSweeperBotComment`. It matches bot comments on the machine marker `<!-- clawsweeper-review-status:started item=N ` (the same string `deletePlaceholderComment` already gates on, now shared through `reviewStartStatusMarker`), mirroring `selectDedicatedReviewStartLeaseComments`. Marker matches win; the placeholder sentence remains a fallback for comments written before the marker existed. This also removes the `max(created_at, updated_at)` promotion hazard, because the pool no longer contains the durable verdict.
- Comment paging goes from 2 to 5 pages with an early exit as soon as a page contains the marker, so a long thread no longer puts the placeholder out of reach while normal threads still cost one request.
- `ReviewPlaceholderRecoverySummary` gains `matched` (the `total_count` the search already returned) and `remaining` (`max(0, matched - checked)`), both printed on the existing summary line. `reviewPlaceholderRecoveryFailureReason` takes an optional threshold and fails the run when `remaining` reaches it. The default of 480 is one full day of best-case drain: 96 sweeps per day at `DEFAULT_REVIEW_PLACEHOLDER_MAX_RECOVERIES` of 5. A remainder above that cannot be worked off by the cadence at all.
- `REVIEW_PLACEHOLDER_LOOKBACK_HOURS` now goes through the existing `boundedPositiveInteger` helper (default unchanged at 48 hours, ceiling one year) and is exposed in `.github/workflows/exact-review-reconcile.yml` together with `REVIEW_PLACEHOLDER_BACKLOG_ALERT`, so both can be widened by repository variable without a code change.

`matched` counts search matches rather than confirmed orphans, so it is an upper bound: an item whose placeholder was already removed can linger in the search index. That is the correct direction for an alert threshold, and `orphaned` remains the precise number.

The placeholder design itself is untouched. A fresh comment per lease acquisition is deliberate (`src/clawsweeper.ts:21501`), and the durable verdict is a separate comment maintained by `upsertReviewComment`. Making the placeholder promise literally true would mean changing how leases publish, which is a much larger change with review-delivery risk. This PR repairs only the backstop that is supposed to converge the difference, and it does so without new state: no cursor, no schema, no extra API surface beyond a response field that was already being fetched and thrown away.

The request budget barely moves. Discovery still issues one search request per state class at default settings, and inspection still visits at most `maximumChecks` items per class. Comment paging only reaches past its first page on threads longer than 100 comments where the marker has not been seen yet, which is the exact case the old two-page ceiling could not resolve.

## Out of scope

- Editing the placeholder in place, or removing the public promise in its body. That is a lease-publication decision, not a recovery bug.
- Draining the pre-2026-07-22 backlog. About 524 open items carrying the marker sit outside the 48 hour lookback and stay invisible even after this PR. Reaching them properly needs date-window partitioning around the 1000-result pagination ceiling on `/search/issues`, which in turn needs resumable sweep state so consecutive runs advance through the windows instead of restarting at the same one. This repository has no such cursor to reuse: `src/review-activity-cursor.ts` is specific to GraphQL connection cursors for a single item's activity and does not generalize to a repository-wide date-partitioned scan. Building that belongs in its own change with its own proof, so this PR only makes the window operator-tunable and makes the resulting backlog visible instead of silent.
- The `cleanupSupersededReviewPlaceholderComments` call sites in `src/clawsweeper.ts`. They are unchanged.

## Verification

- `pnpm run build`
- `pnpm run test:unit`
- `pnpm run format`
- `pnpm run lint:src`

`test/review-placeholder-recovery.test.ts` goes from 12 to 16 cases. Nine summary assertions were extended with the two new fields. Two cases were rewritten because they encoded the defects being fixed: the one asserting that detection considers only the latest bot comment now asserts marker-based selection in the presence of an edited newer verdict, and the failure-reason case now covers the backlog branch as well as the total-action-failure branch. Three cases were added: rank-before-truncate (an item outside the newest-20 window is discovered and inspected first), an edited durable verdict no longer masking a marker-tagged placeholder, and the backlog signal reaching both the summary line and the failure reason. The oldest-orphans-first case from #816 and the closed-cleanup budget case are preserved unchanged apart from the new summary fields.

One pre-existing failure in `test/sweep-workflow.test.ts` ("apply workflow drops a coverage-proof tail only after exact trace examination") reproduces identically on unmodified `origin/main` at 6284161cdc and is unrelated to this change.

## Real behavior proof

Behavior addressed: the 15-minute recovery sweep could not discover stranded review placeholders outside the newest 20 search results, silently skipped any placeholder whose durable verdict comment had been edited more recently, and reported success while a large backlog of placeholders went untouched.

Real environment tested: the compiled recovery module `dist/review-placeholder-recovery.js` executed on Node 24 on Linux, at pristine `origin/main` 6284161cdc and at this branch head c890f8fa39, driven end to end through its exported `runReviewPlaceholderRecovery` entry point against GitHub REST responses reproducing the current openclaw/openclaw placeholder shape: 24 open items inside the 48 hour window (17 of the 20 newest-updated carrying a placeholder, only #906 both past the 2 hour orphan threshold and unmasked, #913 past the threshold but carrying a durable verdict edited 10 minutes ago, #922 stranded for 30 hours and sitting outside the newest-20 window) and 3,288 closed items carrying the marker.

Exact steps or command run after this patch:

```
RPR_DIST=/root/clawsweeper-audit-jul24/dist/review-placeholder-recovery.js node /tmp/pr3-proof-driver.mjs
RPR_DIST=/root/cs-pr3-placeholders/dist/review-placeholder-recovery.js node /tmp/pr3-proof-driver.mjs
```

Evidence after fix:

BEFORE (origin/main 6284161cdc):

```
review-placeholder recovery: enqueued #906 (issue)
review-placeholder recovery: checked=40 orphaned=1 enqueued=1 cleaned=0 escalated=0 errors=0
SEARCH REQUESTS: is:open order=desc page=1 | is:closed order=desc page=1
ITEMS INSPECTED (open): 901,902,903,904,905,906,907,908,909,910,911,912,913,914,915,916,917,918,919,920
ENQUEUED: 906
ESCALATED: (none)
SUMMARY OBJECT: {"checked":40,"orphaned":1,"enqueued":1,"cleaned":0,"escalated":0,"errors":0}
FAILURE REASON: (none, run reported success)
```

AFTER (this branch, head c890f8fa39):

```
review-placeholder recovery: escalated #922 as stuck after repeated orphan cycles
review-placeholder recovery: enqueued #922 (issue)
review-placeholder recovery: enqueued #913 (issue)
review-placeholder recovery: enqueued #906 (issue)
review-placeholder recovery: checked=40 orphaned=3 enqueued=3 cleaned=0 escalated=1 errors=0 matched=3312 remaining=3272
SEARCH REQUESTS: is:open order=asc page=1 | is:closed order=asc page=1
ITEMS INSPECTED (open): 924,923,922,921,920,919,918,917,916,915,914,913,912,911,910,909,908,907,906,905
ENQUEUED: 922,913,906
ESCALATED: 922
SUMMARY OBJECT: {"checked":40,"orphaned":3,"enqueued":3,"cleaned":0,"escalated":1,"errors":0,"matched":3312,"remaining":3272}
FAILURE REASON: 3272 matching placeholders were left unexamined (backlog alert threshold 480)
```

Observed result after fix: with an identical GitHub-side shape and an identical inspection budget of 20 items per state class, the run recovers 3 stranded placeholders instead of 1. #922, stranded for 30 hours and outside the newest-20 window, is discovered for the first time, recovered first because it is the oldest, and labelled `clawsweeper-recovery-stuck`. #913, whose durable verdict had been edited 10 minutes earlier, is no longer hidden behind that edit and is recovered. #906 is still recovered. The run now reports `matched=3312 remaining=3272` and returns a failure reason, so the scheduled job exits non-zero instead of reporting success while 3,272 matching placeholders sit untouched.

What was not tested: no request was issued against api.github.com, so live GitHub rate limiting, `/search/issues` secondary limits, and the real 1000-result pagination ceiling are not exercised here. The API responses were assembled to match the counts and comment layout observed on openclaw/openclaw rather than captured verbatim from a live sweep run. The workflow step exit code was observed in process through the same predicate the CLI guard uses, not by dispatching `.github/workflows/exact-review-reconcile.yml`. Draining the pre-2026-07-22 backlog outside the 48 hour lookback is deliberately not addressed, and the widened lookback was exercised only through the configuration path, not against a real multi-day window.

### DIFF
--- a/.github/workflows/exact-review-reconcile.yml
+++ b/.github/workflows/exact-review-reconcile.yml
@@ -211,6 +211,8 @@ jobs:
           GH_TOKEN: ${{ github.token }}
           TARGET_WRITE_TOKEN: ${{ steps.target-write-token.outputs.token }}
           QUEUE_URL: ${{ vars.CLAWSWEEPER_EXACT_REVIEW_QUEUE_URL || 'https://clawsweeper.openclaw.ai' }}
+          REVIEW_PLACEHOLDER_BACKLOG_ALERT: ${{ vars.REVIEW_PLACEHOLDER_BACKLOG_ALERT || '480' }}
+          REVIEW_PLACEHOLDER_LOOKBACK_HOURS: ${{ vars.REVIEW_PLACEHOLDER_LOOKBACK_HOURS || '48' }}
           REVIEW_PLACEHOLDER_MAX_CHECKS: ${{ vars.REVIEW_PLACEHOLDER_MAX_CHECKS || '20' }}
           REVIEW_PLACEHOLDER_MAX_RECOVERIES: ${{ vars.REVIEW_PLACEHOLDER_MAX_RECOVERIES || '5' }}
           REVIEW_PLACEHOLDER_MIN_AGE_HOURS: ${{ vars.REVIEW_PLACEHOLDER_MIN_AGE_HOURS || '2' }}
--- a/src/review-placeholder-recovery.ts
+++ b/src/review-placeholder-recovery.ts
@@ -9,12 +9,13 @@ export const DEFAULT_REVIEW_PLACEHOLDER_MIN_AGE_HOURS = 2;
 export const DEFAULT_REVIEW_PLACEHOLDER_MAX_RECOVERIES = 5;
 export const DEFAULT_REVIEW_PLACEHOLDER_STUCK_HOURS = 12;
 export const REVIEW_PLACEHOLDER_STUCK_LABEL = "clawsweeper-recovery-stuck";
-export const REVIEW_PLACEHOLDER_LOOKBACK_HOURS = 48;
+export const DEFAULT_REVIEW_PLACEHOLDER_LOOKBACK_HOURS = 48;
+export const DEFAULT_REVIEW_PLACEHOLDER_BACKLOG_ALERT = 480;
 
 const SEARCH_PAGE_SIZE = 100;
 const SEARCH_MAX_PAGES = 2;
 const COMMENT_PAGE_SIZE = 100;
-const COMMENT_MAX_PAGES = 2;
+const COMMENT_MAX_PAGES = 5;
 const CLAWSWEEPER_BOT_LOGINS = new Set(["clawsweeper[bot]", "openclaw-clawsweeper[bot]"]);
 
 export type ReviewPlaceholderComment = {
@@ -29,6 +30,7 @@ export type ReviewPlaceholderCandidate = {
   number?: unknown;
   pull_request?: unknown;
   labels?: unknown;
+  updated_at?: unknown;
 };
 
 export type ReviewPlaceholderRecoverySummary = {
@@ -38,6 +40,8 @@ export type ReviewPlaceholderRecoverySummary = {
   cleaned: number;
   escalated: number;
   errors: number;
+  matched: number;
+  remaining: number;
 };
 
 // The scheduled sweep should go red only when placeholders needed resolution
@@ -46,10 +50,14 @@ export type ReviewPlaceholderRecoverySummary = {
 // Escalation labels are visibility, not resolution, so they never count.
 export function reviewPlaceholderRecoveryFailureReason(
   summary: ReviewPlaceholderRecoverySummary,
+  backlogAlert: number = DEFAULT_REVIEW_PLACEHOLDER_BACKLOG_ALERT,
 ): string | null {
   const resolved = summary.enqueued + summary.cleaned;
   if (summary.orphaned > 0 && summary.errors > 0 && resolved === 0) {
     return "orphaned placeholders remain and every recovery action failed";
+  }
+  if (backlogAlert > 0 && summary.remaining >= backlogAlert) {
+    return `${summary.remaining} matching placeholders were left unexamined (backlog alert threshold ${backlogAlert})`;
   }
   return null;
 }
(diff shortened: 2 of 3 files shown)


yetval:
@codex review

PR #842, opened 2026-07-25
fix(exact-review): supersede stale PR heads before dispatch

brokemac79:
## Summary

- Supersede an obsolete queued pull-request head before it can reserve a workflow lease or dispatch.
- Preserve explicit command/status work by rebinding it to the authoritative current head instead of dropping it.
- Record an atomic, durable supersession audit with a rollback-compatible schema migration.

## Incident timeline and source attribution

- [#840](https://github.com/openclaw/clawsweeper/pull/840) merged at `2026-07-25T00:22:42Z` (`8006cf0bcf7ff810761c11a3cf0609b33b03475b`).
- For [openclaw/openclaw#113453](https://github.com/openclaw/openclaw/pull/113453), the old head `26ec173dd5e9b53a36ce1e3f72795a3f20715be8` was superseded by `97eac4a7486d5f0c94ae91382ebb81f9e2bf9e2f`; the durable current-head verdict already existed.
- A later queue revision nevertheless admitted the obsolete `26ec` head, dispatched it twice, and parked it as `review_retry_exhausted` after eight attempts. The lease correctly rejected the stale source tuple, but only after workflow consumption.

This repair is based on that observed stale-head incident. Validation uses only local durable storage and mocked GitHub responses; it made no production queue, target, comment, workflow, Actions, or gate mutation.

## Root cause

Admission checked only the issue's open/terminal state. An open PR whose queued `sourceHeadSha` was older than GitHub's live PR head remained dispatchable, so the lease-level source validation happened too late.

## Causal boundary and related post–21 July work

- [#749](https://github.com/openclaw/clawsweeper/pull/749) introduced the source-authority and active stale-run protections this path relies on. Its queued-admission coverage did not supersede a surviving legacy event whose PR remained open.
- [#840](https://github.com/openclaw/clawsweeper/pull/840) added terminal admission for closed/missing targets, but deliberately had no open-PR head comparison; this is the direct missing case.
- [#838](https://github.com/openclaw/clawsweeper/pull/838) bounded retry attempts, which made the missed supersession visible as `review_retry_exhausted`; it did not create the stale source condition.

These are scoped causal links, not a blanket attribution: this PR closes the interaction gap between their protections.

## Behavior change and command boundary

Before workflow dispatch, a source-head-bound pull request is read through the authoritative PR endpoint:

- An ordinary queued revision whose head differs from the live open PR head is removed locally as `live_head_advanced`; it records a supersession audit and increments completed/superseded metrics without dispatch or attempt consumption.
[...]

PR #839, opened 2026-07-24
fix(exact-review): classify dispatch failures and bound retries

Takhoffman:
## What changed

- classify exact-review `repository_dispatch` failures by HTTP and transport outcome
- park only permanent 400/404/422 item rejections as `dispatch_rejected`
- globally back off authentication, rate-limit, outage, timeout, and network failures without charging item attempts
- expose sanitized failure details through item status and queue telemetry
- preserve claimed leases after ambiguous dispatch outcomes
- allow a fresh authority event or explicit maintainer command to clear a parked rejection

## Root cause

The dispatcher collapsed every failed GitHub API call into the same per-item retry path. During a GitHub-side or credential failure, every admitted item accumulated attempts independently and retried forever with only a capped five-minute delay.

## Validation

- focused dashboard tests for permanent rejection and command recovery
- focused dashboard tests for rate limiting and GitHub outage backoff
- focused timeout-after-claim lease preservation test
- TypeScript dashboard build and lint
- full `pnpm check` static/build/lint phases passed; the broad high-concurrency test phase hit existing workflow-ledger fixture contention locally

PR #843, opened 2026-07-25
fix(exact-review): retain actionable dispatch 422 detail

brokemac79:
## Summary

Preserve actionable, redacted GitHub validation detail for exact-review dispatch failures, and avoid silently parking a normal `pull_request:synchronize` update on an opaque HTTP 422.

## Problem

On 2026-07-25 at 03:17:12Z, PR #842 received a user-authorized empty synchronize commit (`21f465f05a7d382256de1b6c7a16abaf4d042cae` -> `81e4abc894e7d3ec1fddbd856378b6aadb4392f3`). At 03:17:15.782Z the exact-review queue parked revision 3 with:

```text
state=parked
parked_reason=dispatch_rejected
status=422
class=permanent_rejection
fingerprint=dispatch-72215b4b
```

No exact-review worker run was created. The contemporaneous GitHub-activity run (`30141979462`) was intentionally skipped by its synchronize filter; an unrelated exact repository-dispatch review succeeded at 03:21:17Z, so this was not a general Actions outage.

#839 (`cf01fe706b`) classified every 400/404/422 as an item-scoped permanent rejection and retained only status/class/fingerprint. Its request helper consumed the GitHub response body but did not preserve a safe, actionable diagnosis.

## Implementation

- Move `source_head_sha` under the existing `queue_claim` object. This keeps the exact dispatch client payload at 10 top-level keys even when `review_options` is present, while the workflow continues to read legacy top-level payloads.
- Parse only allowlisted 422 validation field/code tokens. Persist and expose those tokens, never the raw GitHub message, values, paths, headers, or request body.
- Park only structured `client_payload` validation as a genuine item-specific permanent rejection.
- Treat opaque or differently scoped 422s as `validation_unknown`: retain the item, block the dispatcher with bounded global backoff, and expose the redacted diagnostic.
- Preserve secondary-rate-limit 422 handling as global rate-limit backoff. Keep per-target admission 422s item-scoped.

## Validation

```text
pnpm run build:all
node --test --test-name-pattern='(bounds item-specific terminal-state check failures|parks an item after repeated item-specific target-state failures|structured client-payload validation|retains a synchronize update|secondary-limit 422|globally backs off GitHub outages)' test/dashboard-worker.test.ts
node --test --test-name-pattern='exact event item' test/sweep-workflow.test.ts
git diff --check
```

All commands above passed. `actionlint` is not installed on this Windows host. The focused workflow parser test passed. Broader workflow tests have existing native-Windows Bash availability failures unrelated to this patch.

Regression coverage proves:

- a #842-shaped command-plus-synchronize payload has exactly 10 `client_payload` keys, retains source-head semantics, and exposes an actionable global 422 outcome instead of parking silently;
- structured malformed `client_payload` validation parks with redacted field/code detail;
- secondary-limit 422 remains global rate-limit backoff;
- non-422 GitHub outage handling remains retryable; and
- no raw validation message, URL, value, or token-shaped test data reaches durable state.

## Codex review closeout

- Dirty review: `codex review --uncommitted` initially found one accepted issue: shared 422 classification changed per-target admission failures to global scope. The patch now forces admission 422s back to item scope; focused admission tests were rerun.
- Final dirty review: `codex review --uncommitted` reported no actionable defects.
- Branch review: `codex review --base origin/main` reported no actionable correctness issues.

## Risks and rollout

The change affects exact-review dispatch and `sweep.yml` payload compatibility. New producers emit `queue_claim.source_head_sha`; the workflow reads both nested and legacy top-level locations. Opaque 422s deliberately pause the shared dispatcher with capped backoff to prevent a retry storm, leaving the affected update visible for operators rather than discarding it.

The response body from the #842 incident is not recoverable from authorized read-only run logs, artifacts, or the public queue API; this patch records only a bounded allowlist needed for future diagnosis.

## Links

- Incident target: #842
- Causal predecessor: #839
- E2E-001 is planning-only; no sandbox or production event, state, comment, label, workflow, Worker, gate, or queue mutation was performed for this PR.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -17,7 +17,7 @@ run-name: >-
       github.event.action == 'clawsweeper_target_sweep') &&
       format('Review target repo {0}', github.event.client_payload.target_repo || 'openclaw/openclaw') ||
     (github.event_name == 'repository_dispatch' && github.event.client_payload.queue_lease_id != '') &&
-      format('Review exact item {0} rev {1} head {2}', github.event.client_payload.queue_claim.item_key || github.event.client_payload.item_key || '?', github.event.client_payload.queue_claim.lease_revision || github.event.client_payload.lease_revision || '?', github.event.client_payload.source_head_sha || 'na') ||
+      format('Review exact item {0} rev {1} head {2}', github.event.client_payload.queue_claim.item_key || github.event.client_payload.item_key || '?', github.event.client_payload.queue_claim.lease_revision || github.event.client_payload.lease_revision || '?', github.event.client_payload.queue_claim.source_head_sha || github.event.client_payload.source_head_sha || 'na') ||
     (github.event_name == 'repository_dispatch' && github.event.client_payload.dispatch_key != '') &&
       format('Review event item {0}#{1} [{2}]', github.event.client_payload.target_repo || 'openclaw/openclaw', github.event.client_payload.item_number || '?', github.event.client_payload.dispatch_key) ||
     github.event_name == 'repository_dispatch' &&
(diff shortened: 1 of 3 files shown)


brokemac79:
@clawsweeper review

PR #849, opened 2026-07-25
Fix stale exact-review publication retry loop

brokemac79:
## Summary

Treat an exact-review publication queue response marked `superseded: true` as a successful handoff rather than a failed workflow step.

## Problem

During the CSW-059 monitor, [openclaw/openclaw#111368](https://github.com/openclaw/openclaw/pull/111368) repeatedly completed review generation but its publication enqueue returned `ok: true`, `deduped: true`, and `superseded: true` (publication revision 9 superseded by revision 12). The workflow then exited non-zero, so the review queue retried the same stale handoff instead of allowing the queue's newer durable publication to own delivery.

## Implementation

- Accept the queue's documented superseded acknowledgement as terminal success and emit a notice with both revisions.
- Preserve retries for transport failures, malformed responses, and all non-superseded unsuccessful enqueue responses.
- Update the workflow contract test accordingly.

## Validation

- `pnpm run build:all`
- `pnpm exec node --test --test-name-pattern "exact event review hands immutable artifacts to the queue-bounded publisher lane|exact review publication enqueue accepts a superseded acknowledgement" test/sweep-workflow.test.ts`
- `pnpm exec oxfmt --check .github/workflows/sweep.yml test/sweep-workflow.test.ts`
- `git diff --check`

The broader workflow test file has pre-existing Windows-host failures because Bash-dependent tests invoke the `docker-desktop` WSL distribution, which has no `/bin/bash`; the focused affected tests pass.

## Review closeout

- Dirty patch: `codex review --uncommitted` found two workflow/test issues; both were fixed, then focused proof was rerun and the dirty review completed cleanly.
- Committed branch: `codex review --base origin/main` completed with no actionable findings.

## Risk and rollout

This only changes handling of an explicit stale-publication acknowledgement. It does not publish stale output or suppress actual enqueue failures. Delivery still depends on the newer durable publication referenced by the queue response.

Related: CSW-059 monitoring; distinct from the #846 durable-tuple repair and CSW-058 stale-head/422 monitoring.

PR #851, opened 2026-07-25
Fix stale existing exact-review publication revisions

brokemac79:
## Summary

Rebase an existing exact-review item above its target's durable publication head whenever a newer source event updates that item.

## Problem

An existing review item can retain a local revision below the permanent publication head for the same target. A normal re-review then incremented only the stale local value (for example, 7 to 8 to 9 while the durable head was 12), so its artifact was rejected as superseded before publisher, state writer, router, or final comment delivery.

## Implementation

- Compute the next existing-item revision as `max(current + 1, publication head + 1)`.
- Keep the supersession audit's recorded next revision aligned with the rebased value.
- Add a lifecycle regression test for persisted revision 7, publication head 12, and a newer source event producing revision 13 that queues normally.

## Validation

- `pnpm run build:dashboard`
- `pnpm exec node --test --test-name-pattern "a later exact-review cycle still publishes after an earlier cycle raised the publication head|an existing exact-review item rebases above the publication head on a newer source event" test/dashboard-worker.test.ts`
- `pnpm exec oxfmt --check dashboard/exact-review-queue.ts test/dashboard-worker.test.ts`
- `git diff --check`

## Review closeout

- `codex review --uncommitted`: clean after focused proof.
- `codex review --base origin/main`: clean; reviewer also ran all 209 `test/dashboard-worker.test.ts` tests successfully.

## Risk and rollout

This affects only an incoming newer source event updating an existing review item. It does not replay, mutate, or change already completed publications. It closes the pre-existing-item gap left after #836 seeded only new items above the durable head.

Related: #836, #849, and CSW-059 / openclaw/openclaw#111368.

### DIFF
--- a/dashboard/exact-review-queue.ts
+++ b/dashboard/exact-review-queue.ts
@@ -999,6 +999,10 @@ export class ExactReviewQueue {
           // Ordinary source events retain normal replacement behavior, including the
           // command-context merge for pending items.
           if (!ignoredRecovery) {
+            const nextRevision = Math.max(
+              current.revision + 1,
+              this.nextExactReviewItemRevisionSync(key),
+            );
             // Explicit commands arrive through repository_dispatch without a webhook authority
             // tuple. Bind them to the current verified decision via the merge below. An active
             // review keeps its lease and exposes the command as a follow-up revision on completion.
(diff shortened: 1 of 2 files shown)


brokemac79:
@clawsweeper review

PR #852, opened 2026-07-25
Draft: revert #851 existing-review revision rebase

brokemac79:
## Draft reversal for #851

This is the exact inverse of [#851](https://github.com/openclaw/clawsweeper/pull/851), prepared as a draft-only fallback.

It deliberately targets #851's branch rather than `main`, so it cannot become a production rollback until #851 has merged and a maintainer explicitly retargets and reviews it.

## Do not merge

This reversal restores the existing-item revision behavior that can leave a re-review below the durable publication head. Keep it as rollback material only; it needs a separate production diagnosis and fresh review before any merge decision.

Related: #851, #849, #836, CSW-059 / openclaw/openclaw#111368.


steipete:
Thanks @brokemac79 for preparing this contingency. This PR now has zero diff against `main`, so there is no rollback change to review or land. We are closing the no-op draft; please open a fresh revert PR against the then-current `main` only if an actual production incident demonstrates that a rollback is needed.

PR #941, opened 2026-07-29
Add MiniMax built-in provider defaults for the OpenClaw runner

octo-patch:
Reason: There is no MiniMax provider configuration, so `minimax/...` models cannot run through the OpenClaw runner without hand-written provider config.

## Changes
- Add a `minimax` entry to the built-in OpenClaw provider defaults in `src/openclaw-process.ts`, using MiniMax's global OpenAI-compatible endpoint (`https://api.minimax.io/v1`, `${MINIMAX_API_KEY}`, `openai-completions`) with the `MiniMax-M3` (1M context) and `MiniMax-M2.7` (200K context) text models. `MINIMAX_API_KEY` was already on the subprocess environment allowlist, so no key wiring changes were needed.
- Add a unit test asserting the generated provider config for a `minimax/...` model, mirroring the existing kimi/cerebras/zai built-in provider coverage.
- Update `docs/repair/operations.md` to list MiniMax among the built-in providers.

## Checks
- pnpm run build
- pnpm run lint:src
- pnpm run lint:scripts
- pnpm run format:check

Note: the repository's `node --test` TypeScript suite targets Node >=24 and could not be executed in this Node 20 environment; the change was validated by the type build, lint, and format checks above.

### DIFF
--- a/docs/repair/operations.md
+++ b/docs/repair/operations.md
@@ -250,12 +250,14 @@ otherwise ClawSweeper runs `openclaw` from `PATH`.
 Providers that are not bundled with OpenClaw can be supplied through
 `CLAWSWEEPER_OPENCLAW_PROVIDERS_JSON`, a JSON object merged into
 `models.providers`; referenced provider API-key environment variables must also
-be present. Three providers ship as built-in defaults and need only their API key
+be present. Four providers ship as built-in defaults and need only their API key
 in the environment: `kimi/…` (Kimi Code endpoint, `KIMI_API_KEY`; models
 `kimi-for-coding` and `k3`), `cerebras/…` (Cerebras Code plans,
 `CEREBRAS_API_KEY`; model `zai-glm-4.7` until Cerebras retires it on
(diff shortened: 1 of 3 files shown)


brokemac79:
@octo-patch 

Thanks for the contribution — the intent to make MiniMax usable in repair runs is reasonable.

I’m going to close this PR in its current form because it adds a second, ClawSweeper-owned MiniMax provider contract. OpenClaw already owns MiniMax support, including its API/portal paths and configured endpoint overrides; injecting a models.providers.minimax block here can shadow that supported path and drift from it over time.
This is not a rejection of MiniMax itself. MiniMax M3 is supported upstream; the concern is provider ownership and compatibility in the repair runner. The current PR also has static/fake-process coverage only, rather than proof from a real authenticated repair run.

If we later decide ClawSweeper must support an OpenClaw version that predates the bundled MiniMax provider, we can reopen this direction with an explicit maintainer decision, compatibility coverage for that version boundary, and redacted real-run proof. Otherwise, the preferred path is to rely on OpenClaw’s bundled MiniMax provider.

Thank you again for the focused patch and documentation work.

PR #970, opened 2026-07-30
fix(review): stop re-reviewing unchanged heads after duplicate checks

### LINKED ISSUE #967: ClawSweeper's own review writes expire the review content cache, so unchanged heads are re-reviewed in a loop
ClawSweeper's own review writes expire the review content cache, so unchanged heads are re-reviewed in a loop

## What happens

`reviewContentCacheHit` is supposed to skip a full re-review when an item's content
has not changed. For pull requests it effectively never gets the chance: every
completed review invalidates its own cache entry, so the same unchanged head is
reviewed again and again.

The durable comment records prior cycles as
`- reviewed <ts> sha <sha> :: <verdict> :: <asks>`, so the rate is public:

```bash
repo=openclaw/clawsweeper
for n in $(gh pr list --repo "$repo" --state all --limit 400 --json number --jq '.[].number'); do
  gh api "repos/$repo/issues/$n/comments" --paginate \
    --jq '.[] | select(.user.login=="clawsweeper[bot]") | .body' 2>/dev/null |
    grep '^- reviewed ' | sed "s|^|$n |"
done | python3 -c '
import sys, collections, datetime as dt, statistics
by=collections.defaultdict(list)
for line in sys.stdin:
    head, _, rest = line.rstrip("\n").partition(" :: ")
    verdict = rest.split(" :: ")[0]
    f = head.split()
    if len(f) < 6: continue
    by[f[0]].append((f[3], f[5], verdict))
gaps=[]; pairs=0; flips=0
for pr, rows in by.items():
    rows.sort()
    for (t1,s1,v1),(t2,s2,v2) in zip(rows, rows[1:]):
        if s1 != s2: continue            # head changed -> a re-review is expected
        pairs += 1
        if v1 != v2: flips += 1
        a=dt.datetime.fromisoformat(t1.replace("Z","+00:00"))
        b=dt.datetime.fromisoformat(t2.replace("Z","+00:00"))
        gaps.append((b-a).total_seconds()/60)
print(f"consecutive re-reviews of an unchanged head: {pairs}")
print(f"median gap: {statistics.median(gaps):.0f} min")
print(f"under 90 min: {sum(1 for g in gaps if g<90)}")
print(f"recorded verdict changed: {flips} ({100*flips/pairs:.0f}%)")
'
```

| | `openclaw/clawsweeper` | `openclaw/openclaw` |
|---|---|---|
| consecutive re-reviews of an **unchanged head** | 109 | 145 |
| median gap between them | **42 min** | **10 min** |
| under 90 min | 79 | 134 |
| pairs where the recorded verdict **changed** | 16 (15%) | 22 (15%) |

`reviewCadenceMs` never returns less than `HOURLY_REVIEW_MS`, so most of those are
below the intended floor. The counts come only from the history block, which
records *prior* cycles, so they understate the total; they also drift upward as new
reviews land. The verdict changes go both ways, and #835 oscillates on one head:
`needs maintainer review` → `found issues` → `needs changes` → `found issues`.

## Why

`itemContentDigest` includes the PR's check runs verbatim:

```ts
checks: isPull ? (context.pullChecks ?? null) : null,
```

`pullChecksContext` fills that from `commits/{headSha}/check-runs?per_page=100`
with no author or app filter, so it includes ClawSweeper's own runs. That closes a
loop:

1. a review completes and writes its labels and durable comment
2. `clawsweeper-dispatch.yml` (job `dispatch`) and `github-activity.yml` (job
   `notify`) both trigger on `labeled` / `unlabeled` / `issue_comment`
3. they add check runs to the PR head
4. `pullChecks` changes, so `itemContentDigest` changes
5. `reviewContentCacheHit` misses
6. the unchanged head is re-reviewed — back to step 1

On #948's head `144236be`, 26 of 38 check runs are those two jobs: 13 `notify` and
13 `dispatch`. After `compactCheckRun` reduces each run to
`{name,status,conclusion,app}`, 13 of them are byte-identical to each other, so
they carry no check state the first one did not.

The same bot is already excluded from every sibling digest input — only checks
were missed:

| digest input | ClawSweeper's own writes | test |
|---|---|---|
| `timeline` | excluded via `CLAWSWEEPER_BOT_AUTHORS` | `content digest is stable across bot-only context churn` |
| labels | excluded via `isIgnorableSourceRevisionLabel` | `content digest ignores advisory-label timeline churn` |
| PR review comments | excluded via `isClawSweeperComment` | `content digest ignores ClawSweeper's own PR review comments` |
| **checks** | **not excluded** | only `busts when bounded PR check state changes` |

## Reproduction

Driving the exported `itemContentDigestForTest` and `reviewContentCacheHit`
against the check-run list actually observed on that head:

```
=== A. Does ClawSweeper's own reaction churn change the digest? ===

  OK    one more bot reaction run changes the digest               1092d87e96d999aa -> fd84f307765acbe6
  OK    the added entry is a duplicate of an existing one          13 identical {notify,completed,success} entries already present

=== B. The same bot is filtered out of every sibling digest input ===

  OK    bot timeline entry does NOT change the digest              filtered by CLAWSWEEPER_BOT_AUTHORS
  OK    bot advisory label churn does NOT change the digest        filtered by isIgnorableSourceRevisionLabel
  OK    bot PR review comment does NOT change the digest           filtered by isClawSweeperComment
  OK    bot check run DOES change the digest                       pullChecks has no bot filter  <-- the asymmetry

=== E. Effect on the production cache predicate ===

  OK    cache MISSES after only bot reaction churn -> full re-review reviewContentCacheHit = false
```

The harness is in the linked PR's description; it needs no credentials, only
`pnpm run build`.

## Cost

Two things, and I am more confident about the first than the second.

Every wasted cycle is a full Codex review: a leased runner, a checkout, and a
model call, for an item whose content did not change. At the observed rate that is
the majority of same-head cycles.

The verdict instability is the part contributors see. A PR can be re-rated with
nothing changed on its side. I have measured the correlation (15% of same-head
re-reviews change the recorded verdict) but **not** established how much of it this
loop causes rather than ordinary model variance — the loop explains why the extra
cycles happen, not why two runs over identical input disagree.

Note the same raw `pullChecks` also feeds the semantic and structural digests
(`src/review-semantic-cache.ts:785`, `src/clawsweeper.ts:9343`). Those are
consulted as separate, earlier skips, so duplicate runs churn them too. Fixing
only the content digest still prevents the expensive outcome — a full Codex review
of unchanged content — but does not remove the hydration and revalidation work.

## Possible fix

De-duplicating the compacted check runs inside `itemContentDigest` is enough:
once a run is reduced to `{name,status,conclusion,app}`, a repeat of that tuple
adds nothing, while a run that newly fails differs in `conclusion` and keeps its
own entry. That leaves `pullChecksContext` alone, so what the reviewer is shown
through `reviewContextLedger` does not change — only the cache key does.

I have that up as a PR with the harness and tests, but I would rather you pick the
shape. Two alternatives I considered and did not take:

- **de-duplicate in `pullChecksContext`** — also shrinks the noise in the review
  prompt, but changes reviewer input, not just the cache key
- **filter by bot identity** — `app.slug` is `github-actions` for both the
  `dispatch`/`notify` runs and genuine CI, so this would need a workflow-name
  allowlist that drifts as workflows are renamed

This is the same shape as `pullBaseDriftDigestParts` in #589 ("Excluding raw age
avoids daily cache churn"), which normalizes a different digest input for the same
reason. That PR and this one both add a helper next to `reviewTimelineDigestParts`,
so whichever lands second will need a trivial merge.

## Not part of this report

- **Truncation.** `checkRunsTruncated` trips above 100 runs and would make
  `completePullChecksContext` false. These runs accumulate toward that, but nothing
  has crossed it: across the 60 most recent PRs on this repo the largest head
  carried 37 check runs (38 counting #948). Flagging it as a direction, not a live
  failure.
- **Sub-15-minute re-reviews.** 30 same-head pairs on `openclaw/clawsweeper` are
  under 15 minutes and some are under 5, which is faster than this loop alone
  explains. That looks like a separate lane or lease issue and I have not
  separated it out.
- Whether two reviews over identical input *should* agree is a different question
  from how often they are asked to.

PollyBot13:
Closes #967

## What Problem This Solves

Fixes an issue where contributors with an unchanged pull-request head could receive repeated full ClawSweeper reviews after equivalent GitHub Actions check runs accumulated on that head. Those unnecessary cycles consume review capacity and can expose contributors to different model verdicts despite no change to their code or discussion.

## Why This Change Was Made

Cache-key construction now de-duplicates equivalent compacted check-run tuples across the structural, semantic, and content review caches. The full check context supplied to the reviewer is deliberately unchanged; distinct names, apps, statuses, or conclusions still invalidate the relevant caches, and existing completeness/truncation safeguards remain intact.

This keeps the repair at the cache identity boundary requested in #967 rather than introducing a workflow-name allowlist or hiding repeated runs from review context.

## User Impact

An unchanged PR head no longer receives another full review solely because automation appended another equivalent check run. Meaningful CI changes—such as a previously successful check failing—continue to trigger fresh evaluation.

## Evidence

- Added a regression proving one additional `{name,status,conclusion,app}` duplicate leaves the content digest unchanged.
- Added an end-to-end cache-predicate regression proving the unchanged head is reused instead of sent through another full review.
- Added semantic-cache coverage and updated the structural-cache source assertion so all three cache layers use the same normalization boundary.
- Added contributor-suggested identity regressions proving runs that differ only by name or app remain distinct, and that one failure among several repeated runs invalidates the cache.
- Mutation proof: narrowing the identity to name-only makes exactly those two guard tests fail (38/40); restoring full-object identity returns the content-cache suite to 40/40.
- The new regression failed on unmodified `main` before the fix: the two equivalent contexts produced different content digests.
- Rebased onto current `main` at `c8f44886fde43c44187a2f239dc0623fdaa279de`.
- `pnpm run check:static`, `pnpm run build:all`, and `pnpm run lint` pass on Node 26.5.0 / pnpm 11.10.0.
- 148 focused content, semantic, structural, and review-flow tests pass.
- `pnpm run check` completes all changed-surface validation; its five remaining failures are in repair files unchanged from `origin/main` and are Linux Landlock/capability tests running on macOS (`capset` is absent from Darwin libc). Hosted Linux CI is the authoritative full-suite result.
- Codex review of both the uncommitted change and committed branch found no actionable defects.

## Real Behavior Proof

**Claim:** Adding another equivalent compacted check run to an unchanged PR head preserves the review cache identity, while a distinct failure still invalidates it.

**Exercised surface:** The production `itemContentDigestForTest` export and `reviewContentCacheHit` predicate from the built package.

**Scenario:** The fixture keeps the PR source revision, head, base, files, timeline, and check outcome fixed. It compares one successful `notify` run with two byte-equivalent successful `notify` runs, then changes the second run's conclusion to `failure`.

**Command and environment:** A Node 26.5.0 ESM harness against `pnpm run build` output on macOS 15.

**Observed result:**

```json
{
  "priorDigest": "aadae6eeaf1ce3ecfc7c0ef8eeeb72b8d1c18c6c693ff9fd38268154e260cbf5",
  "repeatedDigest": "aadae6eeaf1ce3ecfc7c0ef8eeeb72b8d1c18c6c693ff9fd38268154e260cbf5",
  "duplicateStable": true,
  "distinctFailureInvalidates": true,
  "cacheHit": true
}
```

**Trace:** The executable fixture is captured as the focused regression in `test/review-content-cache.test.ts`; semantic identity is independently covered in `test/review-semantic-cache.test.ts`.

**Limits:** This does not alter check-run collection, the reviewer-visible context, the 100-run completeness boundary, scheduler cadence, explicit re-review commands, or maintainer-request behavior.

## AI Assistance

This PR was implemented and reviewed with OpenAI Codex assistance. The author ran and inspected the reported validation and remains responsible for the change.

### DIFF
--- a/src/clawsweeper.ts
+++ b/src/clawsweeper.ts
@@ -54,6 +54,7 @@ import {
 } from "./github-retry.js";
 import { parseGhJson, parseGhJsonLinesWithRetry, parseGhJsonWithRetry } from "./github-json.js";
 import { stableJson } from "./stable-json.js";
+import { reviewPullChecksDigestParts } from "./review-checks-digest.js";
 import { coverageTrackedItemIdsFromManifest } from "./review-coverage-manifest.js";
 import {
   LEGACY_FIXED_CLOSE_SKIP_ACTIONS,
@@ -3184,7 +3185,7 @@ function itemContentDigest(item: Item, context: ItemContext, git?: GitInfo): str
         ? (context.pullReviewCommentsRevision ??
           reviewCommentDigestParts(context.pullReviewComments))
         : null,
-      checks: isPull ? (context.pullChecks ?? null) : null,
+      checks: isPull ? reviewPullChecksDigestParts(context.pullChecks ?? null) : null,
     }),
   );
 }
@@ -9165,7 +9166,7 @@ function fetchReviewStructuralRecord(options: {
     if (!headSha) return null;
     const pullChecks = pullChecksContext(options.item.number, headSha);
     if (!completePullChecksContext(pullChecks)) return null;
-    pullChecksDigest = sha256(stableJson(pullChecks));
+    pullChecksDigest = sha256(stableJson(reviewPullChecksDigestParts(pullChecks)));
   }
   return reviewStructuralRecordFromGraphql({
     response,
--- a/src/review-checks-digest.ts
+++ b/src/review-checks-digest.ts
@@ -1,0 +1,17 @@
+import { stableJsonCodeUnit } from "./stable-json.js";
+
+export function reviewPullChecksDigestParts(value: unknown): unknown {
+  if (!value || typeof value !== "object" || Array.isArray(value)) return value;
+  const checks = value as Record<string, unknown>;
+  if (!Array.isArray(checks.checkRuns)) return value;
+
+  const seen = new Set<string>();
+  const checkRuns = checks.checkRuns.filter((checkRun) => {
+    const identity = stableJsonCodeUnit(checkRun);
+    if (seen.has(identity)) return false;
+    seen.add(identity);
+    return true;
+  });
+
+  return checkRuns.length === checks.checkRuns.length ? value : { ...checks, checkRuns };
+}
--- a/src/review-semantic-cache.ts
+++ b/src/review-semantic-cache.ts
@@ -6,6 +6,7 @@ import { createVirtualFileSystem, type FileSystem } from "typescript/unstable/fs
 import { API } from "typescript/unstable/sync";
 
 import { REVIEW_CACHE_MAX_AGE_DAYS } from "./scheduler-policy.js";
+import { reviewPullChecksDigestParts } from "./review-checks-digest.js";
 import { stableJsonCodeUnit as stableJson } from "./stable-json.js";
 
 export const REVIEW_SEMANTIC_CACHE_VERSION = 10;
@@ -782,7 +783,7 @@ function semanticContext(input: ReviewSemanticInput): {
     reviewComments:
       input.context.pullReviewCommentsRevision ??
       normalizedComments(input.context.pullReviewComments),
-    checks: input.context.pullChecks ?? null,
+    checks: reviewPullChecksDigestParts(input.context.pullChecks ?? null),
     completeness: {
       commentsTruncated: input.context.counts?.commentsTruncated ?? null,
       timelineTruncated: input.context.counts?.timelineTruncated ?? null,
--- a/test/review-comment-rendering.test.ts
+++ b/test/review-comment-rendering.test.ts
@@ -166,7 +166,10 @@ test("structural cache probes before hydration but acquires a lease before carry
     source.indexOf("function collectItemContext"),
   );
   assert.match(structuralProbeSource, /pullChecksContext\(options\.item\.number, headSha\)/);
-  assert.match(structuralProbeSource, /pullChecksDigest = sha256\(stableJson\(pullChecks\)\)/);
+  assert.match(
+    structuralProbeSource,
+    /pullChecksDigest = sha256\(stableJson\(reviewPullChecksDigestParts\(pullChecks\)\)\)/,
+  );
   assert.match(structuralProbeSource, /if \(!options\.git\.releaseStateComplete\) return null/);
   const gitInfoBlock = source.slice(
     source.indexOf("function gitInfo("),
--- a/test/review-content-cache.test.ts
+++ b/test/review-content-cache.test.ts
@@ -294,6 +294,101 @@ test("content digest busts when bounded PR check state changes", () => {
   assert.notEqual(passing, failing);
 });
 
+test("content digest ignores duplicate compacted PR check runs", () => {
+  const pull = item({ kind: "pull_request", number: 200 });
+  const notify = {
+    name: "notify",
+    status: "completed",
+    conclusion: "success",
+    app: "github-actions",
+  };
+  const once = itemContentDigestForTest(
+    pull,
+    pullContext({
+      pullChecks: {
+        complete: true,
+        checkRuns: [notify],
+        checkRunsTruncated: false,
+        statuses: [],
+        statusesTruncated: false,
+      },
+    }),
+  );
+  const repeated = itemContentDigestForTest(
+    pull,
+    pullContext({
+      pullChecks: {
+        complete: true,
+        checkRuns: [notify, notify],
+        checkRunsTruncated: false,
+        statuses: [],
+        statusesTruncated: false,
+      },
+    }),
+  );
+
+  assert.equal(once, repeated);
+});
+
+test("content digest keeps check runs that differ only in name or app", () => {
+  const pull = item({ kind: "pull_request", number: 200 });
+  const check = {
+    name: "smoke",
+    status: "completed",
+    conclusion: "success",
+    app: "github-actions",
+  };
+  const digest = (checkRuns) =>
+    itemContentDigestForTest(
+      pull,
+      pullContext({
+        pullChecks: {
+          complete: true,
+          checkRuns,
+          checkRunsTruncated: false,
+          statuses: [],
+          statusesTruncated: false,
+        },
+      }),
+    );
+
+  const alone = digest([check]);
+  const byName = digest([check, { ...check, name: "smoke (windows)" }]);
+  const byApp = digest([check, { ...check, app: "blacksmith-sh" }]);
+
+  assert.notEqual(alone, byName);
+  assert.notEqual(alone, byApp);
+  assert.notEqual(byName, byApp);
+});
+
+test("content digest busts when one of several repeated check runs newly fails", () => {
+  const pull = item({ kind: "pull_request", number: 200 });
+  const notify = {
+    name: "notify",
+    status: "completed",
+    conclusion: "success",
+    app: "github-actions",
+  };
+  const digest = (checkRuns) =>
+    itemContentDigestForTest(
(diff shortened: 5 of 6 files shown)


PollyBot13:
@codex review

masatohoshino:
I filed #967 and had a narrower version of this in flight; yours is the better boundary, so I closed mine. Two tests I had that yours doesn't cover, if they're useful — both guard the *identity key* rather than the de-duplication itself.

I checked them against your branch at `db44c8a0`: they pass as-is (41 tests, 41 passing).

To confirm they earn their place, I narrowed your identity key to name-only:

```diff
-    const identity = stableJsonCodeUnit(checkRun);
+    const identity = stableJsonCodeUnit((checkRun as Record<string, unknown>)?.name);
```

```
✖ content digest keeps check runs that differ only in name or app
✖ content digest busts when one of several repeated check runs newly fails
ℹ tests 41   ℹ pass 39   ℹ fail 2
```

Your three current tests still pass under that change — a regression that widens what counts as "the same run" lands silently without these two.

```ts
test("content digest keeps check runs that differ only in name or app", () => {
  const pull = item({ kind: "pull_request", number: 200 });
  const check = { name: "smoke", status: "completed", conclusion: "success", app: "github-actions" };
  const digest = (checkRuns: unknown[]) =>
    itemContentDigestForTest(
      pull,
      pullContext({
        pullChecks: {
          complete: true,
          checkRuns,
          checkRunsTruncated: false,
          statuses: [],
          statusesTruncated: false,
        },
      }),
    );

  const alone = digest([check]);
  const byName = digest([check, { ...check, name: "smoke (windows)" }]);
  const byApp = digest([check, { ...check, app: "blacksmith-sh" }]);

  assert.notEqual(alone, byName);
  assert.notEqual(alone, byApp);
  assert.notEqual(byName, byApp);
});

test("content digest busts when one of several repeated check runs newly fails", () => {
  const pull = item({ kind: "pull_request", number: 200 });
  const notify = {
    name: "notify",
    status: "completed",
    conclusion: "success",
    app: "github-actions",
  };
  const digest = (checkRuns: unknown[]) =>
    itemContentDigestForTest(
      pull,
      pullContext({
        pullChecks: {
          complete: true,
          checkRuns,
          checkRunsTruncated: false,
          statuses: [],
          statusesTruncated: false,
        },
      }),
    );

  const allPassing = digest([notify, notify, notify]);
  const oneFailing = digest([notify, notify, { ...notify, conclusion: "failure" }]);

  assert.notEqual(allPassing, oneFailing);
});
```

The second one matters for the case that started this: the head I measured carried 13 identical `notify` runs, so "several repeats, one of them newly failing" is the realistic shape rather than a single duplicated pair.

I also wrote a third, asserting a checks payload with no `checkRuns` array is passed through. I dropped it — I could not construct a change to `reviewPullChecksDigestParts` that it catches, so it looked like coverage without teeth. Mentioning it only so you know that branch is untested if you'd rather pin it some other way.

One thing your version handles that mine didn't: extracting the helper into a shared module so all three cache identities can use it. I kept mine to the content digest specifically to avoid that relocation, which left the semantic and structural digests still churning. That was the wrong call.

Take or ignore these freely — no need to credit.

PollyBot13:
Thank you — both tests are excellent guardrails, and I added them in `0a3d5c55a2`.

I repeated your mutation proof locally: narrowing the identity to name-only makes exactly those two tests fail (38/40), while restoring the full-object identity returns the content-cache suite to 40/40. I agree the non-list pass-through case did not earn a test without a mutation it could catch.

I also rebased the branch onto current `main`; the broader focused suite is now 148/148. Thanks for the careful review and for steering this toward the shared three-cache boundary.

PR #1001, opened 2026-08-01
fix: prepare frozen Knip for network-isolated OpenClaw validation

steipete:
## Problem

OpenClaw root-cause issue repairs were reaching their real changed-validation gate and then failing because `pnpm dlx --package knip@6.8.0 knip` tried to download Knip inside ClawSweeper's deliberately network-isolated validation sandbox. For example, repair run [30695474227](https://github.com/openclaw/clawsweeper/actions/runs/30695474227) reached `ERR_PNPM_META_FETCH_FAIL`/`EAI_AGAIN` instead of creating its OpenClaw pull request. A green outer workflow alone did not indicate successful repair.

## Changes

- Extract the pinned OpenClaw validation helper into its own narrow repair module.
- Prepare only the checked-in, exactly pinned `knip@6.8.0` helper when the selected changed-validation plan actually includes an OpenClaw source file scanned by Knip; docs-only repairs, disabled changed gates, and ordinary toolchain preparation never fetch it.
- Preserve OpenClaw's **48-hour minimum release age** and any stricter repository policy; install every transitive dependency directly from the committed `config/openclaw-knip-6.8.0.pnpm-lock.yaml` with `--frozen-lockfile`, and fail closed on any graph mismatch before downloaded package code can execute.
- Populate pnpm's existing offline `dlx` cache from that frozen installation without running downloaded package code or package lifecycle scripts.
- Explicitly disable package lifecycle scripts, pnpmfiles, pre/post scripts, and the global virtual store while staging the helper in the disposable validation profile outside the protected OpenClaw checkout.
- Bind its cache into the already immutable Corepack runtime and copy it afresh into each network-isolated validation profile, preserving the real changed-gate dead-code check without granting validation network access.
- Scope Jiti's no-filesystem-cache setting to the trusted Knip launcher only, so Knip cannot write `node_modules/.cache/jiti` while OpenClaw's own plugin-loader cache tests retain their normal behavior.
- Match pnpm's registry-host encoding, including approved registries using nondefault HTTPS ports.
- Configure pnpm 11 with its supported `PNPM_CONFIG_OFFLINE=true` and `PNPM_CONFIG_REGISTRY` variables, while preserving pnpm 10's legacy `npm_config_offline=true`, so either supported pnpm generation remains network-isolated and cannot silently fall back to the public registry or a fresh package download.
- Expand NUL-delimited, ignored-file-aware untracked paths before source classification so a newly created source directory receives the same frozen helper as a modified tracked source file.
- Reject absolute paths, `.` segments, and `..` traversal segments in otherwise trusted mutating formatter hints.
- Preserve standalone local review and leave OpenClaw issue-PR creation on `gpt-5.6-sol` / `xhigh` with `CLAWSWEEPER_ALLOW_MERGE=0`; generated OpenClaw PRs remain create-only and never enable auto-merge.

## Security-boundary disposition

The frozen dependency graph is completely committed and integrity-bound; package staging uses `--frozen-lockfile` and rejects any mismatch before helper execution. Lifecycle scripts, pnpmfiles, pre/post scripts, and the global virtual store are disabled. Setup can write only to a disposable private profile outside the protected OpenClaw checkout. The digest-bound cache is frozen, then copied privately for each isolated validation command. Only the fixed Knip launcher disables Jiti's filesystem cache. Production pnpm is explicitly forced offline and bound to the same approved registry used during frozen setup, including custom registries and nondefault ports. The real pinned helper, the protected target checkout, and the existing plugin-loader suite were verified end to end. OpenClaw fix PRs remain create-only; no target PR receives merge or auto-merge authority.

## Real behavior proof

Verified head: `2261169240cbb4d570a956c521767d7735c786ca`.

The actual pinned package was installed with `--frozen-lockfile`, full lifecycle execution disabled, and release-age enforcement enabled; then the real OpenClaw-style `pnpm check:changed` gate executed real `knip@6.8.0` with networking unavailable:

```json
{
  "actualPinnedKnipExecuted": true,
  "offlineValidationSucceeded": true,
  "commands": ["pnpm check:changed"],
  "targetJitiCacheCreated": false
}
```

The same actual end-to-end proof also succeeds when the target repository's policy is increased to seven days:

```json
{
  "actualPinnedKnipExecuted": true,
  "offlineValidationSucceeded": true,
  "minimumReleaseAgeMinutes": 10080,
  "frozenDependencyGraph": true
}
```

Separately, copying only the staged pnpm cache into a fresh profile and pointing HTTP/HTTPS/all proxies at an unreachable listener still ran the actual pinned package successfully and printed `6.8.0`.

Independent execution also confirmed that real pnpm 11 runs the frozen pinned helper with `PNPM_CONFIG_OFFLINE=true`, while pnpm 10 reports offline mode only when the preserved legacy `npm_config_offline=true` accompanies it. Focused integration coverage requires both settings and the same approved registry for every isolated validation command.

Focused coverage asserts that helper setup never writes to the OpenClaw checkout, installs only its complete frozen `knip@6.8.0` dependency graph, preserves the 48-hour release-age policy, rejects a tampered transitive dependency graph, works with registries on nondefault HTTPS ports, never downloads Knip for docs-only or disabled-gate validation, disables every package execution hook and the global virtual store, restores a clean private cache for each validation command, prevents Knip-only Jiti checkout writes without breaking existing plugin-loader tests, and rejects absolute/traversal formatter paths.

## Verification

- `pnpm run build:all`
- `pnpm run lint:repair`
- `pnpm run lint:scripts`
- `pnpm run format:check`
- `pnpm run check:active-surface`
- `pnpm run check:limits`
- Focused target-validation and process-tree-containment integration suites.
- The existing OpenClaw `src/plugins/plugin-module-loader-cache.test.ts` suite: **19/19 passed** with Knip's cache override scoped to Knip alone.
- Full affected target-validation/source/workflow regression suite: **228 passed**, **3 expected skips**.
- Actual real-Knip, offline, protected-checkout proof described above.
- Independent Codex `/review` on the exact current head.

### DIFF
--- a/config/openclaw-knip-6.8.0.pnpm-lock.yaml
+++ b/config/openclaw-knip-6.8.0.pnpm-lock.yaml
@@ -1,0 +1,838 @@
+lockfileVersion: "9.0"
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+  .:
+    dependencies:
+      knip:
+        specifier: 6.8.0
+        version: 6.8.0
+
+packages:
+  "@emnapi/core@1.10.0":
+    resolution:
+      {
+        integrity: sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==,
+      }
+
+  "@emnapi/core@1.11.2":
+    resolution:
+      {
+        integrity: sha512-TC8MkTuZUtcTSiFeuC0ksCh9QIJ5+F21MvZ4Wn4ORfYaFJ/0dsiudv5tVkejgwZlwQ39jL9WWDe2lz8x0WglOA==,
+      }
+
+  "@emnapi/runtime@1.10.0":
+    resolution:
+      {
+        integrity: sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==,
+      }
+
+  "@emnapi/runtime@1.11.2":
+    resolution:
+      {
+        integrity: sha512-kyOl3X0DuTiT1h2ft8r2fYO8JYtU9a9Xis/zBSiGArNaagCOWx90N1k2wxp18czFDH+OgcWGb5ZP/XMt3dcyPA==,
+      }
+
+  "@emnapi/wasi-threads@1.2.1":
+    resolution:
+      {
+        integrity: sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==,
+      }
+
+  "@emnapi/wasi-threads@1.2.2":
+    resolution:
+      {
+        integrity: sha512-c95qOXkHdydNKhscBTebqEC1CVAZpyqOfVfBzQ1qgzyl3gfeldUjIggDbIZgDKsHLgnsM+igH7TJ/eAasaVuMA==,
+      }
+
+  "@napi-rs/wasm-runtime@1.1.6":
+    resolution:
+      {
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper re-review

PR #1044, opened 2026-08-06
fix: external links with an uppercase URL scheme are published in review evidence

### LINKED ISSUE #1043: Uppercase URL scheme bypasses external-link redaction in published review evidence
## Summary

ClawSweeper strips non-`github.com` links out of worker evidence before publishing
it, and its result validator rejects any that slip through. Both checks match the
URL scheme **case-sensitively**, so a link written as `HTTPS://host/path` is
matched by neither. It is published verbatim into the durable review comment as a
live external autolink, and the worker result still validates as `passed`.

## Impact

`sanitizeEvidenceText` / `sanitizeResultEvidence` (`src/repair/url-safety.ts`) is a
content-safety control on **model-authored** text — worker evidence is written by
Codex and then published to GitHub. Its purpose is to guarantee that no
non-GitHub link reaches a published comment.

The control has an exact-case blind spot:

- `https://vercel.com/x` → redacted to `<external link>` ✅
- `HTTPS://vercel.com/x` → published verbatim ❌

The backstop does not catch it either. `evidenceHasExternalUrl` in
`src/repair/review-results.ts` uses a byte-identical case-sensitive pattern, so
both layers agree and are both wrong. `src/repair/url-safety.ts` carries a header
comment requiring the two patterns to stay identical, which is exactly why the
blind spot is symmetric rather than caught by the second layer.

This matters because **GitHub's autolinker is case-insensitive**. `HTTPS://…`
renders as a live, clickable link in the published comment, so the redaction the
control is supposed to guarantee simply does not happen for that input.

Severity is bounded — the sanitizer restricts *link surface* in bot-published
text; it is not an authentication or authorization boundary. But it is a control
that silently does not hold for a trivially reachable input.

## Reproduction

Against a build at `0588bda9`:

```js
import { sanitizeEvidenceText } from "./dist/repair/url-safety.js";

sanitizeEvidenceText("deploy preview: https://vercel.com/openclaw/abc");
// → "deploy preview: <external link>"          ✅

sanitizeEvidenceText("deploy preview: HTTPS://vercel.com/openclaw/abc");
// → "deploy preview: HTTPS://vercel.com/openclaw/abc"   ❌
```

Confirmed for `HTTP`, `Https`, `hTTpS`, `HtTp`. The bypass survives
`sanitizeResultEvidence` — the pre-publication entry point — across
`actions[].evidence`, `needs_human`, and every `merge_preflight` evidence field.

End-to-end against the real validator CLI, using an otherwise fully valid fixture
so the external-URL failure is the only possible failure:

```
lowercase scheme  → exit 1, failures: ["#1 evidence contains non-GitHub external URL"]
uppercase scheme  → exit 0, status "passed", failures: []
```

The uppercase case clears validation completely.

## Root cause

`src/repair/url-safety.ts:7`

```ts
const URL_PATTERN = /https?:\/\/[^\s)\]"']+/g;   // no `i` flag
```

and its mirror at `src/repair/review-results.ts:592`. The host check downstream is
already case-correct — `new URL(value).hostname` is lowercased by the URL parser —
so only the scheme match is at fault.

## Expected behavior

Scheme case must not affect redaction. `HTTPS://vercel.com/x` and
`https://vercel.com/x` should both become `<external link>`, and both should be
rejected by `evidenceHasExternalUrl`. Legitimate `HTTPS://github.com/...` links
must keep working.

## Suggested fix

Add the `i` flag to both patterns together, preserving the lockstep invariant the
file comment already requires. The existing test in
`test/repair/url-safety.test.ts` that mirrors the validator regex locally should
get the flag too, so the mirror cannot drift.

## Existing coverage gap

`test/repair/url-safety.test.ts` covers hosts (`gist.github.com`,
`raw.githubusercontent.com`, `vercel.com`), non-string entries, and idempotency,
but never varies the scheme case — which is why this was not caught.

## Environment

| | |
|---|---|
| Node | v23.7.0 (repo `engines.node` wants `>=24`) |
| Baseline | `pnpm run test:no-build` green before investigation |

dwin-gharibi:
Closes #1043.

Fixes an issue where a non-GitHub link written with an uppercase URL scheme
(`HTTPS://host/path`) would be published verbatim in ClawSweeper's durable review
comment instead of being redacted to `<external link>`.

Worker evidence is model-authored text that ClawSweeper publishes to GitHub.
`sanitizeResultEvidence` exists to guarantee that no non-`github.com` link ever
reaches a published comment. That guarantee did not hold for uppercase schemes,
and the result validator that is meant to backstop it did not hold either — so a
worker result carrying such a link validated as `passed` and shipped.

Because GitHub's autolinker is case-insensitive, the published `HTTPS://…` text
renders as a live, clickable external link. The redaction the control promises
simply did not happen for that input.

## Why This Change Was Made

Both layers matched the scheme case-sensitively:

- `src/repair/url-safety.ts` — `URL_PATTERN` (the sanitizer)
- `src/repair/review-results.ts` — `evidenceHasExternalUrl` (the validator)

`url-safety.ts` carries a header comment requiring these two patterns to stay
identical, which is precisely why the blind spot was symmetric rather than caught
by the second layer. The fix adds the `i` flag to **both** together, preserving
that lockstep invariant, and extends the comment to record why the flag is load
bearing so a future edit does not drop it.

The downstream host check was already correct — `new URL(value).hostname` is
lowercased by the URL parser — so only the scheme match needed changing. The
allowed-host set (`github.com` only) and the URL terminator character class are
untouched.

Non-goals: this does not change any other URL matcher in the repo. Several
unrelated `github.com`-specific patterns elsewhere remain case-sensitive by
design and are out of scope.

## User Impact

Maintainers and contributors reading a ClawSweeper review comment no longer see
live external links that the sanitizer was supposed to remove; those links now
render as `<external link>` regardless of how the model capitalized the scheme.
Legitimate `HTTPS://github.com/...` references are preserved unchanged.

No configuration, workflow, or data-contract change. No migration.

**OpenClaw Bay:** not affected. Bay is an observer-only status surface and does
not render worker `result.json` evidence strings; its only use of the word
"evidence" is for terminal-outcome proof in the tide buffer, which this change
does not touch. No Bay update or Bay proof is needed.

**Release note:** `fix: redact external links in review evidence regardless of URL scheme case`.
No `CHANGELOG.md` edit is included — happy to add one if this repo's release
policy wants it for this change.

### Diff

```
 src/repair/review-results.ts   |  5 ++++-
 src/repair/url-safety.ts       | 13 +++++++-----
 test/repair/url-safety.test.ts | 47 ++++++++++++++++++++++++++++++++++++++++--
 3 files changed, 57 insertions(+), 8 deletions(-)
```

Plus a new proof package under `docs/proof/evidence-url-scheme-case/`.

### Focused tests

`node --test test/repair/url-safety.test.ts` → **14/14 pass**.

Three tests are new or extended:

- `sanitizeEvidenceText replaces external URLs regardless of scheme case` — covers
  `HTTPS`, `Https`, `hTTpS`, `HTTP`, `Http`, `hTtP`.
- `sanitizeResultEvidence redacts uppercase-scheme URLs on every evidence field` —
  `actions[].evidence`, `needs_human`, and all four `merge_preflight` fields.
- `sanitized evidence does not trigger evidenceHasExternalUrl regex` — the existing
  test mirrors the validator regex locally; the mirror now carries the `i` flag too,
  so it cannot silently drift from the real one, and uppercase inputs were added.

A fourth test, `sanitizeEvidenceText keeps github.com URLs with an uppercase scheme`,
is a regression guard: it passes both before and after, and exists to prove the new
case-insensitive pattern does not start discarding valid GitHub links.

### Red/green

Verified by reverting **only** the compiled `dist/repair/url-safety.js` pattern back
to `/g` and re-running the focused suite:

| build | result |
|---|---|
| pre-fix pattern | **3 fail**, 11 pass — the 3 new/extended tests |
| post-fix pattern | **14 pass**, 0 fail |

No pre-existing assertion changed behavior in either direction.

## Real Behavior Proof

**Claim.** Worker evidence is stripped of every non-`github.com` `http(s)` URL
before publication, and the result validator rejects any that survive, regardless
of the case of the URL scheme.

**Exercised surface.** Both shipped surfaces, against the built `dist/`:
1. `sanitizeResultEvidence()` in `dist/repair/url-safety.js` — the pre-publication
   mutation applied to a worker `result.json`.
2. `dist/repair/review-results.js` run as a **real CLI subprocess**, exactly as
   `dist/repair/run-worker.js` invokes it (`run-worker.ts:373`).

**Scenario / fixture.** `docs/proof/evidence-url-scheme-case/run-proof.mjs` builds a
throwaway run directory containing a `cluster-plan.json` and a *fully valid*
`result.json` (correct `mode`, `idempotency_key`, `target_kind`, and a
`target_updated_at` matching the preflight item). The only variable between the two
cases is the scheme case of one external URL in `actions[0].evidence`. Because the
fixture is otherwise valid, the external-URL failure is the only possible validator
failure. `attacker.example` is a reserved documentation name; the proof makes no
network calls.

Docker-backed Crabbox `local-container` lease, on the current head:

```bash
crabbox run \
  --provider local-container \
  --no-hydrate \
  --timing-json \
  --artifact-glob '.artifacts/evidence-url-scheme-proof/**' \
  --script docs/proof/evidence-url-scheme-case/run-proof.sh
```

| field | value |
|---|---|
| crabbox | `0.15.0` |
| lease | `cbx_9546d1539ae1` (`brisk-krill`) |
| run | `run_f0123d90f1c3` |
| artifact | `.crabbox/runs/run_f0123d90f1c3/run_f0123d90f1c3-artifacts.tgz` |
| exit | `0` |

**Observed result.**

Before (`dist/` built from `0588bda9`):

```
sanitizer  lowercase → proof: <external link>                          REDACTED
sanitizer  uppercase → proof: HTTPS://attacker.example/exfil?...       LEAKED
validator  lowercase → exit 1  ["#1 evidence contains non-GitHub external URL"]  REJECTED
validator  uppercase → exit 0  status "passed"  failures: []                     ACCEPTED
RESULT: FAIL
```

After:

```
sanitizer  lowercase → proof: <external link>                          REDACTED
sanitizer  uppercase → proof: <external link>                          REDACTED
validator  lowercase → exit 1  ["#1 evidence contains non-GitHub external URL"]  REJECTED
validator  uppercase → exit 1  ["#1 evidence contains non-GitHub external URL"]  REJECTED
RESULT: PASS
```

The pre-fix uppercase case clearing validation with `exit 0 / "passed"` is the
sharpest statement of the bug: the malicious evidence did not merely evade
redaction, it passed the backstop too.

**Artifact / trace.** `docs/proof/evidence-url-scheme-case/` contains the runnable
proof script and its contract (`README.md`). Both transcripts above are reproducible
by running the script against a pre-fix and post-fix build.

**Limits.** Covers the sanitizer and the validator for `http`/`https` scheme case
only. It does not exercise a live GitHub publication, Codex worker, or queue; the
"renders as a live link" claim rests on GitHub's documented case-insensitive
autolinking rather than a live post. It does not change the allowed-host set, the
URL terminator class, or any other URL matcher in the repo.

**Privacy.** The fixture uses the reserved documentation name `attacker.example`.
The proof makes no network calls, contacts no GitHub API, and performs no queue,
GitHub, or production mutation. No credential is present in the lease.

The **proof and focused suite now run on Node 24 inside the Crabbox
`local-container` lease** recorded above (`14/14` focused tests in-container).
The notes below concern only the *full* suite run on the macOS host.

`engines.node` is `>=24`. The newest Node available on the macOS host is **v23.7.0**,
and `AGENTS.md` warns that below Node 24 the notifier tests' retry paths misbehave
under the older `node:test` runner. On that host the suite is red *before* any
change, and the failure set is unstable between runs (one early run passed entirely;
later runs produced 3, then 21, then 22 distinct failures).

To separate signal from noise I ran the identical command on both trees:

| tree | `pnpm run test:no-build` |
|---|---|
| clean `main` (fix stashed, rebuilt) | 22 distinct failing tests |
| this branch | 22 distinct failing tests |

`comm` over the two sorted failure sets: **identical, zero failures unique to this
branch**. None of them touch `url-safety` or `review-results`.

Green on this branch:

|---|---|---|
| Crabbox `local-container` proof | Node 24 container | **PASS**, exit 0 |
| `test/repair/url-safety.test.ts` | Node 24 container | **14/14** |

**Please confirm `pnpm run check` on Node 24+ in CI before merging.** The changed
surface is proven on Node 24, but I have not produced a green *full-suite* run and
am not claiming one.

### DIFF
--- a/docs/proof/evidence-url-scheme-case/README.md
+++ b/docs/proof/evidence-url-scheme-case/README.md
@@ -1,0 +1,80 @@
+# Evidence external-URL scheme-case proof contract
+
+## Claim
+
+Worker evidence is stripped of every non-`github.com` `http(s)` URL before it is
+published, and the result validator rejects any that survive — regardless of the
+**case of the URL scheme**. Before this change, `HTTPS://host/path` matched
+neither the sanitizer nor the validator, so it was published verbatim as a live
+external autolink and the result still validated as `passed`.
+
+## Exercised surface
+
+Both shipped surfaces that share the pattern, driven against the built `dist/`:
+
+1. **Sanitizer** — `sanitizeResultEvidence()` in `dist/repair/url-safety.js`.
+   This is the pre-publication mutation applied to a worker `result.json`.
+2. **Validator** — `dist/repair/review-results.js`, executed as a real CLI
+   subprocess exactly as `dist/repair/run-worker.js` invokes it
+   (`run-worker.ts:373`).
+
+The two must stay in lockstep: `url-safety.ts` carries a header comment
+requiring its pattern to match `evidenceHasExternalUrl` in `review-results.ts`,
+because a sanitizer that emits something the validator rejects would deadlock a
+worker run.
+
+## Controlled scenario and fixture
+
+`run-proof.mjs` builds a throwaway run directory in the OS temp dir containing a
+`cluster-plan.json` and a **fully valid** `result.json` — correct `mode`,
+`idempotency_key`, `target_kind`, and a `target_updated_at` that matches the
+preflight item. The only thing that varies between the two cases is the scheme
+case of one external URL in `actions[0].evidence`:
+
+- control: `https://attacker.example/exfil?data=secret`
+- subject: `HTTPS://attacker.example/exfil?data=secret`
+
+Because the fixture is otherwise valid, the external-URL failure is the *only*
+validator failure, so the verdict is unambiguous. `attacker.example` is a
+reserved-for-documentation name; the proof contacts no network.
+
+## Expected observation
+
+Post-fix, both cases behave identically:
+
+- sanitizer output is `proof: <external link>` — the host does not appear;
+- validator exits `1` with exactly `["#1 evidence contains non-GitHub external URL"]`.
+
+Pre-fix, the uppercase case reported `LEAKED` from the sanitizer and
+`exit 0 / status "passed"` with `failures: []` from the validator.
+
+A companion focused test asserts that legitimate `HTTPS://github.com/...` and
+`HTTPS://GitHub.com/...` links are still preserved, so the case-insensitive
+pattern does not begin discarding valid GitHub references.
+
+## Artifact and command
+
+```bash
+pnpm run build:repair
+node docs/proof/evidence-url-scheme-case/run-proof.mjs   # exit 0 = PASS
+```
+
+Focused tests:
+
+```bash
+node --test test/repair/url-safety.test.ts
+```
+
+Red/green was verified by reverting only the compiled `dist/repair/url-safety.js`
+pattern to `/g` and re-running the focused tests: 3 fail pre-fix, 14/14 pass
+post-fix, with no change to the pre-existing assertions.
+
+## Limits
+
+This proof covers the sanitizer and the validator for the `http`/`https` scheme
+case only. It does not exercise a live GitHub publication, Codex worker, or
+queue; the published-comment rendering claim rests on GitHub's documented
+case-insensitive autolinking rather than a live post. It does not change the
+allowed-host set (`github.com` only), the URL terminator character class, or any
+other URL matcher in the repo — several unrelated `github.com`-specific patterns
+elsewhere are deliberately untouched.
--- a/docs/proof/evidence-url-scheme-case/README.md
+++ b/docs/proof/evidence-url-scheme-case/README.md
@@ -54,10 +54,44 @@ pattern does not begin discarding valid GitHub references.
 
 ## Artifact and command
 
+Supported-environment run (Node 24, Crabbox `local-container`):
+
+```bash
+crabbox run \
+  --provider local-container \
+  --local-container-image node:24 \
+  --no-hydrate \
+  --timing-json \
+  --artifact-glob '.artifacts/evidence-url-scheme-proof/**' \
+  --script docs/proof/evidence-url-scheme-case/run-proof.sh
+```
+
+`run-proof.sh` installs the pinned pnpm into a user-writable prefix (the lease
+runs as the unprivileged `crabbox` user, so `corepack enable` cannot symlink into
+`/usr/local/bin`), builds the repair lane, runs `run-proof.mjs`, and then runs the
+focused suite.
+
+Host-only quick check:
+
 ```bash
 pnpm run build:repair
 node docs/proof/evidence-url-scheme-case/run-proof.mjs   # exit 0 = PASS
 ```
+
+## Provenance
+
+- provider: Crabbox `local-container` (Docker/OrbStack)
+- crabbox: `0.15.0`
+- image: `node:24` @ `sha256:934240a162082fd8b8a2f90cd5114446443f1eba1c5378f6687167ca405e6584`
+- container node: `v24.19.0` (satisfies `engines.node >= 24`)
+- lease: `cbx_9546d1539ae1` (`brisk-krill`)
+- run: `run_f0123d90f1c3`
+- artifact: `.crabbox/runs/run_f0123d90f1c3/run_f0123d90f1c3-artifacts.tgz`
+- result: exit `0`; sanitizer REDACTED both schemes, validator REJECTED both;
+  focused suite `14/14`
+- privacy: the fixture uses the reserved documentation name `attacker.example`.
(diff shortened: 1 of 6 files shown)


dwin-gharibi:
@clawsweeper re-review

PR #1046, opened 2026-08-06
Fix/scheduler duplicate candidate stranding

### LINKED ISSUE #1045: Duplicate due candidates end batch selection early and strand due items
## Summary

`selectDueCandidates` can end batch selection early and strand due items while
review capacity is still free. A single duplicated entry in the due list is
enough. Duplicates are a normal GitHub pagination artifact, so this silently
under-fills real sweeps.

## Impact

The sweep quietly reviews fewer items than it has capacity for. There is no
error, no warning, and no telemetry signal — `dueBacklog` still reports the full
backlog while `candidates` is short, so the run looks like a normal
capacity-limited sweep rather than a truncated one.

In the worst shape (the weight-1 `weekly_issue` bucket) **one** duplicate strands
everything queued behind it. Observed with capacity 4 and three distinct due
items:

- no duplicate → selects `#1, #2, #3`
- one duplicate of `#1` → selects `#1` only

Because the stranded items stay due, they are re-selected on a later sweep, so
the effect is delay and wasted capacity rather than permanent starvation. But it
compounds: the weekly-coverage SLO is the outer deadline for canonical records,
and a repeatedly truncated batch eats the headroom that lane is meant to protect.

## Root cause

`src/scheduler-policy.ts:390-400` (and the same shape in `takeWeighted`,
`324-338`):

```ts
const take = (candidate) => {
  if (!candidate || selected.length >= capacity) return;
  const key = schedulerItemKey(candidate.item.repo, candidate.item.number);
  if (selectedKeys.has(key)) return;        // duplicate dropped silently
  selectedKeys.add(key);
  selected.push(candidate);
};

while (selected.length < capacity) {
  const before = selected.length;
  for (const [bucket, weight] of SCHEDULER_BUCKET_WEIGHTS) {
    const candidates = buckets.get(bucket);
    if (!candidates?.length) continue;
    for (let i = 0; i < weight && candidates.length && selected.length < capacity; i += 1) {
      take(candidates.shift());
    }
  }
  if (selected.length === before) break;    // ← treats "hit a duplicate" as "drained"
}
```

`take()` collapses three distinct outcomes into "nothing happened": capacity
reached, bucket empty, and *duplicate skipped*. The outer loop's no-progress
guard then reads a pass that consumed only duplicates as "all buckets are
drained" and breaks, abandoning every still-selectable candidate behind it.

The dedup itself is correct and intentional — `selectedKeys` exists because the
author expects duplicates. Only the termination signal is wrong.

## Why duplicates are realistic

The `due` list is assembled by paging `fetchOpenItemPage(page)`
(`src/clawsweeper-review-planning-selection.ts:90-105`). GitHub's paginated issue
listing is sorted by `updated`, so an item touched mid-pagination shifts pages and
is returned on more than one page. This is a well-known property of that endpoint,
not an exotic edge case.

## Reproduction

Through the real planner path, `batchSize: 10`, three distinct due issues:

```
control  (pages [[#1,#2],[#3]])       -> [1, 2, 3]
subject  (pages [[#1,#2],[#1,#3]])    -> [1]        ← #2 and #3 stranded
```

Directly against `selectDueCandidates`, all in `weekly_issue`, capacity 4:

```js
due = [#1, #2, #3]           → [1, 2, 3]
due = [#1, #1, #2, #3]       → [1]
```

Higher-weight buckets fail the same way with more duplicates — `hot_issue`
(weight 4) needs eight copies to stall a full pass.

## Expected behavior

Selection should continue until the buckets are genuinely empty or capacity is
reached. A duplicate should be skipped without ending the drain.

## Suggested fix

Make the loop stop on candidates **drained** rather than candidates **selected**,
in both the final loop and `takeWeighted`. That is a one-line change per loop and
does not alter behavior for a due list with no duplicates.

A broader alternative — deduplicating `due` before bucketing — would also stop a
duplicate from consuming a weighted slot, but it changes which entry wins when two
copies carry different `review`/`bucket` fields, so it is a larger behavioral
change than this defect requires.

## Existing coverage gap

`test/scheduler-policy.test.ts` covers bucket weighting, priority protection,
bulk-filed ordering, weekly-coverage preselection, and floor backfill, but never
feeds a duplicated item into `selectDueCandidates` — which is why this was not
caught.

## Environment

| | |
|---|---|
| Node | v23.7.0 (repo `engines.node` wants `>=24`) |

dwin-gharibi:
Closes #1045.

## What Problem This Solves

Fixes an issue where a sweep would review fewer items than it had capacity for,
with no error and no telemetry signal, whenever the same item appeared twice in
the due list.

Duplicates are a normal GitHub pagination artifact: `GET /issues` is sorted by
`updated`, so an item touched between two page reads shifts pages and is returned
on both. `selectDueCandidates` already dedupes by `repo#number` — the dedup was
never the problem. The problem was that hitting a duplicate ended selection
entirely, stranding every candidate queued behind it while capacity sat free.

Through the real `plan` CLI with three distinct due PRs and `#1` listed on four
pages, the pre-fix sweep selected `[1]` where the fix selects `[1, 2, 3]` —
`dueBacklog` reported 6 in both cases.

The stranded items stay due, so the effect is delay and wasted capacity rather
than permanent starvation — but it compounds against the weekly-coverage SLO,
which is the outer freshness deadline for canonical records.

**Reachability, precisely.** A stall needs a whole weighted pass to consume
nothing but duplicates, so the threshold is per bucket. Derived by driving the
real `dueCandidate()` logic against a pre-fix build:

| bucket | weight | page entries for one item needed to strand |
|---|---:|---:|
| `hot_pull_request`, `activity`, `recent_issue` | 2 | **4** |
| `daily_pull_request` | 3 | 6 |
| `hot_issue` | 4 | 8 |
| `weekly_issue` | 1 | **unreachable** |

`weekly_issue` cannot reach it: "due" and "weekly-coverage-due" share the same
6-day threshold there, so such an item is always claimed by the coverage preselect
lane, which takes candidates in a plain loop. The candidate must also be
coverage-tracked and reviewed inside the window between its cadence (1 day) and
the 6-day coverage deadline; untracked items are recovered by the final loop.

I originally wrote that a single duplicate was enough. That was wrong — it came
from a hand-built `weekly_issue` candidate the real planner never emits in a
due-but-not-coverage-due state. The defect and the fix are unchanged; the
severity is narrower than I first stated, and this section is the correction.

## Why This Change Was Made

`take()` collapses three outcomes into "nothing happened": capacity reached,
bucket empty, and *duplicate skipped*. The weighted drain's no-progress guard then
read a pass that consumed only duplicates as "all buckets are drained" and broke
out of the loop.

The fix changes what the loop measures — candidates **drained** instead of
candidates **selected** — in the final drain loop and in `takeWeighted`, which has
the same shape. Two lines per loop.

Deliberately minimal. It does **not**:

- deduplicate the `due` list up front (that would change which entry wins when two
  copies carry different `review`/`bucket` fields — a larger behavioral change
  than this defect calls for);
- change bucket weights, ordering, comparators, or the dedup key;
- change behavior at all for a due list with no duplicates.

A duplicate still consumes one weighted slot in its pass; only the premature
`break` is removed. Termination is unchanged in kind: each pass either drains at
least one candidate or breaks, and `due` is finite.

## User Impact

Sweeps fill their configured batch size when the GitHub listing returns an item
twice, instead of silently truncating. Operators see the review throughput the
capacity settings imply; no configuration change is needed.

No data-contract, workflow, or schema change. No migration.

**OpenClaw Bay:** not affected. This changes how many candidates a sweep selects,
not any status, lifecycle, or telemetry shape Bay renders. `dueBacklog`,
`capacityReason`, and `planSelectionTelemetry` keep their existing meanings — the
selected count simply stops being wrong. No Bay update or Bay proof is needed.

**Release note:** `fix: stop duplicate due candidates from truncating the review batch`.
No `CHANGELOG.md` edit is included — happy to add one if this repo's release policy
wants it.

### Diff

```
 src/scheduler-policy.ts       |  19 +++++--
 test/scheduler-policy.test.ts | 129 ++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 144 insertions(+), 4 deletions(-)
```

Plus a new proof package under `docs/proof/scheduler-duplicate-candidate-stranding/`.

### Focused tests

`node --test test/scheduler-policy.test.ts` → **24/24 pass** (20 pre-existing + 4 new).

New tests:

- `duplicate due candidates do not strand the rest of the batch` — contract-level,
  plus assertions that duplicates are still deduped and capacity is still respected.
- `a run of duplicates larger than the bucket weight still drains` — eight entries
  in the weight-4 `hot_issue` bucket, spanning two full passes with no new
  selection in the second.
- `duplicates across separate buckets do not stall the weighted drain` — duplicates
  in two buckets simultaneously.
- `pagination duplicates do not strand candidates the real planner produces` —
  builds its candidate the way `dueCandidate()` does, asserting the derived bucket
  and due-ness first, so the regression guard is pinned to a shape production
  actually emits rather than a hand-built one.

### Red/green

Swapped **only** the compiled `dist/scheduler-policy.js` for a pre-fix build of the
same file, and re-ran with the new tests present:

| build | result |
|---|---|
| pre-fix | **4 fail**, 20 pass — exactly the 4 new tests |
| post-fix | **24 pass**, 0 fail |

All 20 pre-existing assertions were unchanged in both directions.

## Real Behavior Proof

**Claim.** `selectDueCandidates` fills the batch to capacity from the candidates it
is given, even when an item appears twice.

**Exercised surface.** The real `plan` CLI, end to end — **no application function
is substituted**:

```
node dist/clawsweeper.js plan
  -> createReviewPlanningSelection().selectCandidates()
  -> fetchOpenItemPage()  -> ghJsonLines() -> spawnSync -> `gh` subprocess
                          -> JSON-lines parse -> item mapping
  -> dueCandidate()       -> bucket / priority / due derivation
  -> selectDueCandidates()
```

Only the `gh` **binary** is replaced, standing in for GitHub's server. ClawSweeper's
own transport — subprocess spawn, argument construction, JSON-lines parsing,
pagination loop — runs unchanged. This addresses the previous review's finding that
the artifact "substitutes fetchOpenItemPage, so it does not exercise a real
transport client".

**Scenario / fixture.** Three distinct open PRs, `--batch_size 10` (far above the due
count, so capacity is never the limiting factor), `--max_pages 5`. The stub `gh`
returns `#1` on four consecutive pages, reproducing the `updated`-sorted pagination
race in which an item touched mid-scan is listed again on later pages:

```
page 1: [#1, #2]   page 2: [#1, #3]   page 3: [#1]   page 4: [#1]   page 5+: []
```

A prior report exists for each PR in the items dir, making them coverage-tracked —
the state in which the weighted drain, not the coverage preselect lane, owns
selection.

Docker-backed Crabbox `local-container` lease, on the current head:

```bash
crabbox run \
  --provider local-container \
  --no-hydrate \
  --timing-json \
  --artifact-glob '.artifacts/scheduler-duplicate-proof/**' \
  --script docs/proof/scheduler-duplicate-candidate-stranding/run-proof.sh
```

| field | value |
|---|---|
| crabbox | `0.15.0` |
| lease | `cbx_7c89c5c91f66` (`brave-lobster`) |
| run | `run_71734de37d18` |
| artifact | `.crabbox/runs/run_71734de37d18/run_71734de37d18-artifacts.tgz` |
| exit | `0` |

**Observed result.** In-container, on the current head:

```
selected   : [1,2,3]
expected   : [1,2,3]
dueBacklog : 6        (6 page entries for 3 distinct PRs)
deduped    : true
RESULT: PASS

focused suite: 24/24 pass
```

`summary.json` from the artifact tarball:

```json
{ "selected": [1,2,3], "expected": [1,2,3], "dueBacklog": 6,
  "scannedPages": 5, "duplicateFreeSelection": true, "pass": true }
```

Against a pre-fix build of `dist/scheduler-policy.js`, the identical fixture and
CLI invocation yields `shards[0].itemNumbers = [1]` — `#2` and `#3` stranded with
`dueBacklog` still 6.

The proof also asserts the selection stays deduplicated, so the fix cannot be
satisfied by simply letting the duplicate through.

**Two traps worth recording**, both documented in the proof README so the next
person does not silently reintroduce a vacuous proof:

1. `selectCandidates()` calls `selectDueCandidates()` without a `now` argument, so
   the scheduler uses the real `Date.now()`. An earlier fixture used a fixed past
   `reviewedAt`; every candidate then qualified as weekly-coverage-due and was taken
   by the plain `for (const candidate of weeklyCoverageDue) take(candidate)` lane,
   which never reaches the weighted drain — so it **passed on a pre-fix build and
   proved nothing.**
2. Items with **no prior report** take the untracked lane, where survivors are
   recovered by the final loop and nothing is stranded. The fixture therefore writes
   a prior report per item; without that, the proof is also vacuous.

**Artifact / trace.** `docs/proof/scheduler-duplicate-candidate-stranding/` contains
`run-proof.sh` (the Crabbox entry point), `run-proof.mjs` (host-only quick check),
and the contract README. The lease artifact tarball holds `plan.json`,
`summary.json`, `focused-tests.txt`, `install.log`, and `build.log`.

**Limits.** Covers batch selection only. The duplicate is injected at the `gh`
**binary** boundary rather than produced by a real pagination race, since that race
is timing dependent and not reproducible on demand — so GitHub's own server
behavior is modeled, not exercised. No live GitHub, Worker, or queue is contacted,
and the proof performs no mutation. The lease is a local Docker container, not a
cloud host, so it proves the Linux/Node 24 runtime but not cross-host isolation.

**Privacy.** The fixture uses only `openclaw/openclaw` public-shaped item numbers
and a synthetic `contributor` login. No credential is present in the lease; the
stub `gh` makes no network call.

The **proof and focused suite now run on Node 24 inside the Crabbox
`local-container` lease** recorded above (`24/24` focused tests in-container). The
notes below concern only the *full* suite run on the macOS host.

`engines.node` is `>=24`; the newest Node available on the macOS host is **v23.7.0**,
and `AGENTS.md` warns about test misbehavior below Node 24.

| run | tree | distinct failing tests |
|---|---|---|
| clean `main` | fix stashed, rebuilt | 20 |
| run 1 | this branch | 22 |
| run 2 | this branch | 21 |

**Two runs of the identical branch disagree with each other** — `comm -3` over
run 1 and run 2 shows three tests flipping. That is direct evidence of flakiness
rather than a regression.

The two tests that appeared only on the fix branch in run 1
(`comment webhook settles duplicate fast ack comments after dispatch`,
`webhook signature verification uses sha256 body hmac`) both **passed in run 2**,
and both pass **25/25, three runs in a row**, in isolation on *both* the clean tree
and this branch. A module-graph walk from `dist/repair/comment-webhook.js` confirms
`scheduler-policy` is **not reachable** from that code at all.

**No scheduler test failed in any run, on either tree.**

Green on this branch:

|---|---|---|
| Crabbox `local-container` proof | Node 24 container | **PASS**, exit 0 |
| `test/scheduler-policy.test.ts` | Node 24 container | **24/24** |

**Please confirm `pnpm run check` on Node 24+ in CI before merging.** The changed
surface is proven on Node 24, but I have not produced a green *full-suite* run and
am not claiming one.

### DIFF
--- a/docs/proof/scheduler-duplicate-candidate-stranding/README.md
+++ b/docs/proof/scheduler-duplicate-candidate-stranding/README.md
@@ -3,67 +3,133 @@
 ## Claim
 
 `selectDueCandidates` fills the review batch up to capacity from the candidates
-it is given, even when the same item appears twice in the due list. Before this
-change, a single duplicated entry could end selection early and strand every
-remaining candidate while capacity was still free.
+it is given, even when the same item appears more than once in the due list.
+Before this change, a run of duplicates could end selection early and strand
+every remaining candidate while capacity was still free.
 
 ## Exercised surface
 
-The shipped planner path, not the comparator in isolation:
+`run-proof.sh` drives the real `plan` CLI end to end — **no application function
+is substituted**:
 
 ```
-createReviewPlanningSelection(...).selectCandidates()
-  -> src/clawsweeper-review-planning-selection.ts:90-111
-  -> selectDueCandidates()   (src/scheduler-policy.ts)
+node dist/clawsweeper.js plan
+  -> createReviewPlanningSelection().selectCandidates()
+  -> fetchOpenItemPage()  -> ghJsonLines() -> spawnSync -> `gh` subprocess
+                          -> JSON-lines parse -> item mapping
+  -> dueCandidate()       -> bucket / priority / due derivation
+  -> selectDueCandidates()
 ```
 
-The only injected behavior is `fetchOpenItemPage`, which reproduces the real
-GitHub artifact that triggers this: `GET /issues` is sorted by `updated`, so an
-item touched between two page reads shifts pages and is returned on both. Every
-step after that is the real planner. The `selectedKeys` dedup set inside
-`selectDueCandidates` exists precisely because this case is expected.
+Only the `gh` **binary** is replaced, standing in for GitHub's server.
+ClawSweeper's own transport — subprocess spawn, argument construction,
+JSON-lines parsing, pagination loop — runs unchanged. The `selectedKeys` dedup
+set inside `selectDueCandidates` exists precisely because duplicate page entries
+are expected.
 
 ## Controlled scenario and fixture
 
-`run-proof.mjs` builds one planner per scenario over three distinct open issues
-(`#1`, `#2`, `#3`) with `batchSize: 10` — far above the number of due items, so
-capacity can never be the limiting factor.
+Three distinct open PRs, `--batch_size 10` (far above the due count, so capacity
+is never the limiting factor), `--max_pages 5`. The stub `gh` returns `#1` on four
+consecutive pages, reproducing the `updated`-sorted pagination race in which an
+item touched mid-scan is listed again on later pages:
 
-- **control** — two pages, no race: `[[#1, #2], [#3]]`
-- **subject** — two pages, `#1` shifted onto page 2 as well: `[[#1, #2], [#1, #3]]`
+```
+page 1: [#1, #2]   page 2: [#1, #3]   page 3: [#1]   page 4: [#1]   page 5+: []
+```
 
-Every candidate is placed in the `weekly_issue` bucket, whose weight is **1**, so
-one duplicate is enough to consume a whole weighted pass.
+Four entries are needed because these land in `hot_pull_request`, whose weight is
+2 — a stall requires one whole weighted pass to consume nothing but duplicates.
+See **Reachability** below for the per-bucket thresholds.
 
-### Timing detail that matters
+### Two details that are load bearing
 
-`selectCandidates()` calls `selectDueCandidates()` **without** a `now` argument,
-so the scheduler uses the real `Date.now()`. The fixture therefore anchors
-`reviewedAt` to `Date.now() - 1h`, which keeps the candidates *out* of the
-weekly-coverage preselect lane (that lane needs 6 days).
+Both of these silently turn the proof vacuous — it passes on a pre-fix build and
+demonstrates nothing. Do not "simplify" either one away.
 
-This is load bearing. An earlier version of this proof used a fixed past
-timestamp; every candidate then qualified as weekly-coverage-due and was taken by
-the plain `for (const candidate of weeklyCoverageDue) take(candidate)` loop, which
-never reaches the weighted drain. That version passed on a pre-fix build and
-proved nothing.
+1. **Clock.** `selectCandidates()` calls `selectDueCandidates()` **without** a `now`
+   argument, so the scheduler uses the real `Date.now()`. Candidates must be recent
+   enough to stay *out* of the weekly-coverage preselect lane (which needs 6 days).
+   An earlier version used a fixed past timestamp; every candidate then qualified as
+   weekly-coverage-due and was taken by the plain
+   `for (const candidate of weeklyCoverageDue) take(candidate)` loop, which never
+   reaches the weighted drain.
+
+2. **Prior reports.** Items with no existing report are *coverage-untracked* and go
+   through `takeWeighted`, whose cohort is a separate copy of the bucket lists — the
+   final loop then recovers the survivors and nothing is stranded. The fixture
+   therefore writes a prior report per item so the candidates are coverage-tracked
+   and the final weighted drain owns selection.
 
 ## Expected observation
 
-| build | control | subject |
-|---|---|---|
-| pre-fix | `[1, 2, 3]` | `[1]` — `#2` and `#3` stranded |
-| post-fix | `[1, 2, 3]` | `[1, 2, 3]` |
+`shards[0].itemNumbers` from the real `plan` CLI, same fixture and invocation:
 
-The proof also asserts the subject result stays deduplicated, so the fix cannot
-be satisfied by simply letting the duplicate through.
+| build | selected | dueBacklog |
+|---|---|---|
+| pre-fix | `[1]` — `#2` and `#3` stranded | 6 |
+| post-fix | `[1, 2, 3]` | 6 |
+
+The proof also asserts the selection stays deduplicated, so the fix cannot be
+satisfied by simply letting the duplicate through.
 
 ## Artifact and command
+
+Supported-environment run (Node 24, Crabbox `local-container`). `run-proof.sh`
+drives the **real `plan` CLI** with only the `gh` binary stubbed:
+
+```bash
+crabbox run \
+  --provider local-container \
+  --local-container-image node:24 \
+  --no-hydrate \
+  --timing-json \
+  --artifact-glob '.artifacts/scheduler-duplicate-proof/**' \
+  --script docs/proof/scheduler-duplicate-candidate-stranding/run-proof.sh
+```
+
+It installs the pinned pnpm into a user-writable prefix (the lease runs as the
+unprivileged `crabbox` user, so `corepack enable` cannot symlink into
+`/usr/local/bin`), builds, runs the planner fixture, and runs the focused suite.
+
+Host-only quick check of the selection logic (injects `fetchOpenItemPage` rather
+than the `gh` binary, so it does **not** exercise the transport):
 
 ```bash
 pnpm run build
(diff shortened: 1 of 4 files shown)


dwin-gharibi:
@clawsweeper re-review

PR #1036, opened 2026-08-05
fix(review): stop scheduled review storms on unchanged issues

goutamadwant:
Closes #1032.

## What Problem This Solves

Fixes an issue where unchanged issues could be reviewed repeatedly when scheduled hot or normal queue deliveries were handled as explicit reviews. Those duplicate runs generated repeated contributor notifications and consumed review capacity even when only ClawSweeper-owned labels or comments had changed.

## Why This Change Was Made

The exact-review worker now forwards the claimed queue decision's source action as a typed internal review input. Review preparation owns the source-action classification beside the existing explicit-dispatch decision: only `scheduled_hot_intake` and `scheduled_normal_backfill` are automatic, while commands, webhook events, legacy dispatches, unknown actions, and missing actions remain explicit.

## User Impact

Scheduled review lanes can reuse a verified structural receipt for unchanged items instead of starting another full review. Explicit reruns still force fresh work.

OpenClaw Bay is unaffected because this changes review invocation classification only; it does not alter observer schemas, status projections, or dashboard contracts.

## Evidence

The regression failed on current main because exact queued reviews reached review preparation with an item number but no scheduled source input, making them explicit and ineligible for receipt reuse. The focused runtime regression now exercises both scheduled actions, representative explicit actions, missing and unknown actions, the legacy compatibility flag, and non-exact selection.

Validation:

```text
pnpm run build:all
node --test test/review-preparation.test.ts test/sweep-workflow.test.ts test/review-structural-cache.test.ts test/review-semantic-cache.test.ts test/review-content-cache.test.ts test/repair/scheduled-review-enqueue.test.ts
pnpm run check
git diff --check origin/main...HEAD
```

Results: the focused runtime, workflow, scheduler, and cache suites passed on Node 24.16.0. The full check passed formatting, static policy checks, all TypeScript builds, lint, and the exercised test assertions, but its local test process retained an open handle after its last output and required termination; CI is the clean-environment confirmation for that full aggregate command.

## Real Behavior Proof

**Claim:** the shipped workflow forwards claimed queue provenance into the real review-preparation boundary, where both scheduled actions are cache-eligible and sampled non-scheduled actions remain explicit.

**Exercised surface:** the built review-preparation module and the current `event-review-apply` workflow at head `4196b8bf29072d938f6b4330094ec42c1e455e51`.

**Observed runtime behavior:**

```text
scheduled_hot_intake => automatic
scheduled_normal_backfill => automatic
issues_opened => explicit
exact_review_command => explicit
[...]

PR #1053, opened 2026-08-07
fix: scheduled reviews re-run on unchanged items because the review cache is unreachable

### LINKED ISSUE #1052: Scheduled review deliveries cannot reach the structural review cache, so unchanged items are re-reviewed and re-rated every pass
## Summary

The structural review cache cannot be reached on the lane that runs every scheduled review, so unchanged items are hydrated and re-reviewed in full on every cadence pass, and a fresh verdict is published each time.

## Mechanism

The exact-event workflow reserves the durable review lease in its own write-token step, so it must invoke the review command with `--skip-start-comment` (`.github/workflows/sweep.yml:1229`; `prepareReviewCommand` throws if a supplied lease arrives without it). That same flag decided `coordinationEnabled`, and `reviewStructuralCacheProbeDecision` consults `coordinationEnabled` before anything else (`src/review-structural-cache.ts:1219`), so the probe returned `coordination_disabled` and never ran.

Because the probe never ran, `preHydrationStructuralRecord` stayed `null`, and the seeding branch is guarded on that value, so no structural receipt was ever persisted for the next review either. The cache could not bootstrap even once.

Clearing that gate exposes two more conditions on the same path, and all three must go before a scheduled pass can reuse anything:

- **`activity_changed`** — the reservation `reserve-review-lease` posts before the review command runs is itself the item's newest activity, landing after the sync markers the prior review recorded, so ClawSweeper's own write reads as reporter activity.
- **`target_changed`** — `reviewStructuralCacheDecision` compares the target repository's default branch head across reviews (`src/review-structural-cache.ts:1280`). That head advances for reasons unrelated to the item, so on a repository whose default branch moves faster than the review cadence the key can never match and the cache reuses nothing.

This sits directly below the gate that https://github.com/openclaw/clawsweeper/pull/1036 fixed. That PR made `scheduled_hot_intake` and `scheduled_normal_backfill` non-explicit so scheduled lanes could "reuse a verified structural receipt for unchanged items"; the flag that produces the next gate is on the adjacent line of the same `pnpm run review` invocation (`sweep.yml:1229` and `:1234`).

## Observed

Run https://github.com/openclaw/clawsweeper/actions/runs/31131759207 reviewed openclaw/clawscan#7 starting 2026-08-06T23:36:45Z from clawsweeper head `3f368a3e394d76c31584fce700cee9a62485cb66`, after #1036 merged at 18:01Z:

```text
[review] shard=0/1 start-comment=skipped #7
[review] shard=0/1 complete reviewed=1 cache_hits=0 structural_cache_checks=1
  structural_cache_hits=0 structural_cache_revalidations=0 semantic_cache_checks=1
  semantic_cache_hits=0 semantic_cache_ineligible=1 semantic_cache_revalidations=0
  content_cache_hits=0 hydrations=1
```

That pull request has one commit dated 2026-08-03T16:04:07Z (head `bb644fe1646cbd7f74c81947259ec5b042fad776`), its repository's `main` last moved 2026-08-03T16:01:55Z, its durable review comment has existed since 2026-07-01, and the delivery was `scheduled_hot_intake`. So `pull_head_changed`, `target_changed`, `missing_review`, and `explicit_dispatch` are all excluded by public state, and the item was still hydrated and re-reviewed in full.

A separate scheduled delivery on openclaw/openclaw#119986 (run https://github.com/openclaw/clawsweeper/actions/runs/31131564193) published a report whose receipt fields are unseeded, which is the second half of the mechanism:

```text
review_structural_target_head_sha: unknown
review_structural_cache_hit: false
review_semantic_cache_hit: false
```

## Impact

Contributors see verdict churn on a frozen head. On https://github.com/openclaw/openclaw/pull/118777 the head `ed699913b2acf0ec158e705afa289cc2d4c59358` never moved and the body was never edited, yet re-scans walked the rating from `🦞 diamond` to `🦪 silver` and revoked `proof: sufficient` along with `status: 👀 ready for maintainer look`. Two of those re-reviews were scheduled deliveries (runs https://github.com/openclaw/clawsweeper/actions/runs/30955548730 and https://github.com/openclaw/clawsweeper/actions/runs/31049170523).

The `target_changed` condition is what makes this permanent on the busiest target: `openclaw/openclaw` advanced its default branch 100 times in the 21h11m ending 2026-08-07T00:06Z — about once every 13 minutes — against a daily cadence over 5,536 open items (2,150 pull requests, 3,386 issues). One unrelated commit is enough: appending a comment to a README, on an item about a different file, turns a reused verdict into a full hydration and Codex re-run.

Operators pay for the re-derivation. Read from clawsweeper run titles at 2026-08-07T00:12Z, the queue revision counters on long-lived items had reached `openclaw/openclaw#84599` at 389, `#40088` at 307, and `#38283` at 293.

## Scope note

One adjacent condition is deliberately **not** part of this report:

- `artifact_retention_recovery` deliveries are still classified explicit and bypass the cache one gate earlier. Whether a publication-recovery requeue should re-derive a verdict rather than republish the recorded one is a separate question.

masatohoshino:
Closes #1052.

## What Problem This Solves

Fixes an issue where ClawSweeper re-reviews an unchanged item on every scheduled pass and can publish a different verdict each time, because the structural review cache is unreachable on the lane that runs those passes.

Terminology, because the two are easy to conflate: the *exact-event workflow* is the `sweep.yml` job that runs every review for a target repository, and a *scheduled delivery* is one class of payload that job carries — the queue's cadence work, as opposed to a webhook event or a maintainer command.

That workflow reserves the durable review lease in its own write-token step, so it must invoke the review command with `--skip-start-comment`. The same flag decided `coordinationEnabled`, which `reviewStructuralCacheProbeDecision` consults before anything else, so scheduled deliveries returned `coordination_disabled` and re-ran full hydration plus Codex. This is a property of the invocation rather than a sampled rate: `.github/workflows/sweep.yml:1229` passes the flag unconditionally, so no delivery through that job could have had a different value.

Because the probe never ran, `preHydrationStructuralRecord` stayed `null` and the seeding branch could not persist a receipt for the next review either. The cache could not bootstrap even once.

Clearing that gate exposed two more conditions on the same path. All three had to go before a scheduled pass could reuse anything:

- **`activity_changed`** — the reservation the workflow posts before invoking the command is itself the item's newest activity, landing after the sync markers the prior review recorded, so ClawSweeper's own write read as reporter activity.
- **`target_changed`** — the decision compared the target repository's default branch head across reviews. That head advances for reasons unrelated to the item, so where it moves faster than the review cadence the key can never match. `openclaw/openclaw` advanced it 100 times in the 21h11m ending 2026-08-07T00:06Z — about once every 13 minutes — against a daily cadence over 5,536 open items (2,150 pull requests, 3,386 issues), so every scheduled pass there re-ran a full review.

Contributors see this as verdict churn on a frozen head. On https://github.com/openclaw/openclaw/pull/118777 the head `ed699913b2acf0ec158e705afa289cc2d4c59358` never moved and the body was never edited, yet re-scans walked the rating from `🦞 diamond` to `🦪 silver` and revoked `proof: sufficient` together with `status: 👀 ready for maintainer look`. Two of those re-reviews were scheduled deliveries: runs https://github.com/openclaw/clawsweeper/actions/runs/30955548730 and https://github.com/openclaw/clawsweeper/actions/runs/31049170523. The cost side is visible in the queue revision counters carried in clawsweeper run titles; read at 2026-08-07T00:12Z, `openclaw/openclaw#84599` stood at 389, `#40088` at 307, and `#38283` at 293.

## Why This Change Was Made

`coordinationEnabled` is now true when a reserved lease is supplied, the structural cache path claims that reserved lease instead of posting a second start comment, the validated reservation covers its own activity, and the target branch head is no longer compared across reviews.

Those are one change, not two. Widening the coordination decision alone would send the cache path into `postReviewStartStatusComment`, which is exactly the second lease that `--skip-start-comment` exists to prevent — `prepareReviewCommand` throws today if a supplied lease arrives without that flag. Making the cache path claim the reserved lease is what makes the widening safe.

Two things change together in `src/clawsweeper-review-command-workflow.ts`. The structural-cache probe calls read the preparation's coordination decision, and the cache's lease acquisition claims a supplied lease when there is one. The reservation is now validated *before* the cache decision rather than after it, and its timestamp is passed to that decision as `ownedReservationUpdatedAt`, because the workflow posts that comment before invoking this command — so on this lane the reservation is always the item's newest activity, and `activityCoveredByReview` would otherwise read ClawSweeper's own write as reporter activity and return `activity_changed`. Only the reservation this run has already validated as its lease winner may widen that window, and it widens it to its own timestamp and no further: activity after the reservation still forces hydration. The freshness and winner checks are not new — the hydrated claim path already ran them. They move into a helper because the cache path now needs the identical check before hydration; the alternative is a second copy of the same lease-election logic, which is the thing most likely to drift. The extraction moves those statements without changing their order or conditions.

The target branch head is dropped from the cross-review comparison rather than made conditional, so no new constant enters the code. The record still carries `targetHeadSha`, and a single run still uses it to check that `main` held still across its own pre- and post-hydration probes; what goes is only the comparison against the previous review. Periodic re-review is bounded by the ceiling the probe already enforces, `REVIEW_STRUCTURAL_CACHE_MAX_AGE_DAYS`: a recorded verdict is reused for at most that long before a full review runs regardless of what `main` did. The cost is that obsolescence caused by `main` — an `implemented_on_main` landing in files the item never touched — is noticed at that ceiling rather than on the next daily pass.

When the reserved lease no longer describes the item, the cache path takes the same action the hydrated path already took for that condition: increment `leaseAcquisitionFailures`, record the item in `leaseAcquisitionFailureDetails`, and let the run fail at the end so the recovery lane requeues it, rather than silently dropping the item from the batch. That branch is intended parity with the existing hydrated behavior and is not covered by a new test; it is called out here so a reviewer checks it by reading, and the shared helper is what makes the two paths comparable at a glance.

### Which deliveries this affects

`reviewStructuralCacheProbeDecision` checks `explicitDispatch` before `coordinationEnabled`, and `AUTOMATIC_REVIEW_SOURCE_ACTIONS` in `src/clawsweeper-review-preparation.ts` contains exactly `scheduled_hot_intake` and `scheduled_normal_backfill` after https://github.com/openclaw/clawsweeper/pull/1036.

- Those two source actions cleared `explicit_dispatch` after #1036 and then stopped at `coordination_disabled`. They carry the cadence buckets in `docs/scheduler.md` — activity-driven hourly, recent-item and pull-request daily, older-issue weekly — and they are what this change fixes.
- Every other source action, including webhook events, `@clawsweeper` commands, `legacy_dispatch`, and unknown or missing actions, is still classified explicit by that same set and still short-circuits one gate earlier, unchanged by this PR. That is the intended behavior for a requested review.

#118777's most recent re-review is therefore not covered here: it arrived with `source_action: artifact_retention_recovery` (run https://github.com/openclaw/clawsweeper/actions/runs/31081165269), which is not in the automatic set. Whether a publication-recovery requeue should re-derive a verdict rather than republish the recorded one is a separate question.

### Why a pull request's base does not move with the default branch

A reasonable worry about dropping the target-head comparison is that pull requests would still miss, because `sourceRevision` hashes `snapshot.pull.baseSha` and a base change would invalidate them anyway. That does not happen, because `baseRefOid` is pinned at pull-request synchronisation rather than tracked to the base branch:

| | value | dated |
| --- | --- | --- |
| openclaw/openclaw#118777 `baseRefOid` | `d2c8fd2e710f9f20f08938502063170e8554c228` | 2026-08-03T16:46:14Z |
| openclaw/openclaw `main` head at time of writing | `9aa24e8f2e822290a4c4f785d48031f93c3f9fbb` | 2026-08-07T05:02:50Z |

That pull request's base has not moved in three and a half days, across roughly 450 commits on `main`. Four other open pull requests updated within the same hour report four different base oids (`65d2222a`, `f87d8bb7`, `a6b20789`, `f87d8bb7`); if the field tracked the branch they would all equal the current head.

So for a pull request, an advancing default branch changes neither `baseSha` nor the source revision that hashes it. What still invalidates is the pull request actually being synchronised onto a new base, which is a real change to the item. Both directions are pinned by a regression in `test/review-structural-cache.test.ts`.

### Non-goals

- This does not change the semantic or content cache stages, the review verdict itself, or the explicit-dispatch classification.
- This introduces no new constant. Staleness stays bounded by the ceiling the probe already enforces (`REVIEW_STRUCTURAL_CACHE_MAX_AGE_DAYS`): a recorded verdict is reused for at most that long before a full review runs regardless of what `main` did.
- OpenClaw Bay is unaffected: this changes review invocation coordination only, not observer schemas, status projections, or queue contracts.

## User Impact

A scheduled delivery can now reach and use the structural cache instead of being turned away before it is consulted. Where the default branch advances between passes — the case that motivated this — no scheduled pass previously reused a verdict at all, so every one of them re-ran a full review. Where such a pass satisfies every existing eligibility check the cache already enforced — a completed keep-open review under the same policy and model and less than 14 days old, an unchanged bounded source revision, an unchanged target head and pull head, and a claimable lease — it reuses the recorded verdict rather than re-running Codex. This change adds no eligibility relaxation; it only stops the lane from failing the coordination precondition by construction.

Runs that create their own start comment behave exactly as before: the non-supplied-lease branch still calls `postReviewStartStatusComment` with the same arguments. Explicit dispatches, maintainer requests, and `@clawsweeper re-review` still force fresh work.

What the evidence below does and does not establish:

- **Established:** with the workflow's own flags, the probe now returns `hit` where it returned `coordination_disabled`, on the shipped modules.
- **Established against a live GitHub item:** the reserved-lease claim runs and elects the reserved lease, the structural receipt is persisted where it was previously written as `unknown`, and a subsequent scheduled delivery reuses the recorded verdict with `hydrations=0` — no GitHub hydration and no Codex call.
- **Established from public production state:** a scheduled delivery on an unchanged pull request, on a repository whose `main` was also unchanged and with a review already on file, was still hydrated and re-reviewed in full.
- **Not established:** behaviour on a production target. The live runs use one repository I own; that the same code runs on production targets is inferred from the shipped path being identical, not observed.

## Evidence

One caveat first: `pnpm run check` as a single aggregate command is **not** clean in my local sandbox. It also runs `test/repair/target-validation.test.ts`, which fails 13 tests there because the sandbox cannot create `/tmp/node-coverage-*` or spawn its contained commands. Those same 13 fail on unmodified `main` at `2eb1787e0d183a84f29e84614b84f228037ba69f` in the same sandbox, so they are environmental rather than caused by this change. That suite is untouched here, and this PR's own CI check is where it executes in a clean environment.

```text
pnpm run build:all
node --test test/review-preparation.test.ts test/review-structural-cache.test.ts \
  test/review-semantic-cache.test.ts test/review-content-cache.test.ts \
  test/sweep-workflow.test.ts test/local-review.test.ts \
  test/local-range-review.test.ts test/command.test.ts
pnpm run check:static
pnpm run lint
node docs/proof/review-cache-coordination-gate/run-proof.mjs
```

With that suite excluded, the focused run is clean: 269 tests, 0 failures on Node 24.18.0. Static policy checks, formatting, all three TypeScript builds, and lint pass.

Two tests are added:

- `test/review-preparation.test.ts` — "a reserved review lease keeps coordination enabled under `--skip-start-comment`", covering the production pairing plus the three cases that must keep their existing answer.
- `test/review-structural-cache.test.ts` — "an advanced target head alone does not force hydration" and "a pull request survives target-branch movement but not its own base sync", covering both item kinds plus the age ceiling that still bounds reuse.
- `test/review-structural-cache.test.ts` — "the exact-event lane's own invocation reaches the structural receipt", composing the shipped preparation decisions into the real probe using the workflow's own flags.

Both are anchored to this change. Restoring the previous expression in `isReviewCoordinationEnabled` (`return !skipStartComment;`), rebuilding, and re-running `node --test test/review-preparation.test.ts test/review-structural-cache.test.ts` fails exactly those two and no existing test: 48 tests, 46 pass, 2 fail.

## Real Behavior Proof

**Claim:** a scheduled delivery of an unchanged item reuses its recorded verdict without hydrating GitHub or calling Codex, including when the target branch has advanced since that verdict was recorded.

**Level:** module-level proof against the built `dist/` output using production argument values, plus a production observation that narrows the pre-fix reason by elimination. Not a live end-to-end trace; see Limits.

**Exercised surface:** the built `parseArgs`, `suppliedReviewStartLeaseFromArgs`, `isExplicitReviewDispatch`, `isReviewCoordinationEnabled`, and `reviewStructuralCacheProbeDecision`. Post-fix captures taken at `5ba31b23`; pre-fix capture taken from a clean checkout of `2eb1787e0d183a84f29e84614b84f228037ba69f`.

**Scenario and fixture:** the harness parses the flag list out of `.github/workflows/sweep.yml` rather than restating it, and aborts if production stops passing `--skip-start-comment`, `--review-lease-owner`, and `--review-source-action` together. Every argument value comes from one real delivery — openclaw/clawscan#7, reviewed by clawsweeper run https://github.com/openclaw/clawsweeper/actions/runs/31131759207 with `sourceAction: scheduled_hot_intake` and the durable lease marker that run published — so the fixture is not assembled from different items. No production module is stubbed or replaced; the one synthetic value is the prior review record standing in for a completed keep-open review. The probe decision is the last thing the harness observes, which is why its final line is labelled as an implication.

**Command and environment:** `pnpm run build && node docs/proof/review-cache-coordination-gate/run-proof.mjs`, Node 24.18.0, Linux.

**Observed result** — pre-fix, at `2eb1787e0d183a84f29e84614b84f228037ba69f`:

```text
build                     pre-fix (upstream/main)
skipStartComment          true
suppliedReviewLease       {"owner":"github-run-31131759207-1","commentId":5210008452}
explicitDispatch          false
coordinationEnabled       false
structural probe          {"hit":false,"reason":"coordination_disabled"}
implied next step (not run here)  full hydration + Codex
```

Post-fix, same command, same workflow flags:

```text
build                     post-fix
skipStartComment          true
suppliedReviewLease       {"owner":"github-run-31131759207-1","commentId":5210008452}
explicitDispatch          false
coordinationEnabled       true
structural probe          {"hit":true,"reason":"hit"}
implied next step (not run here)  reuse the recorded verdict
```

`explicitDispatch` is already `false` on both sides, which is #1036 working as intended.

**Production observation for the pre-fix side:** run https://github.com/openclaw/clawsweeper/actions/runs/31131759207 reviewed openclaw/clawscan#7 starting 2026-08-06T23:36:45Z, from clawsweeper head `3f368a3e394d76c31584fce700cee9a62485cb66` — after #1036 merged at 2026-08-06T18:01Z. The other reasons the probe could have returned are excluded by public state at that moment:

| Candidate reason | Excluded by |
| --- | --- |
| `explicit_dispatch` | `sourceAction: scheduled_hot_intake`, which #1036 made automatic |
| `missing_review` / `incomplete_review` | the durable review comment on that PR has existed since 2026-07-01 |
| `target_changed` | `openclaw/clawscan` `main` last moved 2026-08-03T16:01:55Z, three days earlier |
| `pull_head_changed` | the PR has one commit, dated 2026-08-03T16:04:07Z; head `bb644fe1646cbd7f74c81947259ec5b042fad776` |

That run logged `--skip-start-comment` taking effect and no cache reuse:

```text
[review] shard=0/1 start-comment=skipped #7
[review] shard=0/1 complete reviewed=1 cache_hits=0 structural_cache_checks=1
  structural_cache_hits=0 structural_cache_revalidations=0 semantic_cache_checks=1
  semantic_cache_hits=0 semantic_cache_ineligible=1 semantic_cache_revalidations=0
  content_cache_hits=0 hydrations=1
```

A pull request whose head had not moved in three days, on a repository whose `main` had not moved in three days, with a review record already on file, was hydrated and re-reviewed in full.

### End-to-end run against a live GitHub item

The captures above stop at the probe decision, so the reserved-lease claim and
the receipt-persistence path were driven through the shipped `review` command
against a real issue in a repository I own
(<https://github.com/masatohoshino/clawsweeper-cache-proof> issue 1). Nothing ran
against a production target. Leases came from `pnpm run reserve-review-lease`,
the same command `sweep.yml` uses, so no marker was hand-written. The runner was
the branch at `5ba31b23`, built from a copy of that tree whose only difference is
one added `generic_fallbacks` entry so my own account is onboarded the way
`openclaw/*` and `steipete/*` already are; no source file differs and that entry
is not in this PR.

**The claim branch this PR adds runs.** Emitted only by that branch:

```text
[review] shard=0/1 structural-cache-start-comment=reserved #1
```

Reaching it requires the probe to pass coordination, the structural decision to
hit, and the claim to elect the reserved lease through the same freshness checks
the hydrated path uses. On `main` the same invocation stops at
`coordination_disabled` first.

**The receipt is persisted.** Two runs differing only in whether a prior
completed review existed:

```text
run 1 report:  review_structural_target_head_sha: unknown
run 2 report:  review_structural_target_head_sha: b575dc7ad47e96e62288ab26eb84d5c5b5b22adf
"structural_cache_revalidation_reasons": { "hydrated_anchor_match": 1, "verdict_input_match": 1 }
```

That reproduces the production symptom — reports published with
`review_structural_target_head_sha: unknown` — and then clears it.

**The cache hits, with no hydration and no Codex.** A run that stopped at
`activity_changed` before the reservation was recognised now reuses the recorded
verdict. Same item, same reserved lease:

```text
[review] shard=0/1 structural-cache-start-comment=reserved #1
[review] shard=0/1 cache-hit structural-unchanged skip-hydration-model #1 (1/1)
[review] shard=0/1 complete reviewed=1 cache_hits=1 structural_cache_checks=1
  structural_cache_hits=1 structural_cache_revalidations=1 ... content_cache_hits=0
  hydrations=0
```

The carried report records `review_cache_hit: true` and
`review_structural_cache_hit: true`. `hydrations=0` is the point: 8 seconds end to
end, against 70-150 seconds for the full reviews that preceded it.

**An unrelated commit on the target branch no longer costs a full review.** The runs above kept the target repository's `main` still. Advancing it by one commit that has nothing to do with the item — a comment appended to `README.md`, where the item is about `src/slug.js` — was enough to lose the hit before this change:

```text
structural_cache_reasons: { "target_changed": 1 }
cache_hits=0  hydrations=1                      (103 s, full hydration + Codex)
```

Same item, same reserved-lease flow, same advanced `main`, after the change:

```text
[review] shard=0/1 structural-cache-start-comment=reserved #1
[review] shard=0/1 cache-hit structural-unchanged skip-hydration-model #1 (1/1)
[review] shard=0/1 complete reviewed=1 cache_hits=1 structural_cache_checks=1
  structural_cache_hits=1 ... content_cache_hits=0 hydrations=0
```

```json
"structural_cache_reasons": { "hit": 1 }
```

8 seconds instead of 103, with no Codex call.

**Not attributed to this PR.** An earlier run in the same sequence skipped the
model through the *content* stage (`cache-hit content-unchanged skip-model`),
whose coordination comes from `Boolean(acquiredReviewLease)` and is already
satisfied on `main` by the pre-existing post-hydration claim. It is mentioned
because it happened, not as evidence. The durable review comment these runs
needed was published with `apply-decisions --sync-comments-only --skip-dashboard`,
the documented local apply repro, not by the production publication lane.

### Controlled environment run (Crabbox `local-container`)

The runs above executed on my host. This repeats the focused validation inside a Docker-backed Crabbox lease at the current head, which is the controlled envelope `AGENTS.md` asks for on code-bearing changes.

| Field | Value |
| --- | --- |
| Provider | `local-container` (Docker) |
| Runtime | Docker 29.4.2, context `default` |
| Image | `ubuntu:26.04` @ `sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03` |
| Lease | `cbx_bfb55b2eb06c` (slug `cache-gate-proof3`, container `c3bc5926759d`, type `ubuntu_26.04`) |
| Head | `5ba31b2373e0c2bd132e0d116d9ce1a3dcb69764` — the current PR head |
| Node | v24.19.0, pnpm 11.10.0 |
| Artifact | `cbx_bfb55b2eb06c-artifacts.tgz`, 9147 bytes, sha256 `fdf8e568d7bd83e8ddaf8eeaab4de0da070e2a5f355478c6051a88b304ddaa02` |
| Captured | 2026-08-07T06:23:16Z |

The artifact tarball only exists on my machine, so the trace it holds is reproduced verbatim below. Nothing is elided; the three files are the whole capture.

`env.txt`:

```text
os=Ubuntu 26.04 LTS
node=v24.19.0
pnpm=11.10.0
dist_built=2026-08-07 06:22:16.422405452 +0000
target_changed_in_dist=0
--- source under test (sha256) ---
2d86fd44d523a3ea86abcbc480898f20231cf97cb71adb2612476b4bbbe2c51c  src/review-structural-cache.ts
57ac6aee69c6008be1382bfb1a7e18f68a1656628c625fc79cd7c0c52be8df38  src/clawsweeper-review-command-workflow.ts
e9a8179fd25a8f98ce438ba2e02c8fe615838c8be2c3512a9833ee64352411e3  src/clawsweeper-review-preparation.ts
7767a232e7ff6f1b1722ad5f0fe5000a2373751d9e20bb0736a769c3fcffe8e4  test/review-structural-cache.test.ts
captured_at=2026-08-07T06:23:16Z
```

`harness.txt`:

```text
build                     post-fix
workflow flags            --target-repo --target-dir --artifact-dir --batch-size --max-pages --codex-model --codex-reasoning-effort --codex-sandbox --codex-timeout-ms --item-numbers --readonly-openclaw --skip-start-comment --review-lease-owner --review-lease-comment-id --shard-index --shard-count --review-source-action
skipStartComment          true
suppliedReviewLease       {"owner":"github-run-31131759207-1","commentId":5210008452}
explicitDispatch          false
coordinationEnabled       true
structural probe          {"hit":true,"reason":"hit"}
implied next step (not run here)  reuse the recorded verdict
```

`tests.txt`, final block (`tests_exit=0`):

```text
ℹ tests 271
ℹ suites 0
ℹ pass 271
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 4837.506964
```

Crabbox syncs the working tree without `.git`, so the lease cannot report a commit id of its own. Rather than assert one, the capture hashes the files under test, and those hashes are checkable against this head's blobs by anyone with the branch.

```sh
for f in src/review-structural-cache.ts src/clawsweeper-review-command-workflow.ts          src/clawsweeper-review-preparation.ts test/review-structural-cache.test.ts; do
  printf '%s  %s\n' "$(git show 5ba31b23:$f | sha256sum | cut -d' ' -f1)" "$f"
done
```

All four match. The lease also built `dist/` itself — `dist/review-structural-cache.js` is stamped `2026-08-07 06:22:16 +0000` and `grep -c target_changed` over it returns `0` — so the code that ran is this head's, not a synced build.

This capture lives in the PR body rather than in a commit on purpose. Recording it in the repository advances the head past the head it names, which is how the previous two revisions of this section ended up identifying an older commit. `AGENTS.md` asks for the proof to be kept current in the PR body, and a body edit does not move the head. `docs/proof/review-cache-coordination-gate/README.md` carries the method, the commands, and the limits; this table is the capture.

Artifact contents are `env.txt`, `harness.txt`, and `tests.txt`. Scanning all three for `token|secret|password|ghp_|github_pat|api[_-]?key|bearer|authorization` returns 11 matches, every one a test name (for example "local-review scrubs both GitHub and GitHub Enterprise token aliases"). No credential, endpoint, or private URL is present; the lease never held GitHub credentials.

Envelope limits, stated so the run is reproducible rather than flattering: the lease is a local Docker container on my machine rather than the organization's AWS Crabbox capacity, and `.crabbox.yaml` targets `provider: aws` — this run overrode it with `--provider local-container`, the Linux path the provider documentation describes. The base image ships without `jq`, `zip`, `unzip`, and `gh`; installing them is part of the run, because without them 29 workflow-shell assertions fail on missing tooling rather than on behaviour. `--no-hydrate` is used because the repository's Actions hydration rejects `pnpm/action-setup@v6.0.9` under local Actions emulation. The live-GitHub end-to-end runs are not repeated inside the lease, which is kept free of GitHub credentials.

**Artifact:** `docs/proof/review-cache-coordination-gate/` — the runnable harness plus a README holding both captures and this observation.

**Limits:**

- No live claimed-delivery trace through the GitHub receipt-cache boundary. That requires the organization's App installation token, which only repository operators can exercise. #1036 disclosed the same limit. The production observation narrows the pre-fix reason by elimination from public state; it does not read the probe's returned reason directly.
- The proof establishes that the probe is reached. Lease claiming against live GitHub, receipt persistence, and reuse in the deployed lane are not exercised.
- No production review jobs were dispatched, no model capacity was consumed, and no contributor notifications were generated while validating this change.

Disclosure: AI was used to investigate the lane, write the fix, and review it.

### DIFF
--- a/docs/proof/review-cache-coordination-gate/README.md
+++ b/docs/proof/review-cache-coordination-gate/README.md
@@ -1,20 +1,28 @@
-# Proof: the exact-event lane could not reach its own structural receipt
+# Proof: scheduled reviews never reused a cached verdict
 
 ## Claim
 
-`.github/workflows/sweep.yml` reserves the durable review lease in its own
-write-token step, so it must invoke the review command with
-`--skip-start-comment`. That flag also decided `coordinationEnabled`, which
-`reviewStructuralCacheProbeDecision` consults before anything else. Scheduled
-deliveries therefore returned `coordination_disabled`, skipped the probe, and
-re-ran full hydration plus Codex on unchanged items — and because the probe
-never ran, `preHydrationStructuralRecord` stayed `null` and the seeding branch
-could never persist a receipt for the next review either.
+A scheduled delivery of an unchanged item re-ran full hydration and Codex every
+pass. Three conditions on the same cache-eligibility path each blocked reuse, and
+all three had to go for the outcome to change:
+
+1. **`coordination_disabled`** — `sweep.yml` reserves the review lease in its own
+   write-token step, so it must pass `--skip-start-comment`; that flag also
+   decided `coordinationEnabled`, which the probe consults before anything else.
+   The probe never ran, so `preHydrationStructuralRecord` stayed `null` and the
+   seeding branch could never persist a receipt for the next review either.
+2. **`activity_changed`** — the reservation the workflow posts is itself the
+   item's newest activity, landing after the sync markers the prior review
+   recorded, so ClawSweeper's own write read as reporter activity.
+3. **`target_changed`** — the decision compared the target repository's default
+   branch head across reviews. That head advances for reasons unrelated to the
+   item, so where it moves faster than the review cadence the key can never match.
 
 ## Level
 
-Module-level proof against the built `dist/` output, using production argument
-values. It is not a live end-to-end trace; see Limits.
+Module-level captures against the built `dist/` output using production argument
+values, plus end-to-end runs of the shipped `review` command against a live
+GitHub item. Not a trace against a production target; see Limits.
 
 ## Exercised surface
 
@@ -117,13 +125,12 @@ hydrated and re-reviewed in full.
   limit. The end-to-end section below closes part of this gap on a repository the
   proof author owns.
 - The end-to-end section below reaches a structural cache hit with no hydration
-  and no Codex call. Separately, this change does not by
-  itself produce cache hits on `openclaw/openclaw`, because `reviewStructuralCacheDecision`
-  separately compares the target repository's `main` head between reviews
-  (`src/review-structural-cache.ts:1280`), and that head advances roughly every
-  13 minutes there (100 commits in 21h11m, measured 2026-08-07). Whether an
-  unrelated `main` commit should invalidate a pull request's verdict is a design
-  question, filed separately.
+  and no Codex call, including after the target branch advanced. The runs use one
+  repository the proof author owns; behaviour on a production target is inferred
+  from the shipped code being identical, not observed. `openclaw/openclaw` is the
+  case that motivates this: its default branch advanced 100 times in the 21h11m
+  ending 2026-08-07, roughly once every 13 minutes, against a daily review
+  cadence.
 - No production review jobs were dispatched, no model capacity was consumed, and
   no contributor notifications were generated while validating this change.
 
@@ -225,6 +232,36 @@ review_structural_cache_hit: true
 `hydrations=0` is the point: the run neither hydrated GitHub context nor called
 Codex. It took 8 seconds end to end, against 70-150 seconds for the full reviews
 in runs 1-3.
+
+### An unrelated commit on the target branch no longer forces a full review
+
+The runs above kept the target repository's `main` still. Advancing it by one
+commit that has nothing to do with the item — a comment appended to `README.md`,
+where the item is about `src/slug.js` — was enough to lose the hit before this
+change:
+
+```text
+structural_cache_reasons: { "target_changed": 1 }
+cache_hits=0  hydrations=1                      (103 s, full hydration + Codex)
+```
+
+Same item, same reserved-lease flow, same advanced `main`, after the change:
+
+```text
+[review] shard=0/1 structural-cache-start-comment=reserved #1
+[review] shard=0/1 cache-hit structural-unchanged skip-hydration-model #1 (1/1)
+[review] shard=0/1 complete reviewed=1 cache_hits=1 structural_cache_checks=1
+  structural_cache_hits=1 ... content_cache_hits=0 hydrations=0
+```
+
+```json
+"structural_cache_reasons": { "hit": 1 }
+```
+
+8 seconds instead of 103, with no Codex call. Staleness is still bounded: the
+probe rejects any recorded verdict older than
+`REVIEW_STRUCTURAL_CACHE_MAX_AGE_DAYS`, so a full review still runs periodically
+regardless of what `main` did. No new constant is introduced.
 
 ### What this run did not establish
 
--- a/docs/proof/review-cache-coordination-gate/README.md
+++ b/docs/proof/review-cache-coordination-gate/README.md
@@ -110,13 +110,15 @@ hydrated and re-reviewed in full.
 
 ## Limits
 
-- No live claimed-delivery trace through the GitHub receipt-cache boundary: that
-  needs the organization's App installation token, which only repository
-  operators can exercise. #1036 disclosed the same limit. The production
-  observation above narrows the pre-fix reason by elimination from public state;
-  it does not read the probe's returned reason directly.
-- This restores the probe and the seeding path. It does not by itself produce
-  cache hits on `openclaw/openclaw`, because `reviewStructuralCacheDecision`
+- The pre-fix production observation narrows the reason by elimination from
+  public state; it does not read the probe's returned reason directly. Reading it
+  directly on a production target would need the organization's App installation
+  token, which only repository operators can exercise. #1036 disclosed the same
+  limit. The end-to-end section below closes part of this gap on a repository the
+  proof author owns.
+- The end-to-end section below reaches the reserved-lease claim and receipt
+  persistence, but not a structural cache hit. Separately, this change does not by
+  itself produce cache hits on `openclaw/openclaw`, because `reviewStructuralCacheDecision`
   separately compares the target repository's `main` head between reviews
   (`src/review-structural-cache.ts:1280`), and that head advances roughly every
   13 minutes there (100 commits in 21h11m, measured 2026-08-07). Whether an
@@ -124,3 +126,85 @@ hydrated and re-reviewed in full.
   question, filed separately.
 - No production review jobs were dispatched, no model capacity was consumed, and
   no contributor notifications were generated while validating this change.
+
+## End-to-end run against a live GitHub item
+
+The module-level captures above stop at the probe decision. This section drives
+the shipped `review` command itself, against a real issue, so the reserved-lease
+claim and the receipt-persistence path are exercised rather than inferred.
+
+### Setup
+
+- Target: <https://github.com/masatohoshino/clawsweeper-cache-proof> issue 1, a
+  repository the proof author owns. Nothing was run against a production target.
+- Runner: the branch at `4b1c8c8f`, built and executed from a copy of that tree.
+  The only difference from the branch is one added entry in
+  `config/target-repositories.json` `generic_fallbacks`, so the proof author's
+  own account is onboarded the same way `openclaw/*` and `steipete/*` already
+  are. **No source file differs**, and that entry is not part of this PR.
+- `CLAWSWEEPER_COMMENT_AUTHOR_LOGIN` names the proof author, which is the
+  supported way to extend `PATCHABLE_REVIEW_COMMENT_AUTHORS`
+  (`src/clawsweeper-review-comment-state.ts`).
+- Leases were created by `pnpm run reserve-review-lease`, the same command
+  `.github/workflows/sweep.yml` uses. No marker was hand-written.
+- Every review ran with `--review-source-action scheduled_normal_backfill`,
+  `--skip-start-comment`, and the reserved lease, matching the workflow.
+
+### Runs
+
+| Run | Prior state | Result |
+| --- | --- | --- |
+| 1 | no prior review | full hydration; report written with an unseeded receipt |
+| 2 | prior review present, receipt unseeded | full hydration; **receipt seeded** |
+| 3 | receipt seeded, no durable review comment | **reserved lease claimed**; revalidation stopped at `previous_review_changed` |
+| 4 | durable review comment published | model skipped by the content stage; structural stopped at `activity_changed` |
+
+### The reserved-lease claim runs
+
+Run 3, the line emitted only by the claim branch this PR adds:
+
+```text
+[review] shard=0/1 structural-cache-start-comment=reserved #1
+```
+
+Reaching that line requires the probe to pass coordination and the structural
+decision to hit, and the claim itself to elect the reserved lease as winner
+through the same freshness checks the hydrated path uses. On `main` this item
+would have stopped at `coordination_disabled` before any of it.
+
+### The receipt is persisted
+
+Run 1 and run 2 differ only in whether a prior completed review existed:
+
+```text
+run 1 report:  review_structural_target_head_sha: unknown
+run 2 report:  review_structural_target_head_sha: b575dc7ad47e96e62288ab26eb84d5c5b5b22adf
+```
+
+Run 2's metrics record the seeding path validating its own anchor:
+
+```json
+"structural_cache_revalidation_reasons": { "hydrated_anchor_match": 1, "verdict_input_match": 1 }
+```
+
+This is the production symptom — reports published with
+`review_structural_target_head_sha: unknown` — reproduced locally and then
+cleared.
+
+### What this run did not establish
+
+- **The structural cache still did not hit.** Run 4 missed with
+  `"structural_cache_reasons": { "activity_changed": 1 }`. The lease comment that
+  `reserve-review-lease` posts moves the item's `updated_at` past the owned-sync
+  markers recorded in the prior report (`review_comment_synced_at`,
+  `labels_synced_at`), so `activityCoveredByReview` cannot attribute it. That is
+  the next gate below the one this PR removes, not a failure of this change, and
+  whether it also fires in production is not established here.
+- **Run 4's cache hit is not attributable to this PR.** It was a content-stage
+  hit (`cache-hit content-unchanged skip-model`, `content_cache_hits=1`), and the
+  content stage takes its coordination from `Boolean(acquiredReviewLease)`, which
+  the pre-existing post-hydration claim already satisfies on `main`. It is
+  reported here because it happened, not as evidence for this change.
+- The durable review comment in run 4 was published with
+  `apply-decisions --sync-comments-only --skip-dashboard`, the documented local
+  apply repro, rather than by the production publication lane.
--- a/docs/proof/review-cache-coordination-gate/README.md
+++ b/docs/proof/review-cache-coordination-gate/README.md
@@ -144,7 +144,7 @@ claim and the receipt-persistence path are exercised rather than inferred.
 
 - Target: <https://github.com/masatohoshino/clawsweeper-cache-proof> issue 1, a
   repository the proof author owns. Nothing was run against a production target.
-- Runner: the branch at `4692e40a`, built and executed from a copy of that tree.
+- Runner: the branch at `51276d4a`, built and executed from a copy of that tree.
   The only difference from the branch is one added entry in
   `config/target-repositories.json` `generic_fallbacks`, so the proof author's
   own account is onboarded the same way `openclaw/*` and `steipete/*` already
--- a/docs/proof/review-cache-coordination-gate/README.md
+++ b/docs/proof/review-cache-coordination-gate/README.md
@@ -137,7 +137,7 @@ claim and the receipt-persistence path are exercised rather than inferred.
 
 - Target: <https://github.com/masatohoshino/clawsweeper-cache-proof> issue 1, a
   repository the proof author owns. Nothing was run against a production target.
-- Runner: the branch at `4b1c8c8f`, built and executed from a copy of that tree.
+- Runner: the branch at `4692e40a`, built and executed from a copy of that tree.
   The only difference from the branch is one added entry in
   `config/target-repositories.json` `generic_fallbacks`, so the proof author's
   own account is onboarded the same way `openclaw/*` and `steipete/*` already
--- a/docs/proof/review-cache-coordination-gate/README.md
+++ b/docs/proof/review-cache-coordination-gate/README.md
@@ -274,3 +274,69 @@ regardless of what `main` did. No new constant is introduced.
   apply repro, rather than by the production publication lane.
 - The runs use one repository the proof author owns. Behaviour on a production
   target is inferred from the shipped code being identical, not observed.
+
+## Controlled environment run (Crabbox `local-container`)
+
+The runs above executed on the author's host. This one repeats the focused
+validation inside a Docker-backed Crabbox lease at the current head, which is
+the controlled envelope `AGENTS.md` asks for on code-bearing changes.
+
+| Field | Value |
+| --- | --- |
+| Provider | `local-container` (Docker) |
+| Runtime | Docker 29.4.2, context `default` |
+| Image | `ubuntu:26.04` @ `sha256:678c6550cc43645e08669028bc177f50be4e7c5b8cca677067b1914d4afc7a03` |
+| Lease | `cbx_6e190a5c1faa` (slug `cache-gate-proof`, container `492644ca26eb`, type `ubuntu_26.04`) |
+| Head | `d8d7c31dbd248c569c0e32368b65e2a9beb58893` |
+| Node | v24.19.0, pnpm 11.10.0 |
+| Artifact | `cbx_6e190a5c1faa-artifacts.tgz`, 8882 bytes, sha256 `65f220b7afa0d4fd4980703b1d2e0f4ff3abf851d9b4b5c1225d6f550ae796e8` |
+| Captured | 2026-08-07T05:53:12Z |
+
+Commands, from a clean lease:
+
+```sh
+crabbox doctor  --provider local-container
+crabbox warmup  --provider local-container --slug cache-gate-proof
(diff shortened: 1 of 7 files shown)


steipete:
Maintainer triage note while this waits on proof: the bug is real and worth fixing — scheduled re-reviews of unchanged items burn Codex spend and API quota every pass. Two asks before this can land:

1. The review bot still marks the proof insufficient: it needs evidence from the scheduled lane itself (a scheduled delivery hitting the cache and completing as a cache-served no-op), not only the harness in docs/proof. The exact-review queue's own run artifacts are the cleanest source.

2. Scope question, honestly asked: this restructures review preparation (+837 lines) to make the cache reachable. Could the cache be made reachable on the scheduled lane with a smaller seam — e.g. threading the existing structural-cache lookup into the scheduled delivery path where it diverges — without moving the preparation logic? If you tried that and it genuinely doesn't work, say why in the PR body and the current shape is fine; VISION.md pushes us to prefer the smallest trustworthy change, so the reasoning deserves a sentence.

No urgency from our side — the fix direction is right. 🦞

steipete:
Superseded by #1067, now merged — thank you for this one. Your diagnosis was exactly right: scheduled deliveries never reached the structural cache, and the re-review churn was burning real model spend. What changed in the landed shape: instead of relocating review preparation, the cache path claims the scheduled delivery's supplied lease inside the existing flow and re-evaluates staleness against the reservation timestamp, which kept the diff at +463/−64 with your test approach preserved. The merged commit carries your Co-authored-by credit, and the maintainer note about preferring the smallest trustworthy seam is thanks to your PR making the tradeoff concrete. 🦞

masatohoshino:
'm honored to have been able to help. I learned a lot from seeing how you narrowed the fix down to the smallest trustworthy seam — really elegant. Thank you for the credit, too! 🦞

PR #1054, opened 2026-08-07
fix(router): fall back when public reads are throttled

giodl73-repo:
## Bounded escalation policy

The fallback budget is process-wide and keyed by the target App credential: at most one eligible throttled public read can escalate during a router run. Every later public throttle fails closed with `GitHubRateLimitError`. The regression test now covers a successful sync fallback, a repeated throttled read failing closed, a successful async fallback under a separate credential boundary, and a subsequent privileged mutation retaining the App-token path.


giodl73-repo:
@clawsweeper re-review

giodl73-repo:
@clawsweeper re-review

giodl73-repo:
Maintainer decision: accept the bounded availability tradeoff. One eligible public-read throttle may consume one App-backed read per target credential and process; later public throttles fail closed, and privileged mutations retain their configured App-token path.

PR #1056, opened 2026-08-07
fix: order canonical JSON keys by code unit so digests are reproducible

### LINKED ISSUE #1055: stableJson orders keys by locale collation, so the same input can hash two ways
## Summary

`stableJson` is the canonical-JSON helper behind ClawSweeper's content digests and
canonical equality checks, but it is not actually canonical. It orders object keys
with `String.prototype.localeCompare`, which is neither a strict total order nor
locale independent, so the same input can serialize — and therefore hash — two
different ways.

## Impact

Two independent defects, both in `src/stable-json.ts:6`:

**1. Not a strict total order.** `localeCompare` returns `0` for distinct strings
whose difference is collation-ignorable — zero-width joiner, soft hyphen, most
control characters. `Array.prototype.sort` is stable, so a tie leaves the keys in
property-insertion order. Two objects with an identical key/value set then
serialize differently:

```js
const ZWJ = "‍";
`a${ZWJ}b`.localeCompare("ab");                    // 0, but the keys are distinct
stableJson({ ab: 1, [`a${ZWJ}b`]: 2 });            // {"ab":1,"a‍b":2}
stableJson({ [`a${ZWJ}b`]: 2, ab: 1 });            // {"a‍b":2,"ab":1}   ← differs
```

**2. Locale and ICU dependent.** With no locale argument, `localeCompare` follows
the runtime default, and collation tables differ between locales and ICU releases.
This is not hypothetical — `cs-CZ` orders `ch` after `h`, which reorders two keys
that appear together in real digest payloads:

```
en-US : additions, baseSha, changedFiles, checksDigest, commitCount, ...
cs-CZ : additions, baseSha, commitCount, deletions, draft, changedFiles, checksDigest, ...
```

A digest computed on one runner would then not match the same input on another.

The practical consequence is digest instability, not an incorrect decision: a
digest that flaps causes spurious "source revision changed" verdicts, avoidable
cache misses, and extra Codex spend. No wrong close or merge follows from it.

## What is *not* affected

`src/review-structural-cache.ts` and `src/review-semantic-cache.ts` already import
`stableJsonCodeUnit as stableJson`, so the review caches were never exposed to this
and need no change. The correct comparator has been present in the module all
along — the unsuffixed export simply never used it.

The remaining locale-ordered consumers are `clawsweeper-source-revision.ts`,
`clawsweeper-item-context.ts`, `clawsweeper-runtime.ts`,
`clawsweeper-context-hydration.ts`, `clawsweeper-apply-proof-freshness.ts`,
`clawsweeper-apply-decision-workflow.ts`, `clawsweeper-assist.ts`, and
`dashboard/exact-review-queue.ts`.

## Suggested fix

Point `sortStable` at the existing `compareCodeUnits`, and keep
`stableJsonCodeUnit` / `sortStableCodeUnit` as aliases so no call site has to
change. Also switch the two `stableJson(...).localeCompare(stableJson(...))` array
sorts in `clawsweeper-context-hydration.ts` to the same comparator; they feed
`pullChecksDigest` and have the identical tie problem.

Measured against a pre-fix build, this changes **no persisted digest** across the
real call-site shapes — the affected key sets happen to order identically both
ways. It is a correctness fix with no migration.

## Related, but needs its own PR

`src/clawsweeper-source-revision.ts:73` sorts comments by
`` `${id}:${updated_at}`.localeCompare(...) ``. Same root cause, but fixing it
*does* change `item_source_revision` for essentially every item: collation orders
`:` before digits while code units order it after, so comment ids of differing
digit length reorder. That is migration-bearing and should be decided separately.

## Existing coverage gap

`test/stable-json.test.ts` has a single case, and it exercises
`stableJsonCodeUnit` — the variant that was already correct. Nothing tested the
unsuffixed `stableJson` that every digest actually used.

## Environment

| | |
|---|---|
| Node | v24.19.0 (Crabbox `local-container`), ICU 76.1 |

dwin-gharibi:
Closes #1055

## What Problem This Solves

Fixes an issue where ClawSweeper's canonical-JSON helper was not canonical: the
same input could serialize — and therefore hash — two different ways, producing
digest instability across runs and across runners.

`stableJson` backs the content digests and canonical equality checks used for
change detection and caching. It ordered object keys with
`String.prototype.localeCompare`, which fails as a canonical comparator twice over:

1. **Not a strict total order.** `localeCompare` returns `0` for distinct strings
   whose difference is collation-ignorable (zero-width joiner, soft hyphen, most
   control characters). `Array.prototype.sort` is stable, so a tie leaves keys in
   property-insertion order and two equal objects serialize differently.
2. **Locale and ICU dependent.** With no locale argument it follows the runtime
   default. This is not hypothetical: `cs-CZ` orders `ch` after `h`, reordering
   `changedFiles` and `checksDigest` — two keys that appear together in real
   digest payloads.

The practical consequence is digest instability, not an incorrect decision: a
flapping digest causes spurious "source revision changed" verdicts, avoidable
cache misses, and extra Codex spend. No wrong close or merge follows from it.

## Why This Change Was Made

`sortStable` now uses the `compareCodeUnits` comparator **that was already in the
module**. `stableJsonCodeUnit` / `sortStableCodeUnit` are kept as aliases, so no
call site changes.

The two `stableJson(...).localeCompare(stableJson(...))` array sorts in
`clawsweeper-context-hydration.ts` move to the same comparator; they feed
`pullChecksDigest` and had the identical tie problem.

Worth stating plainly: `review-structural-cache.ts` and `review-semantic-cache.ts`
already imported `stableJsonCodeUnit as stableJson`. The review caches were never
exposed to this defect and are untouched. The correct comparator has been sitting
in the module the whole time — the unsuffixed export just never used it.

Non-goals: this does not change what goes *into* any digest, the allowed value
types, or `JSON.stringify` semantics.

**Deliberately excluded.** `src/clawsweeper-source-revision.ts:73` sorts comments
by `` `${id}:${updated_at}`.localeCompare(...) ``. Same root cause, but fixing it
*does* change `item_source_revision` for essentially every item — collation orders
`:` before digits while code units order it after, so ids of differing digit length
reorder. That is migration-bearing and needs its own PR and its own decision about
the resulting re-review wave. Measured, not assumed:

```
locale  : [105, 12, 1234567890, 9, 987]
codeUnit: [105, 1234567890, 12, 987, 9]
```

## User Impact

Digests are reproducible for the same input, on any runner, in any locale. Fewer
spurious re-reviews and cache misses; no operator-visible behavior change and no
configuration change.

**No migration.** Measured against a pre-fix build, every persisted digest shape
across the affected call sites serializes byte-identically. Nothing stored is
invalidated — see *Real Behavior Proof*, claim 3.

**OpenClaw Bay:** not affected. No status, lifecycle, or telemetry shape Bay
renders changes; digests keep their existing values. No Bay update or Bay proof is
needed.

**Release note:** `fix: order canonical JSON keys by code unit so digests are reproducible`.
No `CHANGELOG.md` edit is included — happy to add one if this repo's release policy
wants it.

### Diff

```
 src/clawsweeper-context-hydration.ts |  9 ++++--
 src/stable-json.ts                   | 36 +++++++++++++++++++-----
 test/stable-json.test.ts             | 54 +++++++++++++++++++++++++++++++++++-
 3 files changed, 88 insertions(+), 11 deletions(-)
```

Plus a new proof package under `docs/proof/stable-json-canonical-ordering/`.

### Focused tests

`node --test test/stable-json.test.ts` → **6/6 pass** (1 pre-existing + 5 new).

The pre-existing case exercised `stableJsonCodeUnit` — the variant that was already
correct. Nothing tested the unsuffixed `stableJson` that every digest actually used.
New cases:

- insertion order cannot change the output (asserts the `localeCompare` tie exists
  first, so it cannot pass vacuously);
- key order is byte-defined, not locale-defined, including the `cs-CZ`-sensitive trio;
- nested arrays and objects order consistently;
- `stableJsonCodeUnit` remains an alias;
- `compareCodeUnits` is a strict total order — never ties distinct strings, and is
  antisymmetric.

### Red/green

Reverted only `src/stable-json.ts` and `src/clawsweeper-context-hydration.ts`,
rebuilt, and re-ran the same probe:

```
##### PRE-FIX #####
insertion-order independence: VIOLATED
   built one way : {"ab":1,"a‍b":2}
   built other   : {"a‍b":2,"ab":1}
byte-defined key order    : VIOLATED -> {"a":1,"B":2}

##### POST-FIX #####
insertion-order independence: HOLDS
byte-defined key order    : HOLDS
```

Focused suite: pre-fix the file fails to load (the new `compareCodeUnits` export
does not exist), so the probe above is the meaningful red signal rather than a test
count.

## Real Behavior Proof

**Claim.** `stableJson` produces a canonical form — the same key/value set always
serializes to the same bytes, on any runtime, in any locale — **and no stored
digest is invalidated by the change**.

**Exercised surface.** `dist/stable-json.js`, the module every content digest and
canonical equality check goes through, plus the two array sorts in
`clawsweeper-context-hydration.ts` that fed `pullChecksDigest`.

**Scenario / fixture.** `run-proof.mjs` asserts three claims against the built
`dist/`:

1. **Canonical** — for keys `localeCompare` reports as equal (zero-width joiner,
   soft hyphen), objects built in opposite insertion orders serialize identically.
   The tie is asserted first, so the check cannot pass vacuously.
2. **Byte order** — `{a, B}` serializes with `B` first (0x42 < 0x61), and the
   `cs-CZ`-sensitive trio keeps byte order.
3. **No churn** — seven persisted digest shapes, copied from the call sites that
   import the locale-ordered `stableJson`, serialize byte-identically to a
   **pre-fix build compiled inside the lease** from `git show <base>:src/stable-json.ts`.
   If that build is not supplied the proof reports SKIPPED and **fails**, rather
   than passing silently.

Docker-backed Crabbox `local-container` lease:

```bash
crabbox run \
  --provider local-container \
  --no-hydrate \
  --timing-json \
  --artifact-glob '.artifacts/stable-json-canonical-proof/**' \
  --script docs/proof/stable-json-canonical-ordering/run-proof.sh
```

| field | value |
|---|---|
| crabbox | `0.15.0` |
| lease | `cbx_260b82c60c03` (`brisk-lobster`) |
| run | `run_3a13e83c0068` |
| artifact | `.crabbox/runs/run_3a13e83c0068/run_3a13e83c0068-artifacts.tgz` |
| exit | `0` |

**Observed result.** In-container: **15/15 proof checks PASS**, focused suite
**6/6**, exit 0. The lease verified the comparator counts on both sides
(`pre-fix: 1 localeCompare`, `post-fix: 0`) before running the claims.

```
== Claim 3: persisted digest shapes are byte-identical to pre-fix ==
  PASS  itemContentDigest byte-identical
  PASS  itemSnapshotHash byte-identical
  PASS  reviewCommentContentRevision byte-identical
  PASS  pullCommitContentRevision byte-identical
  PASS  reviewTimelineDigestParts byte-identical
  PASS  pullChecksDigestParts byte-identical
  PASS  action-ledger invocation id byte-identical
```

**Known churn, and why it is safe.** One shape *does* change: the activity receipt
(`stableJson(completeActivityContext)`) flips `reviews` and `reviewThreads`, because
collation orders `s` before `T` while code units order `T` (0x54) before `s` (0x73).

That value is never persisted. It is produced at
`clawsweeper-apply-decision-workflow.ts:950` and compared at
`clawsweeper-apply-proof-freshness.ts:132` — both inside one apply run, in one
process, from one build. Both sides shift identically, so the comparison still
matches. I grepped for any persistence of it (front matter, schema, config) and
found none. It is deliberately excluded from the claim-3 fixture set, which covers
only digests that outlive a run.

**Artifact / trace.** `docs/proof/stable-json-canonical-ordering/` holds
`run-proof.sh` (Crabbox entry point), `run-proof.mjs`, and the contract README. The
lease artifact tarball holds `proof-output.txt`, `focused-tests.txt`, `install.log`,
and `build.log`.

**Limits.** Covers key ordering only. Claim 3 is evidence over seven representative
shapes taken from the call sites, **not** an exhaustive enumeration of every object
ever passed to `stableJson` — a shape whose sibling keys differ only by case or by a
`ch`-style collation boundary could still churn; none was found among the real call
sites. `dashboard/exact-review-queue.ts` uses `stableJson` for in-process equality
comparisons, which are order-consistent within a run and were not separately
measured. No live GitHub, Worker, or queue is contacted, and the proof performs no
mutation.

The proof and focused suite run on Node 24 inside the Crabbox lease recorded above
(`6/6` in-container). The note below concerns only the *full* suite on the macOS host.

Green on this branch:

|---|---|---|
| Crabbox `local-container` proof | Node 24 container | **PASS**, 15/15, exit 0 |
| `test/stable-json.test.ts` | Node 24 container | **6/6** |

### DIFF
--- a/docs/proof/stable-json-canonical-ordering/README.md
+++ b/docs/proof/stable-json-canonical-ordering/README.md
@@ -1,0 +1,122 @@
+# stableJson canonical-ordering proof contract
+
+## Claim
+
+`stableJson` produces a canonical form: the same key/value set always serializes
+to the same bytes, on any runtime, in any locale. Before this change it ordered
+keys with `String.prototype.localeCompare`, which is neither a strict total order
+nor locale independent — so the same input could hash two different ways.
+
+The fix does **not** invalidate any stored digest. Every persisted digest shape
+produced by the affected call sites serializes byte-identically before and after.
+
+## Exercised surface
+
+`dist/stable-json.js` — the module every content digest and canonical equality
+check goes through — plus the two array sorts in
+`src/clawsweeper-context-hydration.ts` that fed `pullChecksDigest` through the
+same comparator.
+
+Note that `src/review-structural-cache.ts` and `src/review-semantic-cache.ts`
+already imported `stableJsonCodeUnit as stableJson`, so the review caches were
+never exposed to this defect and are untouched by the fix.
+
+## Controlled scenario and fixture
+
+`run-proof.mjs` asserts three claims against the built `dist/`:
+
+1. **Canonical** — for keys that `localeCompare` reports as equal (zero-width
+   joiner, soft hyphen), two objects built in opposite insertion orders serialize
+   identically. The proof first asserts the tie exists, so it cannot pass
+   vacuously on a runtime where those characters are not ignorable.
+2. **Byte order** — `{a, B}` serializes with `B` first (0x42 < 0x61), and the
+   `cs-CZ`-sensitive trio `changedFiles` / `checksDigest` / `commitCount` keeps
+   byte order. Czech collates `ch` after `h`, which reorders exactly those keys.
+3. **No churn** — seven persisted digest shapes, copied from the call sites that
+   import the locale-ordered `stableJson`, serialize byte-identically to a
+   **pre-fix build compiled from the base commit**. The pre-fix module is compiled
+   inside the lease from `git show <base>:src/stable-json.ts`; if it is not
+   supplied the proof reports SKIPPED and fails rather than passing silently.
+
+## Expected observation
+
+15 checks, all PASS, exit 0. Pre-fix, claims 1 and 2 both fail:
+
+```
+insertion-order independence: VIOLATED
+   built one way : {"ab":1,"a‍b":2}
+   built other   : {"a‍b":2,"ab":1}
+byte-defined key order    : VIOLATED -> {"a":1,"B":2}
+```
+
+## Artifact and command
+
+Supported-environment run (Node 24, Crabbox `local-container`):
+
+```bash
+crabbox run \
+  --provider local-container \
+  --local-container-image node:24 \
+  --no-hydrate \
+  --timing-json \
+  --artifact-glob '.artifacts/stable-json-canonical-proof/**' \
+  --script docs/proof/stable-json-canonical-ordering/run-proof.sh
+```
+
+Host-only quick check (supply a pre-fix build for claim 3):
+
+```bash
+pnpm run build
+node docs/proof/stable-json-canonical-ordering/run-proof.mjs /path/to/pre-fix/stable-json.js
+```
+
+Focused tests:
+
+```bash
+node --test test/stable-json.test.ts
+```
+
+## Provenance
+
+- provider: Crabbox `local-container` (Docker/OrbStack)
+- crabbox: `0.15.0`
+- image: `node:24` @ `sha256:934240a162082fd8b8a2f90cd5114446443f1eba1c5378f6687167ca405e6584`
+- container node: `v24.19.0` (satisfies `engines.node >= 24`)
+- lease: `cbx_260b82c60c03` (`brisk-lobster`)
+- run: `run_3a13e83c0068`
+- artifact: `.crabbox/runs/run_3a13e83c0068/run_3a13e83c0068-artifacts.tgz`
+  (`proof-output.txt`, `focused-tests.txt`, `install.log`, `build.log`)
+- result: exit `0`; 15/15 proof checks PASS; focused suite `6/6`
+- privacy: synthetic fixtures only. The proof makes no network call, contacts no
+  GitHub API, and performs no queue, GitHub, or production mutation.
+
+## Known churn, and why it is safe
+
+One shape does change: the **activity receipt**
+(`stableJson(completeActivityContext)`) flips `reviews` and `reviewThreads`,
+because `localeCompare` orders `s` before `T` while code units order `T` (0x54)
+before `s` (0x73).
+
+That value is never persisted. It is produced at
+`src/clawsweeper-apply-decision-workflow.ts:950` and compared at
+`src/clawsweeper-apply-proof-freshness.ts:132` — both inside a single apply run,
+in one process, from one build. Both sides shift identically, so the comparison
+still matches. It is deliberately excluded from the claim-3 fixture set, which
+covers only digests that outlive a run.
+
+## Limits
+
+Covers key ordering only. It does not change what goes *into* any digest, the
+allowed value types, or `JSON.stringify` semantics.
+
+Claim 3 is evidence over seven representative shapes taken from the call sites,
+not an exhaustive enumeration of every object ever passed to `stableJson`. A
+shape whose sibling keys differ only by case or by a `ch`-style collation
+boundary could still churn; none was found among the real call sites.
+
+**Deliberately out of scope:** `src/clawsweeper-source-revision.ts:73` sorts
+comments by `` `${id}:${updated_at}`.localeCompare(...) ``. That has the same root
+cause, but fixing it *does* change `item_source_revision` for essentially every
+item — collation orders `:` before digits while code units order it after, so ids
+of differing digit length reorder. That is a migration-bearing change and needs
+its own PR and its own re-review-wave decision.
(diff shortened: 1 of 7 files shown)


dwin-gharibi:
Done. @clawsweeper re-review.

steipete:
Verified before merge: ran docs/proof/stable-json-canonical-ordering/run-proof.sh locally on the PR head — environment checks, pre-fix merge-base comparison build, proof assertions, and the focused suite all passed (6/6). CI green on the head SHA. Thanks @dwin-gharibi — surgical fix, exemplary proof discipline. 🦞

PR #1118, opened 2026-08-11
fix(review): hydrate PR blobs with older Git clients

steipete:
## What Problem This Solves

Restricted PR review hydration needed an old-Git-compatible way to classify missing promisor blobs. The first repair used `rev-list --objects --missing=print --no-walk=unsorted`, but `--no-walk` prevented the required tree traversal on Git 2.39.5, so the parser could not observe the selected blob IDs and hydration failed closed before its bounded fetch.

## Root Cause and Masking Path

The old integration assertion counted any Git fetch. On Git 2.39.5, the path-limited `rev-list` path could spawn an implicit promisor fetch while resolving paths; that fallback fetch satisfied the test's count even though the production code's explicit bounded `--stdin` fetch did not run. The proof therefore looked green without proving the intended path.

## Fix

The availability probe now starts from the two exact tree objects:

```text
git rev-list --objects --missing=print <base>^{tree} <head>^{tree} -- <bounded-paths>
```

This recurses through the selected base/head trees, emits their blob IDs, and cannot walk either commit history. It also avoids the incorrect `--not --all` shape, so other refs cannot suppress objects that belong to the reviewed trees.

The strengthened test now:

- asserts that probe output includes every expected blob ID;
- requires both `^{tree}` roots and rejects `--no-walk`;
- rejects a fetch nested under the probe; and
- requires exactly one later explicit fetch using `--stdin`.

Path safety, the 4 MiB review limit, GraphQL size resolution, one-request fetching, and fail-closed behavior are unchanged.

## Real Behavior Proof

- Claim: Git 2.39.5 and current Git traverse only the two reviewed trees, observe every selected blob, perform no implicit probe fetch, and hydrate missing blobs with one explicit bounded fetch.
- Surface: `hydratePullRequestReviewBlobs` plus a genuine filter-capable bare origin and real `blob:none` partial clone.
- Scenario: base/head trees contain four selected blobs; two head blobs begin missing; an earlier history-only blob and a separately fetched unrelated-ref blob must remain unobserved.
- Result on Git 2.39.5 and Git 2.55.0: `expected.size=4`, `observed.size=4`, `observed.complete=true`, `missing.size=2`, `probe_nested_fetches=0`, `bounded_fetches=1`, `bounded_fetch_uses_stdin=true`, `hydrated=true`, and offline objects complete.
- Committed traces: [`docs/proof/blob-hydration-env/artifacts/git-2.39.5-tree-traversal.log`](https://github.com/openclaw/clawsweeper/blob/2439d7190c74/docs/proof/blob-hydration-env/artifacts/git-2.39.5-tree-traversal.log) and [`current-git-tree-traversal.log`](https://github.com/openclaw/clawsweeper/blob/2439d7190c74/docs/proof/blob-hydration-env/artifacts/current-git-tree-traversal.log).
- Git 2.39.5 trace provenance: Crabbox provider `aws`, lease `cbx_7e3080158c5c` (`coral-lobster-9594`), run `run_0a87aaf605e6`, `node:24-bookworm`, exit 0, lease stopped.
- Limits: the fixture uses real Git transport and partial-clone behavior, while the bounded size callback reads from its local source repository rather than GitHub GraphQL.

## Exact Pushed-Head Container Gates

Crabbox checked out the exact pushed PR head with `--fresh-pr openclaw/clawsweeper#1118` and no local patch. Provider `local-container`, lease `cbx_36c3ecd5d6d6` (`swift-shrimp-1a40`), image `node:24-bookworm`, exit 0, `leaseStopped=true`.

```text
head=2439d7190c74dc39a76b0947f28db3bb4dc621d8
branch=crabbox-pr-1118
worktree_changes=1
node=v24.19.0
git=git version 2.39.5
pnpm=11.10.0
jq=jq-1.8.1
jq_checksum=passed
build-all=passed
tests=3318
pass=3310
fail=0
cancelled=0
skipped=8
test-no-build=passed
lint=passed
format-check=passed
check-active-surface=passed
```

`worktree_changes=1` is the proof harness uploaded under `.crabbox/scripts/` after the clean fresh-PR checkout; no patch was applied to the pushed head.

Pre-commit Codex autoreview and the committed branch review against current `origin/main` both completed cleanly with no accepted/actionable findings.

## OpenClaw Bay Impact

None. This changes only local review-context Git object availability detection. It does not change lifecycle, publication, queue, status, telemetry, dashboard contracts, or Bay's observer-only boundary.

### DIFF
--- a/docs/proof/blob-hydration-env/README.md
+++ b/docs/proof/blob-hydration-env/README.md
@@ -14,14 +14,28 @@ isolated genuine blobless clone showed why: `GIT_NO_LAZY_FETCH=1 git cat-file --
 exited 128 with zero output rather than emitting a `missing` row for the unavailable promised blob.
 
 Git's [2.45.0 release notes](https://github.com/git/git/blob/master/Documentation/RelNotes/2.45.0.adoc)
-place the correct client-side `--no-lazy-fetch` behavior in that release. The replacement probe,
-`git rev-list --objects --missing=print --no-walk=unsorted`, was exercised against the same
-Git 2.39.5 partial clone and reported the unavailable object as `?<oid>` without fetching it.
+place the correct client-side `--no-lazy-fetch` behavior in that release. The first replacement
+probe incorrectly retained `--no-walk=unsorted`. On Git 2.39.5, its path-limiting fallback spawned
+an implicit promisor fetch; the integration test counted that fetch without proving it was the
+production path's explicit bounded fetch. That masked the missing traversal and let the old proof
+appear green.
+
+The repaired probe starts from the exact tree objects instead:
+
+```text
+git rev-list --objects --missing=print <base>^{tree} <head>^{tree} -- <bounded-paths>
+```
+
+Tree recursion emits the selected blobs while commit-history traversal is impossible. A fixture
+with an earlier history-only blob and a separately fetched unrelated ref proved that neither was
+observed. The same real blobless-clone scenario passed on Git 2.39.5 and Git 2.55.0: all four
+expected blob IDs were observed, two were reported missing, no fetch ran under the probe, and the
+later explicit `--stdin` fetch ran exactly once.
 
 ## Chosen option
 
 This is the production-path fix from option 1 of the work order, not a test skip. Hydration now
-uses `rev-list` to classify the object IDs reached from the exact base/head commits and bounded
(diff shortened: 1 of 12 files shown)


steipete:
@clawsweeper re-review

steipete:
@clawsweeper re-review

PR #1121, opened 2026-08-11
perf(queue): reuse alarm state snapshots across synchronous sections

steipete:
## Summary

Reuse already-hydrated `ExactReviewQueue` alarm state across two synchronous sections instead of scanning the normalized item table again.

## Why each reuse is safe

| Reused snapshot | Why it cannot be stale |
| --- | --- |
| Alarm housekeeping transaction result | Between the legacy compatibility mirror and the former next-line read, the callback only returns and the synchronous transaction closes; neither path writes the normalized item table or dispatcher. |
| Post-terminal-preflight snapshot used for batch reservation | Between the recheck and reservation, the code performs only synchronous capacity, admission, candidate, and backoff calculations; there is no `await` or normalized-state mutation. |

Every later alarm read remains because an external await or a possible queue-state write intervenes.

## Behavior proof

Claim: the real dashboard Worker and `ExactReviewQueue` Durable Object preserve the public queue state for the same populated batch-publication alarm while eliminating two redundant full item-table hydrations.

- Current head: `44b847da52`
- Exercised surface: independently built merge-base and candidate trees, real local Wrangler Workers, SQLite-backed Durable Objects, eight signed publication enqueues, automatic alarm execution, and a loopback GitHub App/API stub accepting the real batch workflow dispatch.
- Observable result: full normalized item-table scans per alarm decreased from **7 to 5**.
- State result: normalized public queue snapshots were byte-identical before and after the alarm at merge-base and head.
- Focused regression: `exact-review batch alarm avoids redundant full queue hydrations` passed at both revisions with the expected 7/5 assertions.
- Full local suite: 3,313 passed, 0 failed, 9 skipped, serially on Node 24.
- Static gates: lint, format check, dashboard queue boundary, active surface, and limits all passed.
- Reproduction and retained evidence: `docs/proof/alarm-hydration-dedup/README.md` and `docs/proof/alarm-hydration-dedup/run-proof.sh`.

### Exact pushed-head container proof

Crabbox checked out PR 1121 with `--fresh-pr` at exact pushed head
`44b847da529590d9f4e895045c312d1aaf326bbb`, with no local patch. The final run used
`provider=aws`, lease `cbx_1e632a851089` (`harbor-shrimp-945d`), run
`run_ce35bb320c92`, Docker 29.7.1, and `node:24-bookworm`. Corepack supplied pnpm 11.10.0;
jq 1.8.1 was installed only after `jq-linux-amd64: OK` against the official checksum manifest.

The container result was fully green: 3,322 tests, 3,314 passed, 0 failed, 0 cancelled, and 8
platform skips. Build, lint, format check, dashboard queue boundary, active surface, and limits all
passed. The container then ran the real behavior proof: scans decreased 7→5, normalized before and
after states were byte-identical, and Bay impact remained `none`.

Limits: the GitHub service is a loopback behavioral stub; no production repository, Worker, secret, comment, workflow, or queue is contacted. SQL scan counts use the existing test-only storage adapter because Wrangler does not expose per-query counts without production instrumentation. This proves state equivalence and query count, not CPU time.

Proof: sufficient.

## OpenClaw Bay

None. This changes internal alarm hydration reuse only; it does not alter queue/lifecycle data, the public observer schema, or any control boundary.

### DIFF
--- a/dashboard/exact-review-queue.ts
+++ b/dashboard/exact-review-queue.ts
@@ -3323,12 +3323,13 @@ export class ExactReviewQueue {
     await this.storage.deleteAlarm();
     await this.processBranchAuthorityReservations(startedAt);
     await this.processSourceAuthorityReservations(startedAt);
-    this.storage.transactionSync(() => {
+    let snapshot = this.storage.transactionSync(() => {
       this.pruneDeliveryReceiptsSync(startedAt);
       this.directPublicationStore.pruneTerminalSync(startedAt);
-      this.syncLegacyCompatibilitySync(this.readStateSync());
+      const current = this.readStateSync();
+      this.syncLegacyCompatibilitySync(current);
+      return current;
     });
-    let snapshot = this.readStateSync();
     let snapshotBatchOwnership = this.batchStore.activeLeaseSnapshot(startedAt);
     const reclaimedSnapshot = reclaimExpiredExactReviewLeases(
       snapshot,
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper re-review

PR #1129, opened 2026-08-11
fix: prevent apply cursor livelock

steipete:
## Summary

Fix the ongoing `openclaw/openclaw` apply/comment-sync cursor livelock without raising its 300-second runtime budget.

Production run https://github.com/openclaw/clawsweeper/actions/runs/31544381133 selected 39 urgency-ranked records before the only numeric frontier record. It terminally evaluated 25 records, yielded at the runtime budget, and printed `Comment sync made no cursor progress`; the durable cursor therefore stayed at `#105854` and every later cycle selected the same window.

The selector still reserves its first slot for the numeric frontier, which helps non-apply consumers. The load-bearing guarantee now lives in the TypeScript apply loop: with an active comment-sync cursor, it computes the smallest selected item number strictly greater than the cursor and executes that record first regardless of incoming `item_numbers` position. If no selected number is greater, it wraps to the smallest selected number, matching the existing cycle bookkeeping. Cursor completion follows the trusted terminal execution trace rather than the shell list order, so downstream `sort` operations cannot recreate the starvation.

The branch also removes the redundant placeholder sweep attempted before apply-lease acquisition. Those deletes were correctly rejected locally with `apply mutation lease is not held`; the lease reaper then removed the expired placeholders through its authorized path. Post-publication placeholder cleanup remains intact.

## Behavior proof

- Claim: the apply loop executes a terminal numeric frontier first and advances the durable comment-sync cursor even when later urgent work is budget-clipped; clipping the frontier itself does not advance.
- Exercised surface: the real `apply-decisions` loop, `comment-sync-batch` selector, `complete_comment_sync_batch`, cursor writer, canonical closed-record fallback, and wrap/cycle regression seams.
- Scenario: the exact 40 item numbers and cursor/frontier from run `31544381133`, plus adversarial ascending input `[87267,95788,97566,105342,105870]` with frontier `#105870` last.
- Command/environment: `docs/proof/apply-cursor-livelock/run-proof.sh` on Node 24.19.0, pnpm 11.10.0, jq 1.8.2; synthetic filesystem records, no GitHub credentials or mutation.
- Observable result: the adversarial apply report and trace both recorded `#105870` first and persisted cursor `#105870` with every lower urgent record unexamined; the clipped-frontier replay persisted `#105854` with `cursor_count=0`; five focused apply/selector/completion/wrap probes passed.
- Artifact: `docs/proof/apply-cursor-livelock/fixture-result.json`, behavior contract/report, container provenance, and the scripted reproduction in the same directory.
- Limits: the fixture proves apply-loop ordering and cursor-completion semantics, not production GitHub latency. It does not claim urgent records no longer require later repair.

## Finding disposition

- Coordinator P1: accepted in substance. Shell normalization can reorder `item_numbers`; the loop-level numeric comparison is now the authoritative guarantee.
- Pre-commit P1 on `find()` ordering: accepted. The loop now uses an explicit numeric minimum, with forward, wrap, ascending, and unsorted coverage.
- Pre-commit P1 on clipped-frontier trace reconstruction: rejected after source verification. Apply removes an interrupted current item before writing the trusted trace; an empty terminal prefix returns with zero progress before wrap-cycle state can be persisted. The invariant is documented inline and covered by the runtime-yield continuation test.

## Validation

- `pnpm run check`: 3,336 tests; 3,327 passed, 9 skipped, 0 failed; coverage thresholds passed.
- Focused suite: 243 passed, 0 failed.
- Scripted behavior proof: 5 passed, 0 failed; `PROOF_RC=0`.
- Pre-commit and committed-branch Codex autoreviews: clean, no accepted/actionable P0/P1 findings.
- Crabbox container proof: passed on committed code/proof head `2ebe882831f99d9f22dd6f696f48d2a4e80f3195` through configured `provider=aws`, lease `cbx_a9f86ad7a357` (`quick-prawn-5920`), run `run_61dcc42eb8d2`, inside `node:24-bookworm`; 5/5 focused tests passed, exit 0, and lease cleanup is confirmed. Receipt commit `695d7beaca437fe03d1da003c4a69838c043602a` changes documentation only and was included in the same push.
- jq 1.8.2 trust root: the downloaded `jq-linux-amd64` binary matched repository-pinned SHA-256 `b1c22172dd303f3be49e935aa56aa48a8b7a46e0bc838b4997d3bb451495870f` before installation.

## Scope

No runtime-budget increase, cursor schema change, close-policy change, `CHANGELOG.md`, or `CLAUDE.md` change.

OpenClaw Bay is unaffected: this changes internal apply-lane ordering and removes a redundant pre-lease cleanup attempt; no lifecycle publication or observer-only dashboard contract changes.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -5952,6 +5952,7 @@ jobs:
           if [ "$sync_comments_only" = "true" ]; then
             sync_comments_arg=(--sync-comments-only)
           fi
+          prepare_comment_sync_cursor_arg
           min_age_minutes_arg=()
           if [ -n "$min_age_minutes" ]; then
             min_age_minutes_arg=(--min-age-minutes "$min_age_minutes")
@@ -6052,6 +6053,7 @@ jobs:
               --processed-limit "$comment_sync_processed_limit" \
               --max-runtime-ms 300000 \
               --cursor-trace ".artifacts/comment-sync-trace-$checkpoint.json" \
+              "${comment_sync_cursor_arg[@]}" \
               --comment-sync-min-age-days "$comment_sync_min_age_days" \
               "${item_numbers_arg[@]}" \
               "${sync_comments_arg[@]}"
--- a/docs/proof/apply-cursor-livelock/README.md
+++ b/docs/proof/apply-cursor-livelock/README.md
@@ -1,0 +1,60 @@
+# Apply cursor livelock proof
+
+## Claim
+
+The automatic `openclaw/openclaw` comment-sync batch always executes its numeric cursor frontier
+before opportunistic urgent repairs. Per-pass-terminal outcomes can therefore advance the durable
+cursor, while a runtime yield on the frontier itself leaves the cursor unchanged for a truthful
+retry.
+
+This fixes the livelock observed in production run
+https://github.com/openclaw/clawsweeper/actions/runs/31544381133. That run selected 39 urgency-ranked
+records before frontier `#105870`, exhausted its 300-second budget at urgent record `#85937`, and
+never reached the only record that could advance cursor `#105854`.
+
+## Controlled scenario and fixture
+
+`run-proof.mjs` uses the exact 40-number selection from the production run and a synthetic action
+mix matching its terminal classes: durable comment sync, changed-since-review, stale-sync skip,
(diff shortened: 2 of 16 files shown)


steipete:
@clawsweeper re-review

PR #1138, opened 2026-08-12
fix(records): write front matter keys and values literally

### LINKED ISSUE #1135: A GitHub label containing $& corrupts the record it is written to
## Summary

`replaceFrontMatterValue` in `src/clawsweeper-record-metadata.ts` passes the value
it is storing to `String.prototype.replace` as a *replacement string*. In that
position `$&`, `` $` ``, `$'` and `$$` are expansion patterns, not literal text. The
label apply lane stores the synced label set through this function, and those names
come from the reviewed repository — so a repository with a label containing `$` gets
a record that does not say what the sync decided.

## Impact

`src/clawsweeper-apply-report-labels.ts` writes the label set at three sites:

```ts
markdown = replaceFrontMatterValue(markdown, "labels", JSON.stringify(item.labels));
```

Two outcomes, both observed against the shipped module:

| Label name | Stored value | Result |
| --- | --- | --- |
| `bug` | `["bug"]` | correct |
| `$&` | `["labels: []"]` | valid JSON, **wrong label set** — echoes the previous line |
| `` $` `` | `["---` | invalid JSON |
| `$'` | `["` | invalid JSON, and following record text is pulled into the field |
| `a$&b` | `["alabels: []b"]` | valid JSON, wrong label set |

`frontMatterStringArray` falls back to comma-splitting when the value does not
parse, so invalid JSON becomes a plausible-looking but wrong list rather than an
error. That list is what these read:

- `PROOF_OVERRIDE_LABEL` in `src/clawsweeper-orchestration-foundation.ts:436` and
  `src/clawsweeper-report-parser.ts:384`
- `AUTOFIX_LABEL` in `src/clawsweeper-review-comment-automation.ts:156`
- the priority and rating checks in `src/clawsweeper-label-presentation.ts`
- the dashboard audit's label view in `src/clawsweeper-dashboard-audit.ts:244`

The `$'` case is the worst: because the expansion is the text *following* the match,
the write reaches past the field and the record's own `title`, `number` and `---`
delimiters no longer round-trip.

## Reproduction

```js
const record = ["---", "repository: openclaw/openclaw", "labels: []", "---", "", "body", ""].join("\n");
const written = JSON.stringify(["$&"]);
const next = replaceFrontMatterValue(record, "labels", written);
frontMatterValue(next, "labels");
// -> '["labels: []"]'   (expected '["$&"]')
```

## Cause

```ts
function replaceFrontMatterValue(markdown: string, key: string, value: string): string {
  const line = `${key}: ${value}`;
  const pattern = new RegExp(`^${key}:\\s*.*$`, "m");
  if (pattern.test(markdown)) return markdown.replace(pattern, line);
  return markdown.replace(/^---\n/, `---\n${line}\n`);
}
```

`line` carries the caller's value into the replacement-string position of both
`replace` calls.

The same function also interpolates `key` straight into a `RegExp` source, as do the
two regexes in `frontMatterField`. A key containing `(`, `[`, `|`, `+` or `*` throws
`SyntaxError: Invalid regular expression` or matches the wrong line — `a.c` matches
a line reading `aXc:`. No shipped caller passes such a key today, so this half is
latent, but the helpers advertise a plain `key: string` contract.

## Proposed solution

1. Pass a replacement **function** to both `replace` calls so the value is inserted
   literally. This is the reachable fix.
2. Escape `key` with the existing `escapeRegExp` from `src/clawsweeper-text.ts`
   before interpolating it, in `replaceFrontMatterValue` and in both
   `frontMatterField` regexes.

`src/clawsweeper-text.ts` has no imports of its own, so there is no cycle.

dwin-gharibi:
Closes #1135.

---

## What Problem This Solves

Fixes an issue where syncing labels for an item in a reviewed repository writes a
corrupted `labels:` field — and in some cases damages the rest of the record — when
any label name contains `$&`, `` $` ``, `$'` or `$$`.

The label apply lane stores the synced label set with

```ts
markdown = replaceFrontMatterValue(markdown, "labels", JSON.stringify(item.labels));
```

That value went into `String.prototype.replace` as a *replacement string*, where
`$&`, `` $` `` and `$'` are expansion patterns rather than literal text. The label
names come from the reviewed repository, so any repository whose labels contain
those characters produces a record that no longer says what the sync decided.

Two distinct outcomes, both observed:

- `` $` `` and `$'` expand to the text before / after the match, which yields invalid
  JSON *and* pulls surrounding record content into the field.
- `$&` expands to the matched line, so the field silently reads back the *previous*
  label state instead of the new one.

Downstream, `frontMatterStringArray` falls back to comma-splitting when the value
does not parse, so the corrupted text becomes a plausible-looking but wrong label
list. That list is what `PROOF_OVERRIDE_LABEL`, `AUTOFIX_LABEL` and the priority and
rating checks read.

## Why This Change Was Made

`replaceFrontMatterValue` never intended `$` expansion — it is a writer storing a
value verbatim — so the replacement argument is now a function, which inserts the
text literally. The same function interpolated the caller's `key` straight into a
`RegExp` source; that is now escaped, so keys are matched literally as the
`key: string` signature advertises. Both readers in the file (`frontMatterField`)
were interpolating the key the same way and are escaped alongside it.

Non-goals: the front matter format is unchanged, no key is renamed, and the label
sync logic itself is untouched — only how the chosen value is written down.

## User Impact

Repositories whose labels contain `$` characters now get records that match what the
sync actually decided. Operators reading a record's `labels:` field, and the
override and priority checks that read it back, see the real label set instead of a
mangled one.

For the overwhelming majority of repositories — whose label names contain no `$` —
nothing changes; the ordinary case is byte-for-byte identical before and after.

## OpenClaw Bay Impact

Bay reads `labels` from published records. This change makes that field correct for
the affected repositories rather than changing its shape, so no Bay change is
needed. No new field, no removed field, no status or telemetry contract change.

## Documentation Lifecycle

No documentation lifecycle changes. The new file under
`docs/proof/record-front-matter-literal-write/` is PR evidence, not a runbook or
reference.

### Change

- `src/clawsweeper-record-metadata.ts` — import `escapeRegExp`; escape `key` in the
  two `frontMatterField` regexes and in `replaceFrontMatterValue`; pass the
  replacement as a function in both `replace` calls.
- `test/clawsweeper-record-metadata.test.ts` — two regression tests.

`17` changed lines in source, `51` added in tests.

### Focused tests

```
$ node --test test/clawsweeper-record-metadata.test.ts
✔ front matter fields are ambiguous when the same key occurs after the leading block
✔ front matter fields preserve current-format, duplicate, and no-block behavior
✔ label names containing replacement patterns survive a front matter write
✔ front matter keys are matched literally, not as regular expressions
ℹ pass 4   ℹ fail 0
```

Reverting only `src/clawsweeper-record-metadata.ts` and rebuilding turns both new
tests red (`pass 2 / fail 2`) while the two pre-existing tests stay green.

### Full suite and static gates

`pnpm run check:static`, `pnpm run lint`, `pnpm run format:check` and
`pnpm run build:all` all pass. The full suite is `2050 / 2067` with 6 failures, none
of which is caused by this change:

| Failure | Verdict |
| --- | --- |
| `action-ledger-runtime.test.ts` temp-dir cleanup | fails identically on unmodified `main` |
| `automerge-metrics.test.ts` CLI JSON | fails identically on unmodified `main` |
| `apply-runtime-budget.test.ts` ×2 | passes `10/0` in isolation — contention flake |
| `codex-process.test.ts` | `9 pass / 1 fail` on unmodified `main` too |
| `codex-review-runner.test.ts` | `17 pass / 2 fail` on `main`; this branch is `18 / 1` |
| `dashboard-worker.test.ts` ×9 | passes `328/0` in isolation — cascade after one timeout |

None of those files import `clawsweeper-record-metadata`.

## Real Behavior Proof

**Claim.** A label whose name contains `$&`, `` $` ``, `$'` or `$$` is stored in
`labels:` exactly as given, the record around it is unchanged, and front matter keys
are matched literally rather than compiled as regular expressions.

**Exercised surface.** `createRecordMetadata(...).replaceFrontMatterValue` and the
matching readers `.frontMatterField` / `.frontMatterValue` /
`.frontMatterStringArray` in `dist/clawsweeper-record-metadata.js` — the writer
`src/clawsweeper-apply-report-labels.ts` calls at three sites.

**Scenario / fixture.**
`docs/proof/record-front-matter-literal-write/run-proof.mjs` writes label sets
through the shipped writer and reads them back through the shipped readers:
verbatim storage (including a mixed set `["needs-triage", "$&"]` that mirrors a real
sync result), `JSON.parse` round-trip, record integrity (`title`, `number`, body and
`---` delimiters unchanged), and literal key matching for keys containing
`. + ( [ | *`.

**Command and environment.**

```
bash docs/proof/record-front-matter-literal-write/stage-before.sh
crabbox run --provider local-container --local-container-image node:24 --no-hydrate \
  --artifact-glob '.artifacts/**' -- \
  bash docs/proof/record-front-matter-literal-write/run-proof.sh
```

| | |
|---|---|
| provider | `local-container` (runtime `docker`) |
| image | `node:24` → `v24.19.0`, `Linux aarch64` |
| lease | `cbx_241c2987e276` (`swift-crab`) |
| run | `run_b5d44f1fa51f` |
| base staged | `5439582b` · `src/clawsweeper-record-metadata.ts` · sha256 `df480a1e…83f1` |
| exit | `0` |

The script refuses to run below Node 24, builds the Node lane, then runs the
fixtures twice — once against the module compiled from the base commit and once
against this branch. Container images carry no `.git`, so `stage-before.sh` writes
the base version of the changed file into the proof package on the host and it
rsyncs in; when git *is* available the script re-derives that file, so it cannot
drift from the base commit.

**Observed result.** Pre-fix: 17 failing assertions. Beyond the label sets, a label
named `$'` also fails `title survives` and `delimiters are unchanged` — the
expansion drags following record text into the field, so the write damages the
record body, not just the one value. Keys containing `(`, `[`, `|`, `+` or `*`
either throw `SyntaxError: Invalid regular expression` or match the wrong line, and
`a.c` matches a line reading `aXc:`. On this branch every assertion passes. The
ordinary `["bug"]` case passes in **both** runs, and the script fails the proof if
that ever stops holding — which is what shows the change is confined to names
carrying a replacement pattern.

The 17 pre-fix failures inside the lease:

```
pre-fix exit: 1
  FAIL  labels ["$&"]          FAIL  labels ["$`"]        FAIL  labels ["$'"]
  FAIL  labels ["a$&b"]        FAIL  labels ["$$"]        FAIL  labels ["needs-triage","$&"]
  FAIL  JSON.parse round-trips ["$`"] / ["$'"] / ["$&"]
  FAIL  title survives         FAIL  delimiters are unchanged
  FAIL  key "a+b" / "x(y" / "k[0]" / "q|r" / "n*m" reads and updates in place
  FAIL  `a.c` does not match `aXc`
```

**Artifact / trace.**
`.crabbox/runs/run_b5d44f1fa51f/run_b5d44f1fa51f-artifacts.tgz` containing
`.artifacts/record-front-matter-literal-write-proof/` — `before-output.txt`,
`proof-output.txt`, `focused-tests.txt`, install and build logs.

**Limits.** This proves the record reader and writer in
`src/clawsweeper-record-metadata.ts`. The same raw-interpolation shape exists in
`src/repair/workflow-utils.ts`, `src/repair/create-job.ts`,
`src/repair/target-fanout.ts`, `src/repair/comment-router-core.ts`,
`src/commit-sweeper.ts`, `src/decision-packets.ts` and two scripts; those are out of
scope here and named in the linked issue. The proof does not call GitHub — it starts
from the label set the sync returns. No shipped caller passes a key containing regex
syntax today, so the key-escaping half hardens the advertised contract rather than
fixing an observed failure; the value half is the reachable defect.

### DIFF
--- a/docs/proof/record-front-matter-literal-write/.gitignore
+++ b/docs/proof/record-front-matter-literal-write/.gitignore
@@ -1,0 +1,3 @@
+# The pre-fix module staged by stage-before.sh. It is a verbatim copy of a file
+# already in git history, regenerated on demand, so it is not committed.
+before/
--- a/docs/proof/record-front-matter-literal-write/.gitignore
+++ b/docs/proof/record-front-matter-literal-write/.gitignore
@@ -1,3 +1,0 @@
-# The pre-fix module staged by stage-before.sh. It is a verbatim copy of a file
-# already in git history, regenerated on demand, so it is not committed.
-before/
--- a/docs/proof/record-front-matter-literal-write/README.md
+++ b/docs/proof/record-front-matter-literal-write/README.md
@@ -1,0 +1,100 @@
+# Proof: label names are written to a record verbatim
+
+## Claim
+
+A GitHub label whose name contains `$&`, `` $` ``, `$'` or `$$` is stored in a
+record's `labels:` front matter exactly as given, the record around it is left
+intact, and front matter keys are matched literally instead of being compiled as
+regular expressions.
+
+## Exercised surface
+
+`createRecordMetadata(...).replaceFrontMatterValue` and the matching readers
+(`.frontMatterField`, `.frontMatterValue`, `.frontMatterStringArray`) in
+`dist/clawsweeper-record-metadata.js`.
+
+`src/clawsweeper-apply-report-labels.ts` writes the synced label set with
+
+```ts
+markdown = replaceFrontMatterValue(markdown, "labels", JSON.stringify(item.labels));
+```
+
+`item.labels` is the label set returned by the label sync for the reviewed item, so
+the names come from the reviewed repository. `String.prototype.replace` treats `$&`,
+`` $` ``, `$'` and `$$` in its *replacement* argument as expansion patterns, so those
+names were being expanded against the match instead of inserted.
+
+## Scenario / fixture
+
+`run-proof.mjs` starts from a record holding `repository`, `number`, `labels`,
+`title` and a body, then writes label sets through the shipped writer and reads them
+back through the shipped readers:
+
+1. **Verbatim storage** — ordinary names plus every replacement pattern, including a
+   mixed set (`["needs-triage", "$&"]`) that mirrors a real sync result.
+2. **Valid JSON** — the stored text must still `JSON.parse` back to what was written,
+   because `frontMatterStringArray` silently falls back to comma-splitting when it
+   does not.
+3. **Record integrity** — `title`, `number`, the body, and the `---` delimiters must
+   be unchanged after the write.
+4. **Literal keys** — keys containing `.`, `+`, `(`, `[`, `|` and `*` must read and
+   update in place, and `a.c` must not match a line reading `aXc:`.
+
+## Command and environment
+
+```
+bash docs/proof/record-front-matter-literal-write/stage-before.sh
+crabbox run --provider local-container --local-container-image node:24 --no-hydrate \
+  --artifact-glob '.artifacts/**' -- \
+  bash docs/proof/record-front-matter-literal-write/run-proof.sh
+```
+
+Recorded run:
+
+| | |
+|---|---|
+| provider | `local-container` (runtime `docker`) |
+| image | `node:24` → `v24.19.0`, `Linux aarch64` |
+| lease | `cbx_241c2987e276` (`swift-crab`) |
+| run | `run_b5d44f1fa51f` |
+| base staged | `5439582b` · `src/clawsweeper-record-metadata.ts` · sha256 `df480a1e…83f1` |
+| exit | `0` |
+
+`stage-before.sh` exists because container images carry no `.git`: it writes the
+base version of the changed file into `before/` on the host, where rsync picks it up
+as a dirty file. When git *is* available `run-proof.sh` re-derives that file, so the
+staged copy cannot drift from the base commit.
+
+The script refuses to run below Node 24, installs the pinned pnpm into a
+user-writable prefix, builds the Node lane, and runs the fixtures twice: once
+against the module compiled from the base commit and once against this branch.
+
+## Observed result
+
+Against the pre-fix module 17 assertions fail. Beyond the label sets themselves,
+a label named `$'` also fails `title survives` and `delimiters are unchanged` — the
+expansion pulls the text following the match into the field, so the write damages the
+record body, not just the one value. Keys containing `(`, `[`, `|`, `+` or `*` either
+throw `SyntaxError: Invalid regular expression` or match the wrong line, and `a.c`
+matches `aXc`. Against this branch every assertion passes. The ordinary `["bug"]`
+case passes in both runs, which is what shows the change is confined to names that
+carry a replacement pattern; the script fails the proof if that case ever regresses.
+
+## Artifact / trace
+
+`.crabbox/runs/run_b5d44f1fa51f/run_b5d44f1fa51f-artifacts.tgz` holds
+`.artifacts/record-front-matter-literal-write-proof/` with `before-output.txt`,
+`proof-output.txt`, `focused-tests.txt`, and the install/build logs.
+
+## Limits
+
+This proves the record reader and writer in
+`src/clawsweeper-record-metadata.ts`. The same raw-interpolation shape exists in
+`src/repair/workflow-utils.ts`, `src/repair/create-job.ts`,
+`src/repair/target-fanout.ts`, `src/repair/comment-router-core.ts`,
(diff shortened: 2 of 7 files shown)


dwin-gharibi:
@steipete

brokemac79:
Thanks @dwin-gharibi — merged as 04cd809. This ensures dollar-pattern label names are stored literally without corrupting review records. Appreciate the contribution and validation work.

PR #1148, opened 2026-08-12
fix(reconcile): supersede proven stale dead letters

steipete:
## What Problem This Solves

Production reconcile run [31648987909](https://github.com/openclaw/clawsweeper/actions/runs/31648987909) at 2026-08-12T22:56Z was the first run with the skip taxonomy from [openclaw/clawsweeper#1147](https://github.com/openclaw/clawsweeper/pull/1147). It reported 125 skipped targets, with `head_mismatch: 121`, `blocked_alias: 3`, `active_work: 1`, and zero recoveries. The open DLQ tail held 191 rows: 158 `retry_exhausted`, 31 `tuple_protocol_invalid`, and two `workflow_cancelled`.

These head mismatches are immutable publication artifacts for old pull-request heads. Replaying them would publish stale review state, but leaving every row open forever also leaves the DLQ without an endgame.

## Why This Change Was Made

The existing HMAC-authenticated canonical record route is strong enough to prove supersession without adding a new endpoint: `GET /internal/state/records/<repo-slug>/items/<number>`. Its Durable Object envelope supplies canonical content, digest, monotonic revision, and update time; the report binds repository, item number, pull-request type, review status, and reviewed head.

The reconciler now resolves a head-mismatched row as superseded only when that digest-valid record identifies the same pull request, says the review is complete, and records the live GitHub head, which must differ from the dead-letter head. It then immediately re-reads GitHub and requires the same open node identity and head before mutation. Missing, malformed, failed, wrong-target, or wrong-head records leave the row open as `head_mismatch_unproven`; mutation-time drift leaves it open as `head_mismatch_revalidation_changed`.

The authenticated lifecycle audit inventory remains corroborating evidence only. It exposes completed/superseded lanes and canonical receipt outcomes, but deliberately omits source heads, so it cannot independently authorize this resolution.

## User Impact

Each cycle checks at most ten head-mismatch targets and resolves at most 20 supported rows per proven target. Successful resolutions include the full newer head and exact canonical endpoint in the audit note, use the existing alias guard, and increment publication completed/superseded totals without counting a publish. Dry-run reports the plan without mutation.

`tuple_protocol_invalid` and `workflow_cancelled` rows are explicitly excluded. The stage never re-enqueues reviews and never resolves from absence of evidence.

## Evidence

RED on fresh `origin/main` at `9a257905e50be2dff9bb99afecb6cde50f8417f9` retained 81 passing operator scenarios while the four new behavior scenarios failed at the missing contract. After a build, the dedicated Worker RED test also failed on the absent superseded metric accounting.

GREEN passes all 224 focused operator and queue-runtime scenarios. The complete `pnpm run check` gate passes 3,391 tests with nine platform skips and zero failures. Pre-commit and committed-branch Codex autoreviews report no accepted/actionable findings.

The first ClawSweeper verdict raised one accepted P1 finding: a pull request could advance after canonical evidence was read but before the Worker mutation. Fix round 1 adds the immediate same-open-node/same-head recheck and a B→C race fixture that proves zero resolution or recovery on drift. The full gate, autoreview, and container proof were refreshed for the new implementation head.

## Real Behavior Proof

Proof status: sufficient.

Docker-backed Crabbox ran the real operator CLI and Durable Object route through loopback queue, GitHub, canonical-record, resolution, and metrics seams on committed implementation head `10b45760a0c0fe1adf156909e2df0d737d5fef6a`.

- Provider: `local-container`
- Lease: `cbx_a7a8d9e75f26` (`brisk-shrimp-3201`)
- Image: `node:24-bookworm`
- Result: 224 passed, 0 failed; exit 0, including the mutation-time B→C race
- Cleanup: lease stopped automatically and is absent from the provider inventory
- Receipt: `docs/proof/head-mismatch-supersede/receipt.json`
- Contract: `docs/proof/head-mismatch-supersede/behavior-contract.md`
- Source-blind result: `docs/proof/head-mismatch-supersede/behavior-report.json`

The refreshed proof artifacts were committed with fix round 1 in the same push. The branch head is `e44f2bccdbd6da8058c9226e224c39ecd696d93f`; the proof-only child commit does not change the executed code.

The proof uses deterministic loopback fixtures. It does not mutate the production queue, consume GitHub quota, or claim that any particular production row already has matching canonical evidence.

## OpenClaw Bay

No Bay change is needed. This remains queue-owned observer-only state: existing aggregate metrics and the dynamic skip-reason map reflect the new outcome, while Bay gains no queue, workflow, GitHub, recovery, deploy, or rollback action.


steipete:
@clawsweeper re-review

PR #1189, opened 2026-08-17
fix(workflow): publish hot-intake review comments

brokemac79:
## Summary

- route broad hot-intake review results through the existing selected-comment publication path
- keep the background comment-sync route exclusive to non-hot broad work
- add a production-shaped routing and publication regression covering hot intake, explicit items, repository dispatch, scheduled/background producers, and replay idempotency

## Problem

The broad `workflow_dispatch` hot-intake path has no `item_number` or `item_numbers` input. After its review artifacts were committed, `Dispatch background review comment sync` excluded hot intake while `Sync selected review comments` required an explicit item or a non-target repository dispatch. Both terminal-comment routes therefore evaluated false.

This was observed on [openclaw/openclaw#125204](https://github.com/openclaw/openclaw/issues/125204): intake and the model/artifact run completed, but the broad hot workflow skipped both comment-sync steps. Its valid start lease then delayed the exact worker until expiry; the terminal response was eventually published roughly 38 minutes after issue creation.

## Implementation

`Sync selected review comments` now also selects `needs.plan.outputs.hot_intake == 'true'`. It reuses the existing target-write token, sync implementation, canonical state publication, action ledger, fencing, retry, and idempotency behavior. No queue, lease, credential, or publication implementation changes are introduced.

The regression asserts exactly one terminal sync route for:

- broad workflow-dispatch hot intake
- normal broad workflow dispatch
- explicit item and item-list dispatches
- hot and normal target repository dispatches
- exact-item repository dispatch
- scheduled background review

It also executes the real `apply-decisions --sync-comments-only` boundary twice against controlled GitHub state: the first pass publishes the final ClawSweeper comment and the replay is a no-op. Existing lease/fencing and immutable-ledger suites remain unchanged and pass.

## Validation

- `pnpm run build:all`
- focused Node test: `review publication routes hot intake once without changing other producers`
- `node --test --test-concurrency=1 test/sweep-workflow.test.ts` — 121/121 passed on Linux
- focused apply-label lease/fencing suite — 4/4 passed
- focused action-ledger idempotency suite — 2/2 passed
- `pnpm exec oxfmt --check .github/workflows/sweep.yml test/sweep-workflow.test.ts`
- YAML parse of `.github/workflows/sweep.yml`
- actionlint v1.7.12, pinned release archive SHA-256 and repository config
- `git diff --check origin/main...HEAD`

## Real Behavior Proof

[...]

PR #1194, opened 2026-08-17
fix(dashboard): retain Bay projection across queue probe failures

brokemac79:
## Summary

- retain the last complete verified public Bay queue projection when the optional queue `/stats` probe rejects, including after the ordinary 60-second status cache has expired
- clear generation-coupled live activity, keep status degraded/incomplete, and render retained queue references with lower-bound (`≥N`) counts
- bound retention to the existing 15-minute stale window, re-run data through the current public repository allowlist, and fail closed for absent bindings, fulfilled nulls, malformed data, or expired nested samples

## Problem

The Bay intermittently becomes entirely empty while the direct `/api/exact-review-queue` endpoint still returns a complete public projection with active and pending work. During status refresh, `attachExactReviewQueueStatus` tolerates optional queue probe failures with `Promise.allSettled`, but a rejected queue probe leaves `exact_review_queue` as `null`. That `null` is persisted and cached, erasing the last verified public snapshot and making a transient dependency failure look like zero work.

The ordinary cached-status read also expired after 60 seconds, earlier than the intended 15-minute stale fallback window. Retaining a queue projection without adapting the Bay client was insufficient because the client previously discarded all queue references whenever the live activity census was incomplete.

This is an observability defect. It can hide a real review/publication backlog, but it does not itself stop reviews. The durable-lifecycle `over_cap` state and the terminal-review completion outage are separate conditions.

## Implementation

When—and only when—the optional queue probe rejects, status refresh may load and retain the prior queue projection through the existing 15-minute stale window if the nested queue projection is complete and its own generation timestamp remains inside that bound. Before retention, the prior object is reprojected through `publicExactReviewQueueProjection` with the current verified-public repository allowlist.

Generation-coupled live activity is replaced with the canonical incomplete activity shape. The Bay client still renders verified queue references and queue-stage counts, labels them as lower bounds (`≥N`), and does not invent live workers. A fulfilled `null` result, an absent queue binding, malformed/partial input, or a nested queue sample older than 15 minutes still fails closed.

## Validation

- exact committed-tree AWS Crabbox Linux behavior proof: 139/139 scoped tests
- focused retained-projection and Bay client regressions
- dashboard build plus queue public-boundary and strict checks
- dashboard and target lint
- focused formatting and exact committed-blob checks
- `git diff --check origin/main...HEAD`

## Real Behavior Proof

**Claim:** A status refresh that loses only the optional queue probe retains the last complete, privacy-filtered public Bay queue projection through the bounded stale window, clears unverifiable live activity, and renders queue-only references/counts without presenting them as a complete census. An environment with no queue binding and an expired nested sample both fail closed.

**Exercised surface:** the real in-memory ExactReviewQueue, `/api/exact-review-queue`, `/api/status`, persisted status storage, response caching, health diagnostics, public projection filtering, and the production Bay client parser/renderer.

**Scenario/fixture:** enqueue public `openclaw/openclaw#125204`, obtain its complete public projection through the direct queue API, seed it as the prior status snapshot, then make `/stats` reject with a private marker while keeping the events endpoint available. Repeat after the ordinary 60-second cache age but within 15 minutes. Negative cases remove the queue binding, return fulfilled null/malformed data, and age the nested queue projection beyond 15 minutes.

**Command/environment:** AWS Crabbox Linux (`c7a.8xlarge`), provider `aws`, run `run_904311dcdb8d`, lease `cbx_68e44965aaf4`, exact committed-tree head `bea3ef7083fc79dd8cb48d81c2a1ef6e4c7e6416`, proof base `866506fb0f57da42f4664647b3d8619f9944475d`.

**Observed result:** `/api/status` retained the allowlisted queue reference in response, persistence, and cache projections after a rejected probe and after the ordinary cache age elapsed. It cleared live activity, remained degraded/incomplete, withheld the private failure marker, and the Bay rendered queue-only references with lower-bound counts. No binding, fulfilled null, malformed/partial input, and a nested projection older than 15 minutes all failed closed. All 139 tests and the boundary/type/lint/format gates passed.

**Artifact/trace:** [Crabbox run `run_904311dcdb8d`](https://crabbox.openclaw.ai/portal/runs/run_904311dcdb8d), lease `cbx_68e44965aaf4`; terminal proof line: `CSW-132 Bay projection Linux behavior proof: PASS`.

**Limits:** This deterministic exact-runtime proof does not deploy or mutate production. The optional `pr-behavior-proof` skill named by workspace policy was unavailable in this session, so its required contract is reproduced explicitly here. Main later advanced through unrelated agent-runner and live-proof changes in #1196/#1198; the proved four-file tree remains unchanged.

## Reviews

- dirty Codex reviews identified and fixed a no-binding fail-open edge, retained stale live activity, indefinite nested retention, and premature 60-second source expiry; repeated dirty review clean
- committed Codex review against proof base `866506fb0f57da42f4664647b3d8619f9944475d`: clean
- local ClawSweeper committed-range review: `decision=keep_open`, `confidence=high`, `action=kept_open`; no GitHub mutation

## Risks and rollout

The retained projection may be stale, which is why health and activity remain degraded and counts are lower bounds. Retention is bounded by the existing 15-minute stale window, requires a previously complete public projection, and occurs only on an actual optional-probe rejection. It cannot introduce unverified repositories, preserve live activity, or convert an absent binding into apparent data.

Rollback is to revert this PR. No data migration, queue repair, workflow change, or production-state mutation is required.

This PR fixes Bay projection continuity only. It does not repair the separate durable-lifecycle over-cap state, publication backlog, acute model/output generation failures, or hot-intake terminal-comment routing defect tracked in #1189.

## Related

- CSW-132 missing-review and Bay incident investigation
- #1189


brokemac79:
@clawsweeper review

brokemac79:
@clawsweeper re-review

PR #1228, opened 2026-08-23
fix(validation): only exempt string-shaped deprecated fields from install-destination scan

### LINKED ISSUE #1227: `deprecated` field in a lockfile package record trips the install-destination validator
`assertStructuredInstallMetadataDestinations` in `src/repair/target-validation.ts` scans package records for install destinations and rejects any URL that isn't an approved registry. It exempts the `funding` field from that scan, since funding metadata is copied verbatim from the registry and can legitimately contain arbitrary URLs, but it doesn't exempt `deprecated`, which is the same kind of verbatim registry text.

pnpm mirrors npm's deprecation notice into a package's `deprecated` field as-is, and that free text often contains a URL, commonly a GitHub issue link explaining the deprecation. `openclaw/openclaw`'s `pnpm-lock.yaml` has exactly this on `@aws-sdk/core@3.977.1`:

```yaml
'@aws-sdk/core@3.977.1':
  resolution: {integrity: sha512-...}
  engines: {node: '>=20.0.0'}
  deprecated: |-
    Deprecated due to Document number parsing bug in JSON, see
      https://github.com/aws/aws-sdk-js-v3/issues/8246. Newer version available.
```

That GitHub link trips the scan and aborts `prepareTargetToolchain` before any real dependency install happens. Production run [32367984753](https://github.com/openclaw/clawsweeper/actions/runs/32367984753) failed exactly this way on `openclaw/openclaw#126659`, throwing `target dependency install destination is not approved: https://github.com`. The issue being worked in that run (an Android `location.get` bug) had nothing to do with dependency installation, the toolchain-preparation gate itself misfired.

This isn't rare. Across the last 100 runs of the repair cluster worker workflow, 67 failed, and this was the dominant cause. There's no open issue, PR, or fix commit for it right now.

This is the same class of false positive already fixed for the `integrity` field in #649, just for a different field.

**Fix shape:** extend the existing `funding` exemption to also cover `deprecated`, but narrower, since `funding` can be a string, object, or array while npm's registry only ever emits `deprecated` as a plain string. The exemption should require `typeof entry === "string"` so a malformed object- or array-shaped `deprecated` field (not a real registry shape) still hits the fail-closed scan. A regression test should also confirm that a malicious `resolved`/`tarball` URL sitting beside a legitimate `deprecated` string in the same package record is still rejected, the exemption shouldn't widen to sibling fields.

corvid-reads:
Closes #1227

## What Problem This Solves

This bug is the single largest cause of ClawSweeper's own auto-implementation pipeline failing outright. Of the last 100 runs of the repair cluster worker workflow, 67 failed, and this was the dominant cause across that sample: `prepareTargetToolchain` throws `target dependency install destination is not approved` and aborts before any real dependency install is attempted, whenever a target repository's lockfile has a `deprecated` package field containing a URL, which is normal for npm's registry-generated deprecation notices, not something a target repo did wrong. `openclaw/openclaw` currently has one such dependency (`@aws-sdk/core@3.977.1`), so this isn't a rare or synthetic case, it's actively blocking automated fixes on the org's own primary repository right now, and would block any other target repo with a similarly deprecated transitive dependency.

In concrete terms: this is not a cosmetic validator quirk. It is ClawSweeper failing to complete the majority of its own automated repair attempts, for a reason that has nothing to do with whether the underlying fix was correct.

## Why This Change Was Made

`assertStructuredInstallMetadataDestinations` already exempts the `funding` field from the install-destination URL scan on package records, since funding metadata is copied verbatim from the registry and can contain arbitrary URLs. `deprecated` is the same kind of verbatim registry text and needs the same treatment, but not the identical exemption: `funding` can legitimately be a string, object, or array, while npm's registry only ever emits `deprecated` as a plain string. So the new exemption requires `typeof entry === "string"`, meaning a malformed object- or array-shaped `deprecated` field (not a real registry shape) still hits the fail-closed scan. This mirrors the fix already shipped for the `integrity` field in #649.

## User Impact

This unblocks the repair cluster worker for the majority of its currently-failing runs. Before this fix, any target repository with a `deprecated` transitive dependency containing a URL could not get past toolchain setup at all, no fix attempt, however correct, could ever be validated or land. That's a two-thirds failure rate on the workflow ClawSweeper's automated fixes depend on, not a narrow edge case. No behavior changes for any other field; `resolved`/`tarball` URLs and non-string `deprecated` values are still rejected exactly as before.

## OpenClaw Bay Impact

Not affected. This is a repair-worker toolchain-preparation validator, not a lifecycle, queue, status/telemetry, or dashboard data-contract surface.

## Documentation Impact

No user-facing documentation needed updating; this is an internal validator fix with no public API or config surface. The `docs/proof/lockfile-deprecated-field-network-policy/` package added here is historical (completed proof for this fix), not an active runbook.

## Evidence

**Real Behavior Proof**

Claim: `prepareTargetToolchain` no longer throws `target dependency install destination is not approved` when a target repository's lockfile has a `deprecated` package-record field whose text embeds a URL, while a malicious `resolved`/`tarball` URL anywhere in the same lockfile, including beside an exempt `deprecated` string in the same package record, is still rejected.

Exercised surface: `assertStructuredInstallMetadataDestinations` / `assertManifestDependencyDestinations` / `assertApprovedInstallUrl` in `src/repair/target-validation.ts`, reached through the public `prepareTargetToolchain` entrypoint.

Scenario/fixture: four unit fixtures in `test/repair/target-validation.test.ts` (a real-world `deprecated` string that must pass, a dependency literally named `deprecated` that must still fail, a non-string `deprecated` value that must still fail, and a `deprecated` string sitting beside a malicious `resolved` URL that must still fail on the `resolved` field), plus a production fixture: a fresh clone of `openclaw/openclaw` at its actual head, which carries the real triggering lockfile entry, run through `prepareTargetToolchain` with `installTargetDeps: true`.

Command and environment: Docker-backed Crabbox `local-container` lease, `node:24-bookworm`, Node v24.19.0, pnpm 11.10.0. Full detail in `docs/proof/lockfile-deprecated-field-network-policy/receipt.json`.

Observed result:
- Unfixed code, production fixture: throws `target dependency install destination is not approved: https://github.com`, reproducing production run [32367984753](https://github.com/openclaw/clawsweeper/actions/runs/32367984753) exactly, the same failure class behind 67 of the last 100 repair cluster worker runs.
- Fixed code, production fixture: does not throw that error, run twice against two separate fresh clones for consistency.
- Focused suite: `test/repair/target-validation.test.ts`, 202 passing, 3 skipped for a documented kernel-capability gate unrelated to this change, 0 failing.
- `format:check`, `lint`, `build:all`: all clean.

Artifact/trace: `docs/proof/lockfile-deprecated-field-network-policy/red-green.md`, `receipt.json`, `run-proof.sh`.

Limits: the production fixture uses an 8-second proof budget, sufficient to prove the network-policy rejection is gone, not a full dependency install. The full `pnpm run check` gate wasn't run (its `test:coverage:no-build` step needs CI-only secrets and fails identically on unmodified `main`), and no local Codex `/review` pass was performed (that loop is GitHub-tooling-side). Both are stated in the proof `README.md` rather than presented as complete.

I also re-ran `build`, `lint`, `format:check`, and the full `test:repair` suite (129/129 passing) against the current branch head just now as a fresh sanity check, all clean.

### DIFF
--- a/docs/proof/lockfile-deprecated-field-network-policy/README.md
+++ b/docs/proof/lockfile-deprecated-field-network-policy/README.md
@@ -1,0 +1,22 @@
+# Lockfile "deprecated" field misclassified as an unapproved install destination
+
+Production run [`32367984753`](https://github.com/openclaw/clawsweeper/actions/runs/32367984753) (`Execute credited fix artifact` step, 2026-08-20) stopped ClawSweeper's automatic implementation for [openclaw/openclaw#126659](https://github.com/openclaw/openclaw/issues/126659) with `target dependency install destination is not approved: https://github.com`, thrown from `assertApprovedInstallUrl` before any real dependency install was attempted. The issue being worked (Android `location.get` ignoring `maxAgeMs`) has nothing to do with dependency installation; the toolchain-preparation gate itself was misfiring.
+
+`openclaw/openclaw`'s `pnpm-lock.yaml` carries a `packages:` entry for a deprecated transitive dependency, `@aws-sdk/core@3.977.1`, whose `deprecated` field is pnpm's verbatim copy of the npm registry's own deprecation notice for that version:
+
+```yaml
+'@aws-sdk/core@3.977.1':
+  resolution: {integrity: sha512-...}
+  engines: {node: '>=20.0.0'}
+  deprecated: |-
+    Deprecated due to Document number parsing bug in JSON, see
+      https://github.com/aws/aws-sdk-js-v3/issues/8246. Newer version available.
+```
+
+That text is free-form maintainer-authored content, never an install source, but `assertStructuredInstallMetadataDestinations` (`src/repair/target-validation.ts`) only exempted the `funding` field from the install-destination URL scan for package records. `deprecated` wasn't exempted, so the GitHub issue link embedded in the deprecation notice tripped the scan and aborted `prepareTargetToolchain` before the fix worker ever ran. This is the same class of false positive fixed narrowly for `integrity` hashes in [#649](https://github.com/openclaw/clawsweeper/pull/649), a different field.
+
+**Fix**: extend the existing package-record exemption to also cover `deprecated`, but narrower than `funding`'s exemption, since npm's registry only ever emits `deprecated` as a plain string (unlike `funding`, which is legitimately a string, object, or array), the exemption requires `typeof entry === "string"`, so a malformed object- or array-shaped `deprecated` field (not a real registry shape) still hits the fail-closed scan. A companion regression test proves a malicious `resolved` URL sitting beside a legitimate `deprecated` string in the same package record is still rejected, the per-field exemption does not widen to sibling fields.
+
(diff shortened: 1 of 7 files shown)


clawsweeper:
Thanks @corvid-reads for the contribution — this is now merged.

- Reviewed head: `d5f7fbb6ce91752c48db66d3dc3a38e80a91099b`
- Squash merge: `1db170c5bf7ad850739e38ba2e29ab1647de0bba`

The landed change keeps the exemption narrowly limited to string-valued `deprecated` package metadata while preserving fail-closed validation for real or malformed install destinations. The focused positive and negative coverage was especially helpful.

PR #1230, opened 2026-08-23
fix(scripts): use build:node for test:repair so it emits dist/clawsweeper.js

### LINKED ISSUE #1229: `test:repair` script (and ClawSweeper's live-verification of it) fails on a clean checkout, missing the main build
`package.json`'s own `test:repair` script fails on a clean checkout:

```
"test:repair": "pnpm run build:repair && pnpm run test:repair:no-build"
```

`test/helpers.ts`, imported by most files under `test/repair/`, imports `../dist/clawsweeper.js`, an output of the **main** build (`tsc -p tsconfig.json`, the `build` script), not the repair build (`tsc -p tsconfig.repair.json`, which excludes `src/clawsweeper.ts` entirely). `build:repair` alone never produces `dist/clawsweeper.js`, so `pnpm run test:repair` fails with `ERR_MODULE_NOT_FOUND` on any checkout that doesn't already have a stale `dist/` from an earlier full build.

Reproduced directly: a fresh clone with `pnpm install` and no prior build, running `pnpm run test:repair`, fails to even load nearly the entire repair test suite:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/.../dist/clawsweeper.js' imported from /.../test/helpers.ts
```

Rebuilding with `pnpm run build:node` instead (the existing `build && build:repair` script) and rerunning `test:repair:no-build` against the same checkout: 1077 of 1078 tests pass, with one unrelated cancellation in `target-validation.test.ts` (a pending-promise timeout, looks like scratch-environment resource contention, not a module-resolution failure).

This is the same defect class as #946 / #948, a job or script requesting `build:repair` when the code path it exercises needs `dist/clawsweeper.js`, and `build:node` already exists as the fix for exactly this. That PR's own scope was explicitly two CI workflow jobs invoking the bundle directly, and its added guard test (`test/repair/workflow-sparse-checkout.test.ts`) only audits `.github/workflows/*.yml` job definitions for direct `dist/clawsweeper.js` invocations. `test:repair` in `package.json` was never updated to use `build:node`, and isn't covered by that guard since it's a package script, not a workflow job, so this slipped through as a distinct call site of the same underlying problem.

It surfaced concretely on #1228: ClawSweeper's live-verification step ran `pnpm run build:repair && node --test test/repair/target-validation.test.ts` and hit exactly this error on a correct patch, for reasons unrelated to the change under review. The review still reached `proof: sufficient` by falling back to the committed static proof artifact, so it didn't block the PR, but it's a false negative in the live check, and it would happen to any contributor running the repo's own documented `test:repair` script fresh.

**Fix shape:**

1. **Primary, deterministic fix:** change `test:repair` in `package.json` from `pnpm run build:repair && pnpm run test:repair:no-build` to `pnpm run build:node && pnpm run test:repair:no-build`, matching the pattern #948 already established, rather than the broader `build:all` (no need to build the dashboard leg for this target). Confirmed working above.
2. **Secondary, complementary:** update `prompts/review-item.md`'s `liveProofPlan` guidance to prefer invoking this repo's own defined `pnpm run` scripts over hand-composing `tsc`/`node --test` commands, so live-verification doesn't need to independently reconstruct a build sequence that a correctly-fixed `test:repair` would already encode.
3. **Optional, deeper follow-up:** `test/helpers.ts` importing a main-build artifact at module scope is what created this coupling for every repair test file that imports it, not just `target-validation.test.ts`. Splitting a repair-only helpers module that doesn't import `dist/clawsweeper.js` would remove the dependency rather than just building around it.

Option 1 is the narrow, low-risk, immediately verifiable fix. Options 2 and 3 are complementary and deeper, respectively, and don't need to block on each other.

corvid-reads:
Closes #1229

## What Problem This Solves

`pnpm run test:repair`, the repo's own documented standalone test command, fails on a clean checkout before running a single test. Most files under `test/repair/` import `test/helpers.ts`, which imports `../dist/clawsweeper.js`, an output of the main build. `test:repair` only ran `build:repair` (`tsc -p tsconfig.repair.json`, which excludes `src/clawsweeper.ts` entirely), so on a fresh checkout with no prior build, `pnpm run test:repair` failed with `ERR_MODULE_NOT_FOUND` before loading nearly the entire repair suite, not just one file. This also produced a false negative in ClawSweeper's own live-verification of #1228, where the exact same error appeared on an unrelated, correct patch, because it hand-composed `pnpm run build:repair && node --test test/repair/target-validation.test.ts` instead of using the repo's own (broken) script.

## Why This Change Was Made

Two fixes, both recommended by ClawSweeper's review of #1229:

1. **`test:repair` script fix.** `build:node` (`pnpm run build && pnpm run build:repair`) already exists in `package.json` for exactly this defect class, added in #948 to fix two CI workflow jobs that invoked `dist/clawsweeper.js` directly without building it. That fix's own guard test (`test/repair/workflow-sparse-checkout.test.ts`) only audits GitHub Actions workflow job definitions for direct bundle invocations, not `package.json` scripts, so this call site slipped through as a distinct instance of the same underlying problem rather than being caught by that guard. This change points `test:repair` at the script that already does the right thing, and updates the existing contract test (`test/run-node-tests.test.ts`) that had locked in the old, broken composition.
2. **Review-prompt guidance fix.** `prompts/review-item.md`'s `liveProofPlan` instructions had no guidance steering a reviewer toward this repo's own defined scripts, so ClawSweeper's model reconstructed a build/test command from scratch and got it wrong in exactly the way (1) fixes. Added a sentence to that section: for terminal verification, prefer a repository-defined `pnpm run` script over hand-composing individual build/test commands when one already expresses the needed setup. This doesn't require the model to know this specific import-graph fact, it just needs to reach for an existing script instead of reinventing one.

## User Impact

Contributors and CI running `pnpm run test:repair` on a clean checkout now get a working test run instead of an immediate module-resolution crash. Future ClawSweeper live-verification passes on this repo are steered toward using existing correct scripts rather than re-deriving build sequencing per review, reducing the odds of the same false-negative class recurring on a different test file. No behavior change for anyone who already had a full `dist/` build present.

## OpenClaw Bay Impact

Not affected. This is a package script, its own contract test, and reviewer-prompt wording, not a lifecycle, queue, status/telemetry, or dashboard data-contract surface.

## Documentation Impact

None needed for `README.md`, its existing "Standalone `test`, `test:repair`, and coverage commands still build their required outputs" line already documents the intended contract; this change makes the code satisfy it. `prompts/review-item.md` itself is the documentation being updated as part of the fix.

## Evidence

**Real Behavior Proof**

Claim: `pnpm run test:repair` on a clean checkout (no prior build) builds and runs the repair test suite successfully instead of failing with `ERR_MODULE_NOT_FOUND` before any test loads.

Exercised surface: the `test:repair` script in `package.json`, its dependency on `test/helpers.ts` → `dist/clawsweeper.js`, and the `liveProofPlan` section of `prompts/review-item.md`.

Scenario/fixture: a fresh clone of `corvid-reads/clawsweeper` with `pnpm install` and no build step, before and after this patch.

Command and environment: local Node v26.7.0, pnpm 11.10.0.

Observed result:
- Before this patch (fresh clone, `build:repair` only): `pnpm run test:repair` throws `Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../dist/clawsweeper.js' imported from .../test/helpers.ts`, failing to load nearly the entire repair suite.
- After this patch (fresh clone, `build:node`): `pnpm run test:repair` runs to completion, 1275 passing, 3 failing, 0 cancelled out of 1278. The 3 failures are pre-existing and unrelated: process-containment tests requiring kernel capabilities (delegated user namespaces, Landlock ABI 3+) not available in every environment, the same class of test the project's own Crabbox proof documentation already skips for the identical reason.
- Contract test: `test/run-node-tests.test.ts`'s `composed no-build scripts preserve standalone build contracts` test, updated to expect `build:node`, passes.
- `format:check`, `lint`, `build:node`: all clean.
- Full unit suite (`node scripts/run-node-tests.mjs unit`): 2405 passing, 1 failing, 1 skipped out of 2407, both runs (before and after this patch). The 1 failure is a pre-existing locale/number-formatting mismatch (`>2,000<` vs `>2 000<`), confirmed present identically on unmodified `upstream/main`, unrelated to this change.

Artifact/trace: command output captured directly above; no separate proof directory added, this is a small, directly reproducible before/after change.

Limits: the 3 containment-test failures and the 1 locale-test failure are both pre-existing environment limitations, confirmed unrelated to this change by reproducing them on unmodified `upstream/main`, not something this PR attempts to fix. The `prompts/review-item.md` change is a prompt-wording addition; its effect can only be observed in a future live ClawSweeper review, not exercised directly by a local test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "test": "pnpm run build:all && pnpm run test:no-build",
     "test:no-build": "node scripts/run-node-tests.mjs all",
     "test:unit": "node scripts/run-node-tests.mjs unit",
-    "test:repair": "pnpm run build:repair && pnpm run test:repair:no-build",
+    "test:repair": "pnpm run build:node && pnpm run test:repair:no-build",
     "test:repair:no-build": "node scripts/run-node-tests.mjs repair",
     "test:workflow-sparse-checkout": "node --test test/repair/workflow-sparse-checkout.smoke.ts",
     "test:coverage": "pnpm run build:all && pnpm run test:coverage:no-build",
--- a/prompts/review-item.md
+++ b/prompts/review-item.md
@@ -502,7 +502,12 @@ user-visible. Reserve `not_applicable` for a change with genuinely nothing to
 run, such as docs-only edits or generated assets in a repository with no
 meaningful runnable surface. Use `surface: "browser"` for browser behavior and
 `surface: "terminal"` for terminal behavior. The `entry` must be a URL path for
-browser verification or a command for terminal verification. Emit at most ten
+browser verification or a command for terminal verification. For terminal
+verification, prefer invoking a repository-defined `pnpm run` (or equivalent
+package-manager) script over hand-composing individual build/test commands
+whenever one already expresses the needed setup, so the plan does not need to
+independently reconstruct build or dependency sequencing that the script
+already encodes correctly. Emit at most ten
 deterministic, typed `steps`: browser plans may use `goto`, `click`, `fill`,
 `press`, `wait_for`, `wait`, and `expect_text`; terminal plans may use `run`,
 `wait`, and `expect_output`. Include at least one concrete expectation of real
--- a/test/run-node-tests.test.ts
+++ b/test/run-node-tests.test.ts
@@ -99,7 +99,7 @@ test("composed no-build scripts preserve standalone build contracts", () => {
   const scripts = packageJson.scripts as Record<string, string>;
 
   assert.match(scripts.test, /^pnpm run build:all && pnpm run test:no-build$/);
-  assert.match(scripts["test:repair"], /^pnpm run build:repair && pnpm run test:repair:no-build$/);
+  assert.match(scripts["test:repair"], /^pnpm run build:node && pnpm run test:repair:no-build$/);
   assert.match(scripts["test:coverage"], /^pnpm run build:all && pnpm run test:coverage:no-build$/);
   assert.match(
     scripts["test:coverage:changed"],


clawsweeper:
Thanks @corvid-reads for another well-scoped contribution — this is now merged.

- Reviewed head: `bb615763155e71d535b91e0d89bba27a04c38b2e`
- Squash merge: `d5b0aa42ff03edf2752d0f8b960c3a60bb70a246`

The clean-checkout baseline reproduced the missing main-bundle failures, and the landed `build:node` composition removed that failure class while keeping the script and prompt changes tightly scoped. The focused contract test, representative repair test, hosted CI, and CodeQL all passed.

PR #1242, opened 2026-08-25
fix: keep projected GitHub reads out of ETag cache

### LINKED ISSUE #1239: apply: ETag cache key omits the jq projection, so the requested-reviewer close guard reads another call site's body and fails open
## Summary

The durable ETag cache key does not include the `--jq` output projection, so two reads of the same GitHub route that ask for different projections share one durable entry. When the second read gets a 304, the broker returns the first read's body, digest-verified and treated as authoritative. In the apply lane this makes `pullRequestHumanEngagementBlockReason` read a `{body}` projection where it expects `{requested_reviewers, requested_teams}`, so the "requested reviewers or teams indicate active review signal" guard never fires and a pull request with a live requested reviewer becomes eligible for auto-close.

## Environment

- Commit: ed20a56037dc7e1141140513e0307df69f0d394a (origin/main at time of filing)
- Surface: apply lane (`apply-decisions`) when the ETag read broker is active, which is `EXACT_EVENT_PUBLICATION=true`. That is set at `.github/workflows/sweep.yml:1564` and `:2608` for `repair:publish-event-result`, which spawns `apply-decisions` with `env: process.env` (`src/repair/publish-event-result.ts:427`, `:832`), so the broker is live in the event-review publication lane.

## Steps to reproduce

1. In an apply run with the broker active, let any call site read `repos/{owner}/{repo}/pulls/{number}` with `--jq '{body}'`. `src/clawsweeper-apply-close-execution.ts:353` does this on the implemented-on-main close path.
2. Let the inactivity guard read the same route with `--jq '{requested_reviewers:[...],requested_teams:[...]}'`. `src/clawsweeper-apply-guard-activity.ts:199` does this.
3. The second read hits the first read's durable entry, sends `If-None-Match`, and GitHub answers 304 because the resource itself has not changed.

## Expected

Each call site receives the projection it asked for. A pull request with a requested reviewer raises the active-review-signal block reason and is not auto-closed.

## Actual

The guard receives `{"body":"Fixes #9"}`. `requested_reviewers` is `undefined`, `(pull.requested_reviewers ?? []).length > 0` is false, no block reason is returned, and the close proceeds.

## Root cause

`src/github-etag-cache-contract.ts:69-74` builds the key from four dimensions, none of which is the output projection:

```ts
const cacheKey = JSON.stringify([
  GITHUB_ETAG_CACHE_KEY_VERSION,
  options.credentialPool,
  route,
  mediaType,
]);
```

`src/clawsweeper-github-runtime.ts:401-408` decides which invocations may use the broker, and its disqualifier list does not contain `--jq` or `-q`:

```ts
["-i", "--include", "--paginate", "--slurp", "-f", "--raw-field", "-F", "--field"]
```

`githubApiEndpointForEtag` (`:437-458`) additionally steps over `-q`, `--jq`, `-t` and `--template` as value flags, so projected reads deliberately reach the broker. The 304 path itself is rigorous: it verifies the etag and `sha256(body)` against the stored entry. That is exactly why the wrong body is trusted. The digest confirms the body matches what was stored, and what was stored was another call site's projection.

## Sibling surfaces

Route `pull_metadata` is read with at least five incompatible shapes on origin/main: the full object (`clawsweeper-apply-guard-policy.ts:54`, `clawsweeper-coverage-proof.ts:93`, `clawsweeper-close-promotion.ts:80`), `{body}` (`clawsweeper-apply-close-execution.ts:353`, `clawsweeper-status-context.ts:834`), `{requested_reviewers,requested_teams}` (`clawsweeper-apply-guard-activity.ts:199`), a field subset (`clawsweeper-status-context.ts:641`, `clawsweeper-context-hydration.ts:523`), and `.head.sha // empty`, which is a bare string.

Route `issue_metadata` has the same problem and one worse case: `clawsweeper-promotion-facts.ts:485` projects `[.labels[].name]`, a JSON array, against sites that expect the full issue object. An array reaching `asRecord()` yields `{}`, so labels, state and author association all read empty and the protected-label and maintainer-association guards mis-evaluate.

The existing coverage at `test/github-etag-read-broker.test.ts:25-50` enumerates four key dimensions (credential pool, media type, query, page) and does not mention output projection.

## Real behavior proof

Behavior addressed: guard reads another call site's projection through the durable ETag entry and stops blocking an auto-close
Real environment tested: real `githubEtagCacheKey` and real `durableGithubEtagReadSync` imported unmodified from the checkout at ed20a56037dc7e1141140513e0307df69f0d394a and driven directly under Node v24.18.0, with only the GitHub HTTP boundary and the durable store supplied; the store is keyed by `key.cacheKey`, which is how `dashboard/github-etag-cache.ts:40,78,130` keys it
Exact steps or command run after this patch: `node repro.ts`
Evidence after fix:
```
STEP 1 - real githubEtagCacheKey() for the two call sites
  guard site (--jq '{requested_reviewers,requested_teams}')
    [1,"target_app","/repos/openclaw/openclaw/pulls/1234","application/vnd.github+json"]
  close site (--jq '{body}')
    [1,"target_app","/repos/openclaw/openclaw/pulls/1234","application/vnd.github+json"]
  identical: true

=== OBSERVED on origin/main (cache key omits the projection) ===
STEP 2 - real durableGithubEtagReadSync(), close site reads the route first
  read 1 - close site asked for {body}
  conditional request sent: no (live 200)
  body handed back to the caller: {"body":"Fixes #9"}

STEP 3 - guard site now reads the same route
  read 2 - guard site asked for {requested_reviewers,requested_teams}
  conditional request sent: yes (304 answered)
  body handed back to the caller: {"body":"Fixes #9"}

STEP 4 - guard predicate from clawsweeper-apply-guard-activity.ts:205
  live truth on GitHub: requested_reviewers = [{"login":"octocat"}]
  value the guard actually saw: requested_reviewers = undefined
  guard raises 'requested reviewers or teams indicate active review signal': false

RESULT: guard did NOT block. A pull request with a live requested reviewer is eligible for auto-close.

=== EXPECTED with the projection folded into the cache key ===
STEP 2 - real durableGithubEtagReadSync(), close site reads the route first
  read 1 - close site asked for {body}
  conditional request sent: no (live 200)
  body handed back to the caller: {"body":"Fixes #9"}

STEP 3 - guard site now reads the same route
  read 2 - guard site asked for {requested_reviewers,requested_teams}
  conditional request sent: no (live 200)
  body handed back to the caller: {"requested_reviewers":[{"login":"octocat"}],"requested_teams":[]}

STEP 4 - guard predicate from clawsweeper-apply-guard-activity.ts:205
  live truth on GitHub: requested_reviewers = [{"login":"octocat"}]
  value the guard actually saw: requested_reviewers = [{"login":"octocat"}]
  guard raises 'requested reviewers or teams indicate active review signal': true

RESULT: guard blocked the close (correct)
```
Observed result after fix: on origin/main the guard reads `requested_reviewers = undefined` for a pull request that has a requested reviewer and stops blocking the close; folding the output projection into the durable key on identical inputs restores the correct value and the block
What was not tested: I did not run a live end-to-end `apply-decisions` job against a real repository, so the frequency of the two projections landing in the same durable entry within one entry lifetime in production is unmeasured. I also did not verify whether any later guard in the close chain independently re-reads reviewer state and would still block, and I did not exercise the `issue_metadata` array-projection variant described under Sibling surfaces.

### Suggested direction

Either fold the `--jq`/`-t` argument, or a digest of it, into `githubEtagCacheKey`, or add the projection flags to the disqualifier list at `src/clawsweeper-github-runtime.ts:402-408` the way `-i` and `--paginate` are already excluded.

goutamadwant:
Closes #1239

## What Problem This Solves

Exact-publication apply runs could reuse a durable ETag body produced by a different GitHub output projection. A cached projected response could therefore omit requested reviewers or teams from a later close-safety read.

## Why This Change Was Made

Projected `gh api` reads now bypass the durable ETag broker for jq and template output flags, including split, attached, compact, and clustered short-option forms. Full-resource GETs retain the existing durable cache path and cache-key contract.

## User Impact

Pull requests with active requested reviewers or teams are no longer exposed to auto-close eligibility because another call site cached a differently projected body.

## OpenClaw Bay Impact

OpenClaw Bay is unaffected. This changes apply-lane GitHub read admission only and does not alter Bay data contracts, status, or navigation.

## Documentation Impact

No documentation update is required. The existing ETag broker contract remains unchanged; this narrows broker admission for response projections.

## Evidence

- `node --version` (`v24.16.0`)
- `pnpm run build:all`
- `node --test test/github-etag-read-broker.test.ts` (6/6 passed)
- `pnpm run check:static`
- `pnpm run lint`
- `git diff --check origin/main...HEAD`
- `pnpm run check` passed static checks, all builds, lint, and changed coverage before the exhaustive suite reached an unrelated local-review fixture that did not terminate; hosted CI remains authoritative for the full gate.

### Claim

Projected GitHub reads execute their requested projection directly instead of consulting or replaying the durable ETag broker.

### Exercised Surface

The built production `createGitHubRuntime().ghWithPreparedTimeout()` path under exact event publication, using real read-only `gh api` calls and an intentionally unreachable broker endpoint.

### Scenario And Environment

Node 24 and GitHub CLI 2.92.0 read an existing public ClawSweeper pull request using split, attached, compact, and clustered projections. The broker endpoint was deliberately unavailable, so successful results demonstrate direct execution rather than broker fallback or replay.

### Command

```sh
node --input-type=module <<'EOF'
import { spawnSync } from "node:child_process";
import { pathToFileURL } from "node:url";

const { createGitHubRuntime } = await import(
  pathToFileURL(`${process.cwd()}/dist/clawsweeper-github-runtime.js`).href,
);
Object.assign(process.env, {
  EXACT_EVENT_PUBLICATION: "true",
  EXACT_REVIEW_QUEUE_URL: "http://127.0.0.1:9",
  CLAWSWEEPER_WEBHOOK_SECRET: "proof-secret-placeholder",
  GITHUB_REPOSITORY: "openclaw/clawsweeper",
});

const invocations = [];
const runtime = createGitHubRuntime({
  ROOT: process.cwd(),
  targetRepo: () => "openclaw/clawsweeper",
  run: (command, args, options = {}) => {
    invocations.push([command, ...args]);
    const result = spawnSync(command, args, {
      cwd: options.cwd ?? process.cwd(),
      env: { ...process.env, ...options.env },
      encoding: "utf8",
      timeout: options.timeoutMs,
    });
    if (result.error) throw result.error;
    if (result.status !== 0) throw new Error(result.stderr.trim() || `exit ${result.status}`);
    return result.stdout.trim();
  },
});

const route = "/repos/openclaw/clawsweeper/pulls/1236";
const body = JSON.parse(runtime.ghWithPreparedTimeout(["api", route, "--jq", "{body: .body}"], 20_000));
const reviewers = JSON.parse(runtime.ghWithPreparedTimeout(["api", route, "--jq", "{requested_reviewers,requested_teams}"], 20_000));
const attached = JSON.parse(runtime.ghWithPreparedTimeout(["api", route, "--jq={state: .state}"], 20_000));
const compact = JSON.parse(runtime.ghWithPreparedTimeout(["api", route, "-q{state: .state}"], 20_000));
const clustered = runtime.ghWithPreparedTimeout(["api", route, "-iq", "{state: .state}"], 20_000);

console.log(JSON.stringify({
  brokerEndpointReachable: false,
  directGhInvocations: invocations.length,
  body: typeof body.body === "string",
  reviewerArrays: Array.isArray(reviewers.requested_reviewers) && Array.isArray(reviewers.requested_teams),
  attached: typeof attached.state === "string",
  compact: typeof compact.state === "string",
  clustered: clustered.includes("HTTP/") && clustered.includes('"state"'),
  productionMutations: 0,
}, null, 2));
EOF
```

### Result

All five projected reads completed through direct GitHub CLI invocations and returned their requested shapes while the broker was unreachable. Production mutations: 0.

Recorded after-fix output for head `c6c0908b99ff824a0a83abe922e0b4a7c136b1a4`:

```json
{
  "exactEventPublication": "true",
  "brokerEndpointReachable": false,
  "directGhInvocations": 5,
  "firstProjectionReturnedBody": true,
  "secondProjectionReturnedReviewerArrays": true,
  "attachedProjectionReturnedState": true,
  "compactProjectionReturnedState": true,
  "clusteredProjectionReturnedHeadersAndState": true,
  "productionMutations": 0
}
```

### Limits

The live proof is read-only and did not run a live `apply-decisions` close attempt. Full-resource GET broker behavior remains covered by the existing focused tests.

Disclosure: AI was used to understand the codebase and review the fix.

### DIFF
--- a/src/clawsweeper-github-runtime.ts
+++ b/src/clawsweeper-github-runtime.ts
@@ -400,10 +400,25 @@ export function createGitHubRuntime(dependencies: CreateGitHubRuntimeDependencie
   ) {
     if (process.env.EXACT_EVENT_PUBLICATION !== "true" || args[0] !== "api") return null;
     if (
-      args.some((arg) =>
-        ["-i", "--include", "--paginate", "--slurp", "-f", "--raw-field", "-F", "--field"].includes(
-          arg,
-        ),
+      args.some(
+        (arg) =>
+          [
+            "-i",
+            "--include",
+            "--paginate",
+            "--slurp",
+            "-f",
+            "--raw-field",
+            "-F",
+            "--field",
+            "-q",
+            "--jq",
+            "-t",
+            "--template",
+          ].includes(arg) ||
+          arg.startsWith("--jq=") ||
+          arg.startsWith("--template=") ||
+          /^-i*[qt]/.test(arg),
       )
     ) {
       return null;
--- a/test/github-etag-read-broker.test.ts
+++ b/test/github-etag-read-broker.test.ts
@@ -8,6 +8,7 @@ import {
   GithubEtagResponseStore,
 } from "../dashboard/github-etag-cache.ts";
 import worker, { ExactReviewQueue, githubJsonForTest } from "../dashboard/worker.ts";
+import { createGitHubRuntime } from "../dist/clawsweeper-github-runtime.js";
 import {
   githubEtagCacheKey,
   githubEtagCacheRequestBody,
@@ -48,6 +49,58 @@ test("ETag keys fence credential pool, media type, query, and page", () => {
   assert.equal(page1.page, 1);
   assert.equal(page2.page, 2);
   assert.equal(page2.route, "/repos/openclaw/openclaw/issues/42/comments?page=2&per_page=100");
+});
+
+test("projected GitHub reads bypass the durable ETag broker", () => {
+  const keys = ["EXACT_EVENT_PUBLICATION", "EXACT_REVIEW_QUEUE_URL", "CLAWSWEEPER_WEBHOOK_SECRET"];
+  const previous = Object.fromEntries(keys.map((key) => [key, process.env[key]]));
+  Object.assign(process.env, {
+    EXACT_EVENT_PUBLICATION: "true",
+    EXACT_REVIEW_QUEUE_URL: "http://127.0.0.1:9",
+    CLAWSWEEPER_WEBHOOK_SECRET: "etag-broker-secret-placeholder",
+  });
+  const invocations: string[][] = [];
+  const runtime = createGitHubRuntime({
+    ROOT: process.cwd(),
+    targetRepo: () => "openclaw/openclaw",
+    run: (_command, args) => {
+      invocations.push(args);
+      return JSON.stringify({ projected: true });
+    },
+  });
+  const route = "/repos/openclaw/openclaw/pulls/1234";
+  const projections = [
+    ["--jq", "{body}"],
+    ["-q", "{requested_reviewers,requested_teams}"],
+    ["--template", "{{.head.sha}}"],
+    ["-t", "{{.state}}"],
+    ["--jq={body}"],
+    ["--template={{.head.sha}}"],
+    ["-q={body}"],
+    ["-q{body}"],
+    ["-t={{.state}}"],
+    ["-t{{.state}}"],
+    ["-iq", "{body}"],
+    ["-it", "{{.state}}"],
+  ];
+
+  try {
+    for (const projection of projections) {
+      assert.equal(
+        runtime.ghWithPreparedTimeout(["api", route, ...projection], 1_000),
+        JSON.stringify({ projected: true }),
+      );
+    }
+    assert.deepEqual(
+      invocations,
+      projections.map((projection) => ["api", route, ...projection]),
+    );
+  } finally {
+    for (const [key, value] of Object.entries(previous)) {
+      if (value === undefined) delete process.env[key];
+      else process.env[key] = value;
+    }
+  }
 });
 
 test("durable broker revalidates every read and keeps wire calls while avoiding quota charges", () => {


steipete:
Squash-merged as https://github.com/openclaw/clawsweeper/commit/a2641b27c403012451cf307d03fb699933fef6a5 and verified that commit is on remote main. https://github.com/openclaw/clawsweeper/issues/1239 auto-closed. Thanks @goutamadwant.

Validated exact head `c6c0908b99ff824a0a83abe922e0b4a7c136b1a4`: the current durable runner proof reports `node --test test/github-etag-read-broker.test.ts` passing 6/6 (https://github.com/openclaw/clawsweeper/pull/1242#issuecomment-5404506811). Hosted `pnpm check`, sparse build, and Windows launcher checks passed at https://github.com/openclaw/clawsweeper/actions/runs/32803164537; CodeQL passed too. Independent Codex review reported no accepted/actionable findings. Before merge, the head, body, and comments were unchanged, the source-revision hash matched the durable review, and no inline review comments were present.

The main PR body's built-runtime, real read-only GitHub proof shows five projected reads returning their requested shapes with the broker deliberately unreachable. No live apply-decisions closure was run. Projected reads now incur direct REST quota; that accepted trade-off prevents another projection's cached body from weakening close-safety reads while preserving full-resource caching. No safety policy or OpenClaw Bay contract changed.

PR #1244, opened 2026-08-25
fix: repair setup fails for target repos with no committed lockfile

### LINKED ISSUE #1243: Checkout-identity guard treats a freshly-materialized pnpm-lock.yaml as a mutation on lockfile-less target repos
## Summary

`prepareTargetToolchain`'s dependency install fails closed with `target dependency setup mutated checkout identity: contentTreeSha, status, worktreeSha256` on any target repo that has no committed `pnpm-lock.yaml` and no dependencies. This isn't a real checkout mutation. It's the install step doing exactly what it's supposed to do.

## Root cause

`openclaw/crabpot` has empty `dependencies`/`devDependencies` in `package.json` and no `pnpm-lock.yaml` in git. When ClawSweeper runs its install command (`pnpm install --frozen-lockfile --prefer-offline --ignore-scripts --ignore-pnpmfile ...`) against a repo with no lockfile, pnpm doesn't error. It generates a new `pnpm-lock.yaml` on disk and reports "Already up to date."

`validationSourceIdentity`'s `worktreeSha256` (via `worktreeContentSha256`) and `contentTreeSha` (via `rawWorktreeTreeSha`) both fold in untracked files through `git ls-files --others --exclude-standard`. `pnpm-lock.yaml` isn't gitignored in this repo, so the newly-created file gets picked up as untracked content, and the before/after identity comparison in `assertValidationSourceIdentity` (target-validation.ts:439) sees a real diff and throws.

## Reproduction

Confirmed locally against `openclaw/crabpot` at `0cb363fdae97a06cc91f96525529cb3907ae20ad`:

```
$ git status --porcelain
$ pnpm install --frozen-lockfile --prefer-offline --ignore-scripts --ignore-pnpmfile \
    --config.registry=https://registry.npmjs.org/ --config.engine-strict=false \
    --config.enable-pre-post-scripts=false
Already up to date
Done in 197ms using pnpm v11.22.0

$ git status --porcelain
?? pnpm-lock.yaml
```

The tree is clean before, has one untracked file after. That's enough to trip the guard.

## Production impact

This isn't intermittent. The same job, `automerge-openclaw-crabpot-283`, retried 5 times over 6 hours on 2026-08-24/25 and hit the identical failure at the same ~2s mark every time, against the same `source_head`. Any target repo with no committed lockfile hits this on every repair attempt. Sample runs:
- https://github.com/openclaw/clawsweeper/actions/runs/32812081965
- https://github.com/openclaw/clawsweeper/actions/runs/32804218844
- https://github.com/openclaw/clawsweeper/actions/runs/32803229484

## Suggested fix direction

Either exclude a newly-materialized, previously-absent, untracked lockfile from the post-install identity check the same way `runtimeInputsSha256` is already excluded via `ignoreRuntimeInputs`, or write the freshly-generated lockfile into the disposable validation profile instead of the checkout when none was committed, so the checkout doesn't actually change.

corvid-reads:
Closes #1243

## What Problem This Solves

Fixes an issue where ClawSweeper's dependency setup would fail with `target dependency setup mutated checkout identity` for any target repo that has no committed `pnpm-lock.yaml`. Every repair and automerge attempt against such a repo failed before validation even started. `openclaw/crabpot` hit this on every single retry.

## Why This Change Was Made

A frozen-lockfile install never rewrites a lockfile that already exists, but pnpm silently creates one when there's nothing to freeze against. That new file wasn't there before setup, so the post-install checkout-identity check read it as a mutation and aborted. The fix snapshots the pre-install lockfile state and, only when it was genuinely absent, removes the one pnpm just wrote before the identity check runs. A lockfile that was already tracked or already present untracked is left untouched, so real mutations still fail closed.

## User Impact

Target repos with no committed lockfile (zero-dependency packages, most commonly) can now go through repair setup at all. No change for repos that already have a lockfile.

## OpenClaw Bay Impact

None. This only touches `prepareTargetToolchain`'s local dependency setup, before anything reaches review, publication, or dashboard state.

## Documentation Impact

None needed beyond the added proof package. This is an internal correctness fix to an existing guard's behavior, not a new contract or option.

### Real Behavior Proof

Full controlled proof package, in the `docs/proof/<slug>/` format this repo's `AGENTS.md` asks for: [`docs/proof/crabpot-lockfile-checkout-identity/`](docs/proof/crabpot-lockfile-checkout-identity/README.md). Docker-backed Crabbox `local-container` lease (`cbx_3e31cb90a5dd`, `coral-krill`, `node:24-bookworm`), the committed `run-proof.sh` executed literally end to end in-lease, full focused test suite (203 passing, 3 skipped for a documented kernel-capability gate, 0 failing), clean build/lint/format, and a real `openclaw/crabpot@0cb363fdae97a06cc91f96525529cb3907ae20ad` RED/GREEN reproduction. See the package's README for one caveat worth flagging directly: this bug's own code path needs a completed namespace-contained install to reach, and this specific Docker lease can't itself nest that (confirmed symmetrically on unfixed and fixed code, not masking the bug), so the RED/GREEN comparison ran against a toolchain with working namespace/Landlock support instead of inside the lease; everything else in the package did run inside it.

**Claim:** `prepareTargetToolchain` no longer throws `target dependency setup mutated checkout identity` for a target repo with no committed lockfile, and leaves the checkout exactly as it found it.

**Exercised surface:** `preparePnpmToolchain` inside `src/repair/target-validation.ts`, the same install path every pnpm-based repair and automerge run goes through.

**Scenario:** the real `openclaw/crabpot` repository at `0cb363fdae97a06cc91f96525529cb3907ae20ad`, the exact commit that failed in production (see linked run below). No dependencies, no committed lockfile.

**Observed result, before fix (unmodified upstream code):**
```
RESULT: prepareTargetToolchain threw: target dependency setup mutated checkout identity: contentTreeSha, status, worktreeSha256
```

**Observed result, after fix:**
```
RESULT: prepareTargetToolchain succeeded, no identity mutation error
```
`git status --porcelain` on the checkout was empty both before install and after setup completed, no `pnpm-lock.yaml` left behind.

**Limits:** this reproduces and fixes the pnpm path specifically, matching the reported issue and its real-world trigger. `prepareNpmToolchain` already branches on lockfile existence and isn't affected. `prepareBunToolchain` uses the same unconditional `--frozen-lockfile` shape as the old pnpm code did, and may have the same latent issue for lockfile-less bun targets, but I haven't confirmed bun's actual behavior in that case and didn't touch it here, that's outside what this PR's evidence covers.

### Tests

Added `dependency setup removes a lockfile pnpm materializes where none existed` in `test/repair/target-validation.test.ts`, alongside the existing `pnpm lockfile fallback restores a pre-existing untracked lockfile exactly` and `dependency setup rejects tracked source mutation`, so the fix's scope is asserted from all three sides: absent, pre-existing untracked, and tracked.

Full results (build, lint, format, focused test suite) are in the proof package above; summary: all clean, 203 passing / 0 failing / 3 skipped (documented kernel-capability gate, unrelated to this change).

I can't run ClawSweeper's own Codex `/review` loop locally, that's GitHub-tooling-side. It runs now that the PR is open.

Production failure for reference: https://github.com/openclaw/clawsweeper/actions/runs/32812081965


corvid-reads:
@clawsweeper re-review

brokemac79:
Thanks @corvid-reads — this focused fix and thorough proof package made the lockfile-less target failure straightforward to verify.

PR #1251, opened 2026-08-26
fix(queue): keep the owed source-drift review when a completion callback is lost

### LINKED ISSUE #1245: Lost lease-completion callbacks drop the owed source-drift follow-up review and misclassify completion-failure runs
## Summary

When the `Complete exact-review queue lease` callback cannot reach the durable queue (outage, repeated 5xx, a non-whitelisted conflict), the aftermath is wrong on both of the run's possible conclusions:

1. Direct publication accepted: the run stays green by design and the workflow_run reconciler heals the still-leased item from the run conclusion alone. That bare-outcome heal deletes the item, silently discarding the source-drift follow-up review that the completion payload's `direct_lifecycle_requeue: true` would have created. An item whose source drifted during publication, exactly the item that most needs a fresh review, never gets one.
2. Direct publication not accepted (durable-queue lane): the run goes red, and the only operator-facing annotation blames `classification=codex_or_content_failure` even though its own fields show `review_outcome=success` and the actual failure was queue connectivity.

## Environment

- Commit: a958131e8846d0e07673655fb1db2795971e4c12 (origin/main at time of filing)
- Surfaces: `.github/workflows/sweep.yml` `event-review-apply` job (completion step and fail gate), `dashboard/exact-review-queue.ts` `/reconcile` and `/complete` handlers, `.github/workflows/exact-review-reconcile.yml` outcome mapping

## Steps to reproduce

Defect 1 (lost source-drift follow-up):

1. Lease a review item and deliver a direct publication receipt whose lifecycle is `{ kind: "requeue" }` through the real signed ingress route `/internal/exact-review/publication-results` (source drift detected during publication, follow-up owed).
2. Control: deliver the completion callback with `direct_lifecycle_requeue: true`. The item is requeued as a fresh `source_drift_requeue` revision.
3. Defect path: the completion POST fails all 3 in-step attempts. The run concludes green (the fail gate's first clause requires `steps.direct-exact-review-publication.outputs.accepted != 'true'`, sweep.yml:2073). The reconciler then posts the run's green conclusion to `/internal/exact-review/reconcile` exactly as `exact-review-reconcile.yml` maps it. The item is deleted instead of requeued.

Defect 2 (misclassified red run):

1. Same completion-callback failure in the durable-queue publication lane (`accepted != 'true'`, review generated and publication queued successfully, generation result `outcome=success`).
2. The fail gate's first clause fires on the completion failure and the step's `run` block emits its hardcoded annotation.

## Expected

1. A completion callback loss after an accepted direct publication should not erase the owed source-drift follow-up review. The lifecycle projection already holds the `requeue` terminal fact at heal time, so the information survives the lost callback; only the reconciler's bare-outcome completion discards it.
2. A red run caused by a completion-callback failure should say so. `classification=codex_or_content_failure` should be reserved for runs where the review itself failed.

## Actual

1. `/reconcile` completes the leased item with `finishExactReviewQueueItem(state, item, now, run.outcome, 0, false, undefined, this.random)` (dashboard/exact-review-queue.ts:3477), a bare outcome with no lifecycle consultation. For a green run this takes the `delete state.items[item.key]` path (dashboard/exact-review-queue.ts:12056). The `source_drift_requeue` revision that `/complete` would have created via `requeueDirectLifecyclePublicationSync` (dashboard/exact-review-queue.ts:2269, 7866) never exists, and with the item deleted the 130-minute lease-expiry backstop can no longer fire either. The drifted head goes unreviewed until some unrelated future event re-enqueues the item.
2. The fail gate's annotation is a single hardcoded string for every trigger path (sweep.yml:2091):

```yaml
echo "::error::Exact review generation failed: classification=codex_or_content_failure reservation=$RESERVATION_STATUS review_outcome=$REVIEW_OUTCOME review_exit=$REVIEW_EXIT_CODE"
```

Nothing else in the repository emits or parses that string, so it is the only classification an operator sees. The same mislabel covers the held-reservation deferral whose completion failed (`reservation=held`, review skipped). The completion step's real HTTP error is only buried log text of a `continue-on-error: true` step.

## Root cause (if known)

The completion callback is the only carrier of the completion metadata (`direct_lifecycle_requeue`, `requeue_latest`, `completion_kind`, `reason_code`), and both fallback paths that engage when it is lost ignore that metadata: the reconciler substitutes a bare run-conclusion outcome, and the fail gate substitutes a fixed classification.

Fix direction constraint worth recording: making the producer run red on a completion failure after an accepted publication is not a viable fix. The reconciler maps a red run to a failure completion for the still-leased item, which requeues a complete duplicate review of an already-published item, burns `reviewFailureAttempts` toward `review_retry_exhausted`, and contradicts the documented contract at dashboard/exact-review-queue.ts:2224 that an auxiliary-step failure must not make the reconciler requeue review work that already succeeded. A queue-side option: `/reconcile` could consult the lifecycle projection (which already records `terminalDisposition.kind === "requeue"` for the fenced revision) before finishing a leased review item with a bare success.

## Sibling surfaces

- https://github.com/openclaw/clawsweeper/issues/894 covered the inverse false-red on benign completion conflicts; its fix (https://github.com/openclaw/clawsweeper/pull/895) whitelists `lease_superseded` 409s inside the completion step and is untouched by this report. Genuine completion failures (outage, 5xx, unexpected conflicts) still take the paths above.
- https://github.com/openclaw/clawsweeper/pull/966 fixed a different path that wrongly burned failures as `codex_or_content_failure` (head-drift reservations), showing this classification is treated as meaningful; the completion-failure trigger path still hardcodes it.
- The publish job's `Fail unsuccessful exact review publication` guard and the terminal-acknowledgement job surface completion failures by reddening their runs; those jobs' items are publication or acknowledgement leases whose red-run reconciliation is the intended retry, so they do not have defect 1.

## Real behavior proof

Behavior addressed: a lost lease-completion callback after an accepted direct publication silently deletes the owed source-drift follow-up review, and the same loss in the durable-queue lane produces a red run blamed on codex_or_content_failure.

Real environment tested: the production Cloudflare worker and ExactReviewQueue Durable Object code at commit a958131e8846, driven end to end through the real signed ingress route /internal/exact-review/publication-results, the real /internal/exact-review/reconcile route with the exact payload shape exact-review-reconcile.yml sends for a green run, and the real /complete handler; storage substituted in memory at the Durable Object storage seam, everything else real. The fail gate's production if-expression was evaluated by GitHub's own Actions expressions engine and its run block executed verbatim with the scenario's step results.

Exact steps or command run after this patch: node repro-drift.ts and node repro-annotation.mjs from /tmp/openclaw-bug-repros/clawsweeper-completion-metadata, against the workflow file and dashboard sources at a958131e8846.

Evidence after fix:

```
[completion-delivered] direct publication receipt (source drift detected, follow-up requeue owed): HTTP 202 {"ok":true,"accepted":true,"deduped":false,"superseded":false,"superseded_revisions":[],"canonical_target_key":"openclaw/openclaw#703","fence_key":"openclaw/openclaw#703","state_commit_sha":"do-revision:4"}
[completion-delivered] completion callback delivered: {"ok":true,"requeued":true}
[completion-delivered] queue item afterwards: {"state":"pending","revision":5,"sourceAction":"source_drift_requeue"}

[completion-unreachable-green-run] direct publication receipt (source drift detected, follow-up requeue owed): HTTP 202 {"ok":true,"accepted":true,"deduped":false,"superseded":false,"superseded_revisions":[],"canonical_target_key":"openclaw/openclaw#703","fence_key":"openclaw/openclaw#703","state_commit_sha":"do-revision:4"}
[completion-unreachable-green-run] completion callback: POST /internal/exact-review/complete unreachable for all 3 attempts; workflow run stays green (fail gate first clause requires accepted != 'true'), item still leased
[completion-unreachable-green-run] reconciler heals green run: HTTP 200 {"ok":true,"reconciled":1,"requeued":0,"completed":1}
[completion-unreachable-green-run] queue item afterwards: DELETED - no source-drift follow-up review remains anywhere

scenario: durable-queue publication lane (direct publication not accepted), review generated and
          published fine, only the lease completion callback failed (queue outage, 3 attempts)
guard step if-expression evaluates: true -> step runs, job goes red
guard step run block executed verbatim, exit code 1, sole annotation emitted:
::error::Exact review generation failed: classification=codex_or_content_failure reservation=posted review_outcome=success review_exit=0
```

Observed result after fix: on identical inputs the delivered callback requeues the item as a fresh source_drift_requeue revision, while the green-run reconcile heal deletes it outright, and the red-lane annotation blames codex_or_content_failure while reporting review_outcome=success.

What was not tested: no live Cloudflare Durable Object or live sweep run was driven; the Durable Object storage API was substituted in memory while all queue, route, signature, and workflow logic stayed real; the reconciler's GitHub run lookup was not exercised, its documented conclusion-to-outcome mapping was applied to construct the reconcile payload.

yetval:
Fixes https://github.com/openclaw/clawsweeper/issues/1245.

When a review publishes successfully but its queue-completion callback is lost, the workflow reconciler currently uses only the green run conclusion. It deletes the lease even when the accepted publication receipt says a source-drift follow-up is owed.

The queue now reads that durable receipt and reuses its existing requeue operation. Accepted or deduplicated requeue receipts preserve one fresh review; superseded receipts and newer queued revisions do not manufacture another review. The workflow also reports queue-completion failures separately from Codex/content failures, without changing which runs fail or introducing another retry loop.

This refresh incorporates current main, preserves the contributor's implementation and credit, consolidates repeated regression setup, and retains the one-off runtime drivers in their immutable original commit rather than adding them to the maintained tree. The final diff is about 300 lines including tests and documentation, versus over 1,100 in the original proposal.

## Current behavior proof

- Candidate head: `7cbe4493cf05142ee3338c63e64bbb291e2e7da1`; tree `2e0888cdcc4280894627530b9b6fb6521890541b`.
- Baseline: `0d07398d4fabc829c25e22c19a06a0656f49030b`.
- Provider: AWS Crabbox; lease `cbx_7031f88109bf`; image `ami-0461d919be7deb53c`; Node `v24.18.1`, pnpm `11.10.0`, Wrangler `4.107.0`.
- Environment: fresh isolated checkout, no instance role (IMDS credential endpoint returned 404), no Tailscale, synthetic items, disposable SQLite, no production credentials or state mutations.
- Surface: production Worker signed publication/completion/reconcile HTTP routes and the real ExactReviewQueue Durable Object under local workerd.
- Commands: the immutable `docs/proof/exact-review-completion-metadata/run-proof.sh` and `annotation-proof.mjs` from commit `138eacf9b30ced7465be928d9d6022f5647fa7d9`, run separately against the baseline and integrated candidate. Reproduction commands are in the compact proof README retained by this PR.
- Proof run: `run_a106148f5b89`; artifacts: `proof-summary.json`, `transcript.md`, and annotation output. The original contributor head was independently checked first in `run_400419e5dd76`.

| Scenario | Baseline | Candidate |
| --- | --- | --- |
| Completion callback delivered | Pending source-drift revision 5 | Same |
| Callback lost after accepted requeue receipt | `requeued: 0, completed: 1`; item deleted | `requeued: 1, completed: 0`; pending revision 5, cleared lease, `source_drift_requeue`, terminal disposition `requeue` |
| Callback lost after superseded receipt | Completes without requeue | Same |

The real workflow annotation shell emitted `queue_completion_failure` for a successful review whose completion failed and for a held deferral whose completion failed. A genuine review failure remained `codex_or_content_failure`. The original failure-gate expression is unchanged.

Limits: initial lease setup and state inspection use proof-only seams. All subsequent signed HTTP, storage, receipt, reconciliation, and lifecycle transitions use production code. The proof supplies the terminal-run payload rather than calling GitHub's run lookup. Workflow expressions are evaluated locally and the shell executes under bash. This proves local workerd behavior, not Cloudflare edge availability; it does not induce an outage in production. The baseline ran from an archive of the stated main SHA and the candidate proof ran before commit, with the reviewed source tree carried unchanged into the candidate head.

## Validation and review disposition

- `pnpm run check`: passed; 4,140 tests passed, 8 skipped (4,148 total), plus all 13 static test cases.
- Focused lifecycle/workflow regression coverage passed. The consolidated receipt matrix covers accepted, repeated/deduplicated, superseded, and absent-requeue plans; repeating reconciliation creates no extra revision.
- Documentation checks passed after the final proof-instruction edit.
- Current-main conflict resolved while retaining the newer queue failure-reason argument.
- Original review's superseded-receipt finding remains fixed. No broader retry policy or capacity setting changed.
- The remaining rank-up/monitoring request is accepted: verify the Worker deployment and inspect live reconciliation/queue health after landing.

OpenClaw Bay needs no code change: the real before/after proof confirms that a `requeue` terminal fact leaves Bay event/pending tables empty and the public snapshot unchanged. Bay remains observer-only, with no new browser calls or mutation controls.


yetval:
Refreshed against the verdict at `1995f4f779c3`. New head `18af960afb1e`. The verdict reported no findings and no security items, so the production diff and the regression tests are byte-identical to the reviewed head; only the proof material changed.

**Add real behavior proof (Docker-backed runtime proof).** Done. The in-memory driver is replaced by `docs/proof/exact-review-completion-metadata/run-proof.sh`, which runs under Crabbox `local-container` (provider `local-container`, image `node:24-bookworm`) and starts the real Wrangler Worker and its `ExactReviewQueue` Durable Object on workerd. Both scenarios are driven over the real signed `/internal/exact-review/publication-results`, `/internal/exact-review/complete`, and `/internal/exact-review/reconcile` routes against real Durable Object SQLite storage. Before half on pristine main `afe976209aa5`, lease `cbx_f8b4d8594074`, run `run_3aaf48af0028`: the reconciler reports `requeued: 0, completed: 1` and the item is deleted. After half on `18af960afb1e`, lease `cbx_7a9d11ed8b8f`, run `run_29c8406b8f61`: `requeued: 1, completed: 0` and the same fresh `source_drift_requeue` revision the delivered callback produces. Artifact captured through `--require-artifact`. Full transcript is in the PR body.

**Resolve merge risk (P1, in-memory seams).** Same change addresses it. The only remaining substitution is the initial leased item, which is seeded through the production state writer inside the Durable Object because a real lease requires a GitHub dispatch; every transition after that point is real. That limit is stated in the body and in the proof README.

**Resolve merge risk (P1, Bay impact).** Stated and verified: no Bay change is needed and none is produced. `recordLifecycleTerminal` does call `syncBayLifecycle`, but `bayLifecycleEvent` maps only `review_completed_routed` and `failure` to a Bay outcome and returns null for `requeue`, so no Bay event, timing, or tide-buffer row is written, and there is no prior terminal fact to retract. The proof reads the `exact_review_lifecycle_bay_event_v2`, `exact_review_lifecycle_bay_pending_v2`, and `exact_review_lifecycle_bay_tide_buffer_v2` tables and the public `GET /api/durable-lifecycle-bay` snapshot on both sides of the recovery: all zero rows, snapshot identical, on the healed path and the delivered-callback control alike.

**Live Verification failure on the previous head.** That was my proof artifact, not the patch. `drift-proof.ts` imported the shared test harness, which imports `../dist/repair/*.js`, so it could not run in a checkout without a build. It has been removed rather than patched, because the container harness supersedes it with a strictly stronger version of the same two scenarios. `annotation-proof.mjs` is unchanged and depends only on repository dependencies.

Gates re-run inside the same local-container image: `check:static` and `lint:scripts` pass. `node --test` for the four queue and lifecycle suites is 230/230 and `test/sweep-workflow.test.ts` is 124/124 on a host with full tooling.

Nothing here is an open maintainer decision from my side. `@clawsweeper re-review`

yetval:
Refreshed against the verdict at `18af960afb1e`. New head `138eacf9b30ced7465be928d9d6022f5647fa7d9`, rebased onto main `db14db010bf3044be85f25fe40a587a5ca77523a` (43 commits, no conflicts). All 3 items are closed; nothing is left as a maintainer decision.

**[P1] Exclude superseded receipts from recovery requeues** (`dashboard/exact-review-queue.ts:3478-3481`): fixed, and the finding was correct. The recovery predicate now requires an `accepted` or `deduped` receipt outcome:

```ts
const directLifecycle = item.decision.publication?.directLifecycle;
const owedDirectLifecycleRequeue =
  run.outcome === "success" &&
  item.revision <= leaseRevision &&
  directLifecycle?.plan.kind === "requeue" &&
  (directLifecycle.receiptOutcome === "accepted" ||
    directLifecycle.receiptOutcome === "deduped");
```

Confirming your three evidence points independently, plus one more that makes the case stronger than stated. `sweep.yml:1614` wraps the whole block that can set `direct_lifecycle_requeue=true` in `if [ "${DIRECT_PUBLICATION_SUPERSEDED:-false}" != "true" ]`, so a superseded receipt sends a plain completion. But the closer analogue is the publication lane's own direct-lifecycle recovery step at `sweep.yml:2392-2396`, which short-circuits on `if [ "$receipt_outcome" = "superseded" ]` to `completion_kind=superseded reason_code=remote_newer_tuple` **before** it dispatches on `lifecycle_kind`. That is the same converted publication lease reading the same `directLifecycle` metadata, and it was already gated. `/reconcile` was the only path that was not, so this is now equivalent to the normal completion path rather than broader than it.

**Resolve merge risk (P1)** and **Complete next step (P2)**: both closed by that same change. This is maintainer option 1, and the production delta is 3 lines.

**Rank-up move, both controls in the Durable Object proof**: done. `run-proof.sh` now runs a third scenario, `completion-unreachable-superseded-receipt`. The superseded receipt is a real one rather than seeded state: the driver publishes revision 5 on the same fence first, so the production direct-publication store settles the run's own revision-4 receipt as `superseded` while it still carries `plan.kind === "requeue"`, which is exactly the shape the finding describes.

Before, on the previous head `18af960afb1e`:

```
Proof assertion failed: superseded_receipt_completes_without_requeue
  {"item_afterwards":{"state":"pending","revision":5,"lease_id":null,"source_action":"source_drift_requeue","admission_delivery_id":"direct-lifecycle-requeue:openclaw/openclaw#707:4","has_publication":false},
   "terminal_disposition":"requeue","healer_response":{"ok":true,"reconciled":1,"requeued":1,"completed":0}}
```

After, on `138eacf9b3`, identical inputs:

```
[completion-unreachable-superseded-receipt] healer /internal/exact-review/reconcile -> {"ok":true,"reconciled":1,"requeued":0,"completed":1}
[completion-unreachable-superseded-receipt] queue item afterwards: DELETED - no source-drift follow-up review remains anywhere
[completion-unreachable-superseded-receipt] lifecycle terminal disposition: superseded
```

The two existing scenarios are unchanged across both halves, so the accepted-receipt recovery this PR opened for still works and the delivered-callback control is unaffected.

**Regression coverage** (your note that the added test was accepted-path only): `reconciling a lost completion callback completes a superseded direct receipt` drives the real `/internal/exact-review/publication-results` route to a genuine `superseded` receipt, asserts the stored `directLifecycle` is `{plan: {kind: "requeue"}, receiptOutcome: "superseded"}`, then reconciles. It fails without the gate (`requeued: 1, completed: 0`) and passes with it (`requeued: 0, completed: 1`), keeping the `superseded` terminal disposition.

Acceptance criteria: `node --test test/dashboard-worker-publication-lifecycle.test.ts` 26/26, `node --test test/sweep-workflow.test.ts` 125/125, and the queue suites together 232/232. `check:static`, `lint:dashboard`, and `lint:scripts` pass.

One environment note on the third acceptance command: crabbox's `local-container` provider could not publish its SSH port on this host, so the driver was invoked with plain `docker run` on the same `node:24-bookworm` image instead. That changes the wrapper around the container and nothing inside it; the driver, the routes, and the artifact are the same, and `proof-summary.json` now records the head SHA `138eacf9b30ced7465be928d9d6022f5647fa7d9` itself.

Good catch on this one. It would have turned a lost callback into duplicate review work on items a newer publisher already owned.

steipete:
@clawsweeper re-review

PR #1338, opened 2026-09-01
fix(review): reuse exact Codex source cache during remote outages

vincentkoc:
## What Problem This Solves

Fixes an issue where repeated OpenClaw review setup for the same exact Codex pin failed when the source remote was temporarily unavailable, even though the local bare cache already contained the complete requested commit graph.

Observed failures:

- https://github.com/openclaw/clawsweeper/actions/runs/33506178940
- https://github.com/openclaw/clawsweeper/actions/runs/33507058170

## Why This Change Was Made

The setup action is the owner of Codex source provisioning. The workflow initially prepared exact Codex 0.151.0 commit `78c290807ce710180111df227df3b7a4fe845452`, but `runCodex` synchronously invoked the installer again and the installer fetched unconditionally before checking whether the exact cached graph was complete.

The installer now:

1. Reuses a cache only when it is a valid bare repository, its configured origin exactly equals the requested source URL, the exact requested tag peels to a commit, and `git fsck --connectivity-only --no-dangling <head>` confirms the requested graph after partial-clone/promisor configuration is cleared.
2. Performs zero network access when the requested tag is complete.
3. Fetches a missing requested tag exactly once from an otherwise valid cache.
4. Rebuilds before that single fetch when the cache is invalid, its origin differs, or the requested tag exists with an incomplete graph.
5. Fails immediately when the requested tag remains absent or incomplete after the fetch. It does not retry or fall back to a previous pin.

The exact-head clean checkout replacement, PR-review sibling symlink, path validation, environment exports, and `prepareOpenClawCodexSourceForReview` behavior are unchanged. Bash 3-compatible review-tree numeric validation replaces the extglob expression.

Production delta: `+36/-16`, net `+20`.
Test delta: `+238/-115`, net `+123`.

## User Impact

OpenClaw reviews can reuse a complete exact-pin Codex source cache while the remote is unavailable. Missing or incomplete requested pins still fail closed instead of silently reviewing against a previous pin.

## OpenClaw Bay Impact

Bay is unaffected. This changes source provisioning before review execution and does not alter observer, status, queue, telemetry, or dashboard contracts.

## Documentation Impact

No documentation change is needed. The existing exact-pin source-preparation contract remains the same; this repair makes its cache behavior match that contract.

No changelog entry is included because this is an internal review-infrastructure repair with its operational behavior and proof recorded here.

Follow-up: source-preparation failures need a distinct safe diagnostic classification/detail rather than `codex_execution` with null process metadata.

Follow-up: failed exact-review setup must release or promptly expire its lease so queued retries do not wait for the full lease window.

### Pre-fix reproduction

AWS Crabbox run: https://crabbox.openclaw.ai/portal/runs/run_c19792b36f3a

Using the actual installer, the run created a temporary Git remote and exact tag, invoked the installer to populate a complete same-pin cache, made the origin unavailable, then invoked the installer again:

```json
{"firstStatus":0,"secondStatus":128,"failedBeforeCodex":true}
```

This reproduces the unconditional pre-validation fetch failure in the original execution order.

### Automated validation

- `bash -n .github/actions/setup-openclaw-codex-source/install.sh`
- `node --test test/setup-openclaw-codex-source-action.test.ts`: 7 passed, 0 failed
- Fresh Linux Crabbox `pnpm run build:all && pnpm run test:unit && pnpm run check`: succeeded
  - 2,967 unit tests: 2,960 passed, 7 platform skips, 0 failed
  - Full run: https://crabbox.openclaw.ai/portal/runs/run_dce37e978958
- `git diff --check`
- Local autoreview with `gpt-5.6-sol`, high reasoning: correct, no accepted findings, confidence 0.98

The focused behavior cases cover same-pin offline reuse; cached changed-pin offline retargeting with wrong/dirty checkout replacement; one-fetch missing-pin recovery at the peeled commit; incomplete object-graph rebuild; unavailable-remote fail-closed behavior without previous-pin fallback; sibling linking; and escaped-manifest rejection. Fixtures disable commit and tag signing.

Mutation proof is preserved by the pre-fix actual-installer failure and independent cases that detect unconditional fetch, missing-tag false success, omitted connectivity validation, retry/fallback behavior, and skipped checkout replacement.

### Codex hard gate

Personally inspected `openai/codex` source at `codex-rs/cli/src/main.rs:220-245` and `codex-rs/cli/src/main.rs:2840-2870`. These sections cover CLI command/completion dispatch and prompt CRLF normalization/test setup; they do not provide or perform OpenClaw's Codex source-cache fetching. The source-preparation action remains the correct owner.

## Real Behavior Proof

- **Claim:** a complete exact-pin cache permits a second same-pin installer invocation with no network access.
- **Exercised surface:** `.github/actions/setup-openclaw-codex-source/install.sh`.
- **Scenario:** create a temporary Git source and exact tag, run the actual installer once, remove the source remote, then run the actual installer again against the same cache.
- **Command and environment:** final branch bytes on a fresh AWS Linux Crabbox using `crabbox run --no-hydrate`; both invocations execute the actual installer.
- **Observed result:**

```json
{"firstStatus":0,"secondStatus":0,"secondFetches":0,"exactHead":true,"clean":true}
```

- **Artifact/trace:** https://crabbox.openclaw.ai/portal/runs/run_2789ed0dbf1f
- **Limits:** the fixture uses a temporary local Git remote rather than GitHub. It exercises the real installer, bare cache, tag peel, object connectivity, offline reuse, and final checkout state; broader repository validation is covered by the separate fresh-Linux full gate above.

### DIFF
--- a/.github/actions/setup-openclaw-codex-source/install.sh
+++ b/.github/actions/setup-openclaw-codex-source/install.sh
@@ -54,7 +54,7 @@ pin_root="$(cd "$pin_dir_input" && pwd -P)"
 review_tree_root="$review_artifact_root/review-trees"
 if [[ "$pin_root" != "$target_root" ]] &&
   { [[ "$(dirname "$pin_root")" != "$review_tree_root" ]] ||
-    [[ "$(basename "$pin_root")" != +([0-9]) ]]; }; then
+    [[ "$(basename "$pin_root")" == *[!0-9]* ]]; }; then
   echo "Codex version pin must come from the target checkout or one of its PR review trees." >&2
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
Maintainer landing note: accepted the reviewed explicit one-fetch policy: a complete matching exact-pin cache skips network access; missing or incomplete requested pins fetch exactly once, then fail closed with no previous-pin fallback. The latest durable exact-head review listed no `Rank-up moves`, so there were none to apply or skip. Squash landed as `6ce4fe3c6bab4478fd540446e27fd1264acc3cb2`.

PR #1299, opened 2026-08-29
fix(state): bound record export candidates

brokemac79:
## Summary

Bound canonical record-export preflight work to the same 50-record budget used
for reconstruction. Preserve pagination with a bounded terminal lookahead.

## Incident

Two scheduled canonical Worker runs failed during `hydrateState` while exporting
`openclaw-openclaw` records. The Worker correctly failed closed with the public
`500 exact_review_queue_unavailable` response; no incomplete dispatch occurred.

The live Durable Object exception is intentionally redacted, so this change does
not claim a verified lower-level Cloudflare error. It fixes a verified
availability/correctness problem in the export path: source-byte preflight joined
up to the request limit (100) even though reconstruction cannot consume more than
50 records.

## Implementation

- Cap the metadata-joining candidate query at `min(limit, maxRecords)`.
- Probe for a continuation only after a full materialized page.
- Add `(repo_slug, section, store_revision)` for the section-filtered lookahead,
  so it cannot scan unrelated later sections.
- Cover the candidate cap, multi-page manifest parity, terminal boundary, and
  SQLite query plan.

This remains fail-closed: it does not add retries, suppress notifications, alter
workflows/gates, or permit dispatch with incomplete state.

## Real Behavior Proof

Claim: the signed Worker-to-Durable-Object records export remains bounded to 50
source/reconstruction candidates, advances when more records exist, and does not
advertise an empty terminal page.

Exercised surface: local Worker route `/internal/state/records/export` with the
SQLite-backed `ExactReviewQueue` Durable Object harness.

Scenario and fixture: 58 canonical/backfill records, including source-byte and
serialized-size boundaries, materialized over pages `[50, 6, 1, 1]` with exact
manifest parity; a separate 50-item terminal page followed by 100 later `closed`
records proves the lookahead is section-indexed and returns `nextCursor: null`.

Commands and observed result:

```text
pnpm run build:dashboard
pnpm run build:node
node --test test/record-export-bounds.test.ts test/dashboard-worker-durable-object-error.test.ts
pnpm run lint:dashboard
git diff --check
```

All passed: 10 focused Worker/DO/export tests and dashboard lint.

Docker-backed proof: fresh Crabbox local-container lease
`cbx_3a943b51e1a0`, image `mcr.microsoft.com/playwright:v1.60.0-noble`, ran the
same dashboard/node builds and 10 focused tests successfully; the lease was then
released. The controlled proof has no production Cloudflare CPU claim.

## Review

[...]

PR #1340, opened 2026-09-01
fix(state): retry canonical record reads

brokemac79:
## Summary

- give canonical Worker record reads a three-attempt retry budget
- wait 30 seconds after the first eligible failure and 60 seconds after the second
- keep unrelated signed requests on their existing short retry schedule
- validate successful endpoint envelopes inside the same bounded retry budget
- retry malformed successful snapshot chunks inside that same budget

## Problem

Scheduled state hydration can fail closed during setup when the authenticated canonical-record boundary returns the sanitized `exact_review_queue_unavailable` 500. The failure recurred after the bounded record-export repair in #1299. Public evidence still does not expose the lower-level Worker/Durable Object exception, so this change does not claim to repair that unobserved root cause; it adds bounded client resilience for the read-only, idempotent side of the boundary.

## Implementation

Canonical record reads now make at most three attempts, with delays of 30 seconds and 60 seconds between eligible failures. The extended schedule is limited to:

- record export journal pages
- stored snapshot metadata and chunks
- repository slug discovery
- canonical item listing
- direct canonical record reads

Eligible failures are transport errors, 5xx responses, malformed/invalid successful response envelopes, and invalid successful snapshot chunk status/range/length. The attempts share one budget per request or chunk. Deterministic 4xx responses and pre-fetch configuration errors still fail immediately. After the third eligible failure, hydration still fails closed with the original sanitized Worker error status/code where available.

The exported generic `signedPost` path retains its existing 250 ms / 500 ms schedule, so write-capable or unrelated callers do not inherit the long waits.

## Validation

Base: `48bd2b42f1dd0504c9afc8643c9781290604b3b2`

Head: `e5b25f3f7980af50c68782095604df8ea9155fa0`

- `node --test test/worker-records-request.test.ts` — 21/21 passed
- focused malformed-response hydration tests — 3/3 passed on Windows
- `pnpm run build:all` — passed
- `pnpm exec oxlint scripts/worker-records.ts test/worker-records-request.test.ts test/repair/focused-state-hydration.test.ts` — zero warnings/errors
- `pnpm exec oxfmt --check` on the three changed files — passed
- `git diff --check` — passed on the host checkout

One existing test in `focused-state-hydration.test.ts` checks for a literal `/records` suffix and therefore misses the simulated rename failure on Windows. The same failure reproduces at the exact base commit and the unchanged test passes in the Linux proof below; it is unrelated baseline platform-fixture noise.

## Real Behavior Proof

**Claim:** the production canonical-record client retries read-only transient failures on the requested 30-second / 60-second schedule, accepts a later valid response, rejects deterministic failures immediately, and remains fail closed after exhaustion.

**Exercised surface:** the real `scripts/worker-records.ts` request, validation, pagination, snapshot, and focused-hydration paths. Timers and fetch responses were controlled; the production client implementation itself was not replaced.

**Scenario / fixture:** focused tests inject mixed 502/500 `exact_review_queue_unavailable` responses, transport failures, malformed 2xx bodies/envelopes/rows/cursors, malformed successful snapshot chunk status/range/length, eventual valid responses, persistent 500s, deterministic 4xx responses, and pre-fetch configuration errors. Timer capture verifies the exact `[30000, 60000]` delay sequence without making the proof wait 90 wall-clock seconds. A separate loopback scenario uses the production client's global `fetch` over a real HTTP socket, validates the actual HMAC signature and request body at the server, returns the sanitized 500 once, and then returns a valid canonical export page.

**Command / environment:** Docker-backed Crabbox `local-container`, image `node:24-bookworm`, Node `v24.20.0`:

```text
C:\Users\marti\.local\bin\crabbox.exe run --provider local-container --local-container-image node:24-bookworm --no-hydrate --timing-json --script C:\clawsweeper-work\notes\csw-147-crabbox-proof.sh
```

The clean host checkout was at head `e5b25f3f7980af50c68782095604df8ea9155fa0` when Crabbox synchronized it.

**Observed result:** lease `cbx_70e698f9996b`, run `run_2d0d1bcc1043`, provider `local-container`, type `node_24-bookworm`, exit `0`. All 32 affected Linux tests passed, including the staged-install restoration fixture, the signed real-transport recovery case, and malformed successful snapshot-chunk recovery; all three TypeScript builds passed; focused oxlint completed with zero warnings/errors.

**Trace:** the passing test output explicitly records the signed loopback HTTP recovery, the three-attempt 30s/60s case, the shared mixed-failure budget, malformed envelope/row/cursor and snapshot-chunk retries, immediate configuration failure, persistent fail-closed snapshot reads, canonical item pagination, and 32 tests / 32 passed / 0 failed.

**Limits:** the loopback case proves recovery through the production client's real HTTP/HMAC transport, but it is not a live Cloudflare Worker/Durable Object trace. This proof does not contact Cloudflare, rerun a scheduled workflow, mutate a queue or Durable Object, or demonstrate recovery from a live production incident. The public error remains redacted, so the lower-level cause of the recurrent 500 is still unknown. Production workflow-timeout impact is not empirically measured; a persistent eligible failure can add 90 seconds per failed canonical read before the existing fail-closed result.

## Review Closeout

- dirty-patch Codex review: no actionable defects after accepted fixes for endpoint-envelope validation and pre-fetch error classification
- committed branch Codex review against `origin/main`: no actionable correctness issues
- first hosted ClawSweeper review: accepted P2 finding that malformed successful snapshot chunks bypassed the shared retry budget; fixed in `974ec0fb0b` with focused coverage
- user-requested timeout revision: reduced the delays from 60s/180s to 30s/60s in `e5b25f3f79`, reducing per-read wait from four minutes to 90 seconds
- final dirty-patch and committed-range Codex reviews: no actionable findings; the committed review preserved immediate 4xx handling, fail-closed exhaustion, and short retries for generic callers
- final local committed-range ClawSweeper review: `keep_open`, high confidence, overall correctness `patch is correct`, no review comments, security cleared; its current-head proof gap is satisfied by the exact-head Crabbox evidence above

## Risks / Rollout

- persistent availability failures take up to 90 additional seconds per affected canonical read before failing closed
- multi-read hydration can still accumulate waits; affected job timeouts range from 10 to 360 minutes, including 30 minutes for scheduled target fanout
- the repair does not diagnose or mask the underlying Worker/Durable Object failure
- no Worker, Durable Object, workflow, queue, configuration, deployment, notification, or public error-redaction behavior changes

Related: CSW-147; follow-up to #1299.

### DIFF
--- a/scripts/worker-records.ts
+++ b/scripts/worker-records.ts
@@ -115,6 +115,25 @@ export class WorkerRecordRequestError extends Error {
   }
 }
 
+type SignedRequestRetryDelays = readonly [number, number];
+
+const DEFAULT_SIGNED_REQUEST_RETRY_DELAYS_MS = [250, 500] as const;
+const WORKER_RECORD_READ_RETRY_DELAYS_MS = [60_000, 180_000] as const;
+
+type SignedRequestOptions = {
+  baseUrl: string;
+  path: string;
+  webhookSecret: string;
+  body: unknown;
+  method?: "GET" | "POST";
+  fetch?: typeof globalThis.fetch;
+  retryDelaysMs?: SignedRequestRetryDelays;
+};
+
+type SignedPostOptions = Omit<SignedRequestOptions, "method"> & {
+  validateResponse?: (value: unknown) => boolean;
+};
+
 // Cold hydration (no stored snapshot yet) replays the full journal from
 // revision 0, so it must stay a small-repo affordance: a slug whose record set
 // outgrows this bound has earned a real snapshot and still refuses cutover.
@@ -155,7 +174,7 @@ export async function exportWorkerRecords(options: {
   let cursor: number | null = 0;
   let revision = sinceRevision;
   do {
-    const page = await signedPost<ExportPage>({
+    const page = await signedWorkerRecordReadPost<ExportPage>({
       baseUrl: options.baseUrl,
       path: "/internal/state/records/export",
       webhookSecret: options.webhookSecret,
@@ -167,6 +186,30 @@ export async function exportWorkerRecords(options: {
         limit: options.limit ?? 100,
       },
       fetch: options.fetch,
+      validateResponse: (value) => {
+        const page = value as Partial<ExportPage> | null;
+        if (
+          typeof page === "object" &&
+          page !== null &&
+          !Array.isArray(page) &&
+          page.repoSlug === options.repoSlug &&
+          Number.isSafeInteger(page.revision) &&
+          Number(page.revision) >= 0 &&
+          Array.isArray(page.records) &&
+          (page.nextCursor === null ||
+            (Number.isSafeInteger(page.nextCursor) &&
+              Number(page.nextCursor) >= 1 &&
+              page.nextCursor !== cursor))
+        ) {
+          try {
+            for (const record of page.records) validateWorkerRecord(record);
+            return true;
+          } catch {
+            return false;
+          }
+        }
+        return false;
+      },
     });
     if (page.repoSlug !== options.repoSlug || !Number.isSafeInteger(page.revision)) {
       throw new Error("Worker returned an invalid record export envelope");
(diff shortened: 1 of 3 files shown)


brokemac79:
@clawsweeper review

brokemac79:
@clawsweeper re-review

brokemac79:
@clawsweeper re-review

PR #1371, opened 2026-09-03
fix(publish): prevent stale queue writes after transient failures

vincentkoc:
## What Problem This Solves

Resolves a publication safety issue where an ambiguous transient failure after a
review commit could be retried in-process after newer queue state, allowing a
stale post-effect request to arrive out of order.

## Why This Change Was Made

Post-effect enqueue, router-receipt, and terminal-disposition requests now make
one bounded request. Heartbeats retain their existing three-attempt, 45-second
retry window because they renew the currently owned lease rather than replaying
a post-commit state transition.

This is immediate containment. Worker replay ordering remains the durable
follow-up; this PR does not change Worker, workflow, schema, queue policy,
capacity, cadence, or live operations.

## User Impact

An ambiguous post-effect failure now fails the current publication run and
leaves unfinished work for scheduled recovery instead of risking a stale
in-process retry. This can delay later batch members, but preserves newer queue
state.

There is no end-user UI or Telegram-visible change.

## OpenClaw Bay Impact

Bay is unaffected. This PR changes only the batch client's retry admission and
does not modify Worker lifecycle projection, Bay synchronization, dashboard
source, or any public/private dashboard data contract. Failed post-effects still
leave the member unfinished so the existing release path returns it to scheduled
recovery.

## Documentation Impact

`docs/live-dashboard.md` is an active operator runbook. It now distinguishes
bounded heartbeat retries from deliberately one-shot post-effects.

- Role owner: ClawSweeper maintainers
- Runtime source of truth: `src/repair/exact-review-batch-queue-client.ts`
- Verified revision: `288389c13afc1e2d8079cbf7a2a0362f9a82037c`
- Update triggers: heartbeat/post-effect retry policy or Worker replay-ordering changes

## Evidence

Exact head: `288389c13afc1e2d8079cbf7a2a0362f9a82037c`

- Signed semantic commit: verified locally with a good ED25519 signature.
- Focused/expanded tests: 67 passed, 0 failed across the queue client, CLI, and
  unchanged batch workflow tests.
- Build: all three TypeScript projects passed (`tsconfig.json`,
  `tsconfig.repair.json`, and `tsconfig.dashboard.json`).
- Static gates: active-surface, dashboard queue boundary, dashboard strict,
  documentation, limits, repository formatting, repair/test lint, and
  `git diff --check` passed.
- Production delta: `+1/-1`; four files total.
- Precommit and committed-range Codex reviews found no actionable defect.

### Infrastructure disclosure

- AWS Crabbox could not create a lease: coordinator HTTP 500, error code 1101.
  Crabbox recorded a durable cancellation.
- The local-container backend was unavailable because the Docker socket was not
  running.
- A disposable local full-check fallback passed install, build, static checks,
  formatting, and lint. Its full coverage suite failed only in unrelated
  terminal-proof infrastructure: repeated tmux pane-cleanup timeouts and one
  synthetic profile test observing the host home directory. This PR does not
  claim a green full `pnpm run check`.

## Real Behavior Proof

**Claim:** post-effect HTTP 500 and network failures issue exactly one request,
while heartbeat HTTP 500 still retries with byte-identical signed payloads.

**Exercised surface:** the built production
`ExactReviewBatchQueueClient` over a loopback HTTP server.

**Scenario:** the loopback server returned HTTP 500 for a router receipt and
HTTP 500 then 200 for a heartbeat; a separate production-client request threw a
network error.

**Command/environment:** Node 26.7.0, built `dist` client, loopback HTTP transport,
synthetic credentials and identifiers only.

**Observed result:**

```json
{
  "headSha": "288389c13afc1e2d8079cbf7a2a0362f9a82037c",
  "postEffectHttp500Attempts": 1,
  "postEffectNetworkAttempts": 1,
  "heartbeatAttempts": 2,
  "heartbeatSignedBytesIdentical": true
}
```

**Artifact/trace:** `.artifacts/post-effect-containment-proof.json` in the local
proof workspace; the redacted result is reproduced above.

**Limits:** this proves the production client transport boundary, not a
Cloudflare edge, live queue, or end-to-end Worker release/recovery cycle. The
offline ClawSweeper review requested a Worker injected-fault recovery trace.
That rank-up move is not performed here because this containment scope
explicitly excludes Worker/dashboard changes and live queue mutation. The Bay
non-impact disposition is recorded above.

### DIFF
--- a/docs/live-dashboard.md
+++ b/docs/live-dashboard.md
@@ -471,14 +471,17 @@ counts but are not serialized by the public projector. Document effective
 production values from `dashboard/wrangler.toml`, not only fallback constants
 in `dashboard/exact-review-queue.ts`.
 
-Batch publication heartbeats and post-effect enqueue, router-receipt, and
-terminal-disposition POSTs retry network errors, timeouts, and HTTP 5xx responses
-(including `exact_review_queue_unavailable`) up to three attempts within 45
-seconds, with at most 20 seconds per attempt. Retries preserve the signed bytes,
-use jittered exponential backoff, and honor `Retry-After` up to 10 seconds;
-heartbeats additionally stop at the last confirmed lease expiry. Validation,
-authentication, and fence rejections are never retried, and receipt retries do
-not repeat the GitHub router dispatch.
+Batch publication heartbeats retry network errors, timeouts, and HTTP 5xx
+responses (including `exact_review_queue_unavailable`) up to three attempts
+within 45 seconds, with at most 20 seconds per attempt. Heartbeat retries
+preserve the signed bytes, use jittered exponential backoff, honor `Retry-After`
+up to 10 seconds, and stop at the last confirmed server lease expiry.
+Validation, authentication, and fence rejections are never retried.
+
+Post-effect enqueue, router-receipt, and terminal-disposition POSTs are
+deliberately one-shot until Worker replay ordering is repaired. An ambiguous
+failure ends the publication run so scheduled recovery can replay it without an
+in-process retry arriving after newer queue state.
 
 Lifecycle receipt replays are no-ops for the whole operation, including terminal
 transitions and acknowledgement drivers. Terminal-disposition requests from the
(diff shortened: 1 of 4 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
Maintainer decision: accept the temporary availability tradeoff. Ordering safety
outweighs same-run post-effect retry latency while the Worker replay-ordering
P1s remain open. Unfinished members retain scheduled recovery, and heartbeat
retries remain bounded. The optional full Worker injected-fault trace is
deferred to the separate Worker repair; it is not a blocker for this containment
PR.

PR #1368, opened 2026-09-03
fix(publish): retry transient queue failures during batch publication

steipete:
## What Problem This Solves

Batch publication could stop on a transient queue failure after its GitHub or canonical-record effects succeeded. Heartbeats and post-effect POSTs now retry bounded transient failures with unchanged signed bytes and the last confirmed lease expiry.

Lifecycle retries also need to distinguish a completed operation from a partially written legacy receipt. A lost response must not restore an old terminal state over a newer requeue, and reusing an operation ID for a different terminal kind must fail explicitly.

## What Changed

- Router receipts carry `operationComplete: true` only through the atomic handler path. Completed replays are durable no-ops. Legacy receipts without the marker repair missing terminal/driver effects; if a different terminal disposition is at least as recent as the receipt, the handler marks completion without overwriting that disposition or changing the queue.
- Terminal operation history stores `{ operationId, kind }`. Same-ID/same-kind replay is a no-op; a different kind returns HTTP 409 `conflicting_lifecycle_terminal_operation`, and the store rejects it too. Existing string IDs migrate on read to `{ operationId, kind: null }`, deliberately retaining wildcard dedupe compatibility for that unreleased field.
- Each projection retains the newest 32 terminal operations in FIFO order; duplicate IDs are not appended. IDs are per workflow run, attempt, and fence.
- Receipt completion, terminal transitions, and acknowledgement-driver writes remain one SQLite transaction. Failure rolls back every durable effect.
- Client retries remain limited to three attempts within 45 seconds, with 20-second attempt limits, jittered backoff, bounded `Retry-After`, and no heartbeat past its last confirmed lease. HTTP 4xx failures remain non-retryable.
- Client failure messages retain the HTTP status and only a validated JSON server `error` code from at most 512 response bytes. Raw bodies, URLs, and payloads are not included.

OpenClaw Bay needs no frontend or public-schema change. It consumes the same terminal projection, preserving the authoritative requeue; completion markers and terminal operation history remain internal. Release-note context is maintained here; this fix round changes only the two lifecycle source files and their existing publication-lifecycle test file.

## Validation and finding disposition

[...]

PR #1375, opened 2026-09-03
fix(queue): preserve lifecycle state across replay failures

vincentkoc:
Related: https://github.com/openclaw/clawsweeper/pull/1368

## What Problem This Solves

Fixes an issue where a lost Worker response could replay a committed lifecycle operation, mutate canonical state twice, or fail to restore derived Bay and alarm state after the canonical transaction had already committed.

This is the server prerequisite only. It does not re-enable client-side post-effect retries.

## Why This Change Was Made

Router receipts and terminal dispositions now use durable operation identity inside the lifecycle projection transaction. Same-identity replays skip canonical mutation but always rerun idempotent Bay materialization and scheduling; conflicting terminal operation kinds return HTTP 409.

The router endpoint preserves the deployed f7 ordering rule for a matching receipt followed by a newer or equal non-routed terminal, and additionally preserves any authoritative requeue when a delayed router request had not committed a receipt. This endpoint rule is intentionally narrower than the lifecycle store contract: `applyTerminalDisposition` still permits requeue-to-route transitions for other callers, as covered by queue-policy tests.

Terminal operation IDs remain for the projection lifetime with no eviction. Old rows without the field load as an empty ledger; malformed or duplicate entries are rejected.

Peter Steinberger authored the original publication-retry work and is credited in the signed commit. This prerequisite is separated because the server behavior must differ from the source PR before post-effect retries can safely return.

## User Impact

Operators get deterministic lifecycle recovery after ambiguous Worker failures: duplicate delivery cannot recreate canonical effects, overwrite a later requeue, or leave Bay/alarm projections stale.

There is no new public API or configuration surface.

## OpenClaw Bay Impact

The public Bay projection and schema are unchanged. Duplicate router and terminal deliveries rerun the existing idempotent `syncBayLifecycle` path after the canonical transaction, so a post-commit materialization failure can recover without changing lifecycle facts.

## Documentation Impact

Reviewed `docs/live-dashboard.md`. No documentation change is needed: post-effect requests remain deliberately one-shot, and this PR only repairs the Worker prerequisite needed before that availability tradeoff can be reconsidered.

## Evidence

- Signed head: `ba3822b6cf179d98bcaf2444e6974deb6095e752`
- Base: `674841bcdc0a826ff4555f945cc81f122db0d6c2`
- Scope: three files only.
- Production: `+98/-80`, net `+18`.
- Tests: `+238/-87`, net `+151`.
- Focused proof: 436 passed, 0 failed:
  - publication lifecycle: 79
  - Bay routes: 115
  - queue runtime: 187
  - queue policy: 55
- `CI=1 pnpm run check:static`: passed.
- `pnpm run build:all`: passed.
- `pnpm run lint`: passed.
- `pnpm run format`: passed on 808 files.
- `git diff --check`: passed.

Regression coverage includes:

- all eight non-routed terminal kinds with matching newer/equal receipt timestamps remain unchanged;
- terminal older than its matching receipt returns 409 atomically;
- receiptless non-requeue returns 409 without recording a receipt;
- receiptless requeue returns 200 and remains authoritative without a driver;
- post-commit derived failure replay reruns Bay and scheduling without canonical mutation;
- terminal same-ID/same-kind replay is a no-op; same-ID/different-kind returns 409;
- more than 32 terminal IDs remain deduplicated for the projection lifetime;
- malformed and duplicate ledgers are rejected, while a row with no ledger loads;
- transaction rollback and the existing direct-command acknowledgement finalizer coverage remain intact.

### Review precedence

Two local review suggestions were considered and rejected after direct source/history audit:

1. A distinct router receipt does not automatically override an existing requeue at this delayed-router endpoint. That would reintroduce the failed-before-commit ordering bug. Other store callers retain the intentional requeue-to-route transition.
2. String terminal IDs and `{ kind: null }` entries are not deployed durable states. Before f7 the field did not exist; f7 introduced the field and writes only concrete `{ operationId, kind }` entries. The removed reader bridge covered an unmerged predecessor shape, not a shipped migration.

## Real Behavior Proof

**Claim:** the real Worker HTTP boundary preserves a requeue across delayed router delivery, makes completed router and terminal replays canonical no-ops, and rejects a conflicting terminal operation kind.

**Surface:** Wrangler 4.107 running the production Worker and SQLite Durable Object locally over HTTP loopback. GitHub App installation and repository metadata were served by an ephemeral local mock; no private endpoint, credential, or repository data was retained.

**Command:**

```sh
bash /tmp/clawsweeper-worker-replay-proof.sh "$PWD" 8796
```

**Observed result:**

```json
{"ok":true,"transport":"HTTP loopback","runtime":"Wrangler Worker with SQLite Durable Object","router":{"delayed_receipt_preserved_requeue":true,"duplicate_left_updated_at_unchanged":true},"terminal":{"duplicate_left_updated_at_unchanged":true,"conflicting_kind_status":409}}
```

**Limits:** this proves the production Worker/SQLite/HTTP boundary locally. It does not claim deployment or a live queue mutation. Docker was unavailable. Two AWS Crabbox attempts failed before running repository commands: one because local Actions hydration did not support `pnpm/action-setup@v6.0.9`, and one on broker DNS/EOF plus cancel-create timeout. A local full `pnpm run check` fallback reached unrelated terminal proof/PTY infrastructure failures; the focused owner-boundary suites and all static/build/lint/format gates above are green.

No deployment, queue dispatch, cancellation, retry, or other live operation was performed.

### DIFF
--- a/dashboard/exact-review-lifecycle.ts
+++ b/dashboard/exact-review-lifecycle.ts
@@ -6,8 +6,6 @@ export const EXACT_REVIEW_LIFECYCLE_AUDIT_READ_LIMIT = 10_000;
 export const EXACT_REVIEW_LIFECYCLE_AUDIT_PAGE_MAX = 100;
 export const EXACT_REVIEW_LIFECYCLE_AUDIT_SNAPSHOT_TTL_MS = 5 * 60 * 1000;
 export const EXACT_REVIEW_LIFECYCLE_AUDIT_SNAPSHOT_MAX_ACTIVE = 4;
-// IDs are per run/attempt/fence, so 32 covers any realistic replay window.
-const EXACT_REVIEW_LIFECYCLE_TERMINAL_OPERATION_LIMIT = 32;
 const EXACT_REVIEW_LIFECYCLE_AUDIT_SNAPSHOT_TABLE = "exact_review_lifecycle_audit_snapshots_v1";
 const EXACT_REVIEW_LIFECYCLE_AUDIT_SNAPSHOT_ROW_TABLE =
   "exact_review_lifecycle_audit_snapshot_rows_v1";
@@ -225,7 +223,7 @@ export type ExactReviewLifecycleProjection = {
    */
   terminalDispositions: Array<{ kind: LifecycleTerminalDisposition; observedAt: number }>;
   /** Applied terminal POST identities, retained with the revision's receipt history. */
-  terminalOperationIds: Array<{ operationId: string; kind: LifecycleTerminalDisposition | null }>;
+  terminalOperationIds: Array<{ operationId: string; kind: LifecycleTerminalDisposition }>;
   /** Current terminal outcome used to derive lifecycle and acknowledgement state. */
   terminalDisposition: { kind: LifecycleTerminalDisposition; observedAt: number } | null;
   /**
@@ -524,7 +522,7 @@ export class ExactReviewLifecycleProjectionStore {
       operationComplete?: true;
     },
   ) {
-    return this.storage.transactionSync(() => this.recordRouterReceiptSync(input));
+    return this.storage.transactionSync(() => this.recordRouterReceiptSync(input).projection);
   }
 
   recordRouterReceiptSync(
@@ -537,7 +535,7 @@ export class ExactReviewLifecycleProjectionStore {
   ) {
     this.validateIdentity(input);
     if (!validText(input.receiptId, 1, 300)) throw new Error("invalid lifecycle router receipt");
-    return this.mutateSync(input, (projection) => {
+    return this.mutateIdempotentlySync(input, (projection) => {
       const next = {
         outcome: input.outcome,
         receiptId: input.receiptId,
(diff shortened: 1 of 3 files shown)


vincentkoc:
@clawsweeper re-review

Please review exact head `ba3822b6cf179d98bcaf2444e6974deb6095e752`. The PR body contains the current-main proof, real Worker/SQLite transport trace, and explicit maintainer precedence rationale for the endpoint requeue fence and deployed ledger shape.

vincentkoc:
Maintainer disposition for the remaining rank-up move: Docker-backed Crabbox proof is deferred because Docker is unavailable on this host and both recorded AWS attempts failed before repository execution (Actions hydration incompatibility, then broker DNS/EOF). This is an optional rank-up, not a code finding; the exact-head review found no actionable defect and states real-behavior proof is not required for this maintainer-authored PR. The PR body retains the redacted Wrangler Worker + SQLite HTTP proof and the precise proof limits.

vincentkoc:
Late compatibility finding resolved by deployment/source provenance: pre-`f7c9afb15584b1837893877fdb3eca64f321b8bd` rows had no `terminalOperationIds` field and load as `[]`; deployed `f7c9afb` only wrote concrete `{ operationId, kind }` entries. Dashboard deploy history contains `f7c9afb` (run 33712345842) and this PR's merge `69ec53b00f118593cf878d21d2975c8a52a84e3c` (run 33762545402), not intermediate `ce24cc9`/`29f171d3` heads. No deployed string/null ledger needs compatibility, so strict concrete-kind conflict detection remains correct; follow-up #1376 is closed unmerged.

PR #1389, opened 2026-09-04
fix(queue): recover publication claims after transient failures

vincentkoc:
## What Problem This Solves

Fixes an issue where a dispatched publication workflow could lose a claim response during a transient Worker or network failure, exit without a lease, and leave already-reserved publication work waiting for later recovery.

## Why This Change Was Made

Dispatched claim requests now retry transient network, timeout, and HTTP 5xx failures up to three attempts within the existing 45-second request budget. Retries reuse the exact serialized body and signature.

The Worker rechecks for an already-created lease after asynchronous target admission and replays the complete existing-batch admission path. The batch store remains the single transactional owner, so the same batch ID and lease owner can create at most one lease.

Undispatched claims, completion requests, post-effects, workflow behavior, capacity, batch size, schema, and configuration remain unchanged.

## User Impact

Maintainers should see fewer failed publication runs and less delayed backlog recovery when a claim succeeded server-side but its response was lost.

## OpenClaw Bay Impact

Bay is unaffected. This changes only the signed internal claim transport and existing-batch replay path; public projections, fields, and observer behavior are unchanged.

## Documentation Impact

Reviewed `docs/live-dashboard.md`. No documentation change is required because the public queue contract, operator controls, configuration, and projection fields are unchanged. The retry is bounded internal transport recovery for dispatched claims.

## Evidence

- Signed head: `3a5ba4d056590aab12747d7be03c8c81559bef97`
- Base: `72041f31e1d34cff19f50da4b23c4a22a78c828a`
- Scope: five files.
- Production: `+56/-43`, net `+13`.
- Tests: `+237/-0`.
- `build`, `build:repair`, and `build:dashboard`: passed.
- Focused suites: 306 passed, 0 failed:
  - `test/repair/exact-review-batch-queue-client.test.ts`
  - `test/exact-review-publication-batches.test.ts`
  - `test/dashboard-worker-queue-runtime.test.ts`
- Targeted `oxlint`: passed.
- Targeted `oxfmt --check`: passed.
- `check-active-surface`, `check-dashboard-queue-boundary`, and `check-dashboard-strict`: passed.
- `git diff --check`: passed.

Regression coverage includes:

- a dispatched claim retries timeout and HTTP 500, then returns the lease with identical body and HMAC on every attempt;
- HTTP 4xx claims remain single-attempt;
- undispatched claims, completion, and post-effects remain single-attempt;
- concurrent same-identity claims replay the first committed lease after asynchronous admission clears the reservation;
- replay responses retain the original lease identity while allowing current main's per-response `server_time` to advance;
- batch-store same-identity replay returns the original lease without extending or replacing it.

The broad local `pnpm run check` was not rerun. A previous attempt on this worktree hung in unrelated test infrastructure after about 11 minutes; this branch instead uses the focused owner-boundary suites and targeted static gates above. Hosted CI remains required.

## Real Behavior Proof

**Claim:** the built queue client retries a transient dispatched-claim HTTP 500 while preserving the exact request identity, then accepts the returned lease.

**Surface:** Node 26.7.0 running `dist/repair/exact-review-batch-queue-client.js` against an ephemeral local HTTPS server. The server captured each received body and signature and returned HTTP 500 before a successful lease response.

**Command:** an inline Node ESM proof imported the built client, started an ephemeral loopback HTTPS server with a synthetic key and webhook secret, issued one dispatched claim, and asserted the received bytes, HMAC, attempt count, and returned batch identity.

**Observed result:**

```json
{"result":"pass","attempts":2,"identical_body":true,"identical_signature":true,"claimed_batch":"proof-claim","server_time_preserved":true}
```

**Limits:** this proves the built client and real HTTPS request boundary locally, including current main's internal `server_time` response field. The Worker concurrency and SQLite lease replay are covered by the focused production-queue harness tests; this draft does not claim deployment or a live queue mutation.

No deployment, queue dispatch, cancellation, retry, capacity change, or other live operation was performed.

### DIFF
--- a/dashboard/exact-review-queue.ts
+++ b/dashboard/exact-review-queue.ts
@@ -7030,8 +7030,31 @@ export class ExactReviewQueue {
     // visibility can change while the lease is active. Re-probe each unfinished
     // repository before returning any credential-bearing work.
     let existingBatch = this.batchStore.fetch(claimId, leaseOwner, now);
-    if (existingBatch?.state === "leased") {
-      const unfinished = existingBatch.items.filter((item) => item.terminalOutcome === null);
+    const claimedResponse = async (
+      batch: NonNullable<typeof existingBatch>,
+      lifecycleBatch = batch,
+    ) => {
+      if (batch.items.length && lifecycleBatch.state === "leased") {
+        this.recordLifecycleBatchClaim(lifecycleBatch, state, now);
+      }
+      await this.scheduleNext(state, now);
+      const oldestCandidateAt = batch.items.reduce(
+        (oldest, membership) =>
+          Math.min(oldest, state.items[membership.itemKey]?.createdAt ?? batch.createdAt),
+        batch.createdAt,
+      );
+      return json({
+        ok: true,
+        claimed: true,
+        batch: exactReviewPublicationBatchJson(batch),
+        configured_batch_size: batch.configuredBatchSize,
+        batch_wait_ms: Math.max(0, now - oldestCandidateAt),
+        requested_max_items: requestedSize,
+        effective_max_items: leaseSize,
+      });
+    };
+    const replayExistingBatch = async (batch: NonNullable<typeof existingBatch>) => {
+      const unfinished = batch.items.filter((item) => item.terminalOutcome === null);
       const unfinishedRepos = new Map<string, string>(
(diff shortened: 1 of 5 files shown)


vincentkoc:
@clawsweeper re-review

PR #1447, opened 2026-09-05
fix(repair): accept compressed and empty cluster snapshots

steipete:
## What Problem This Solves

Cluster repair intake requires a raw SQLite file even when the store exports only a gzip archive. After materialization, an empty portable cluster schema is incorrectly treated as legacy, causing a query against a nonexistent `clusters` table. The missing-file failure is visible in https://github.com/openclaw/clawsweeper/actions/runs/33955163748.

## Why This Change Was Made

Copy the store's portable-artifact materializer into an owner-only temporary directory and verify a reviewed SHA-256 before executing it, after the processed-SHA check. Pass the materialized temporary database to the importer. Detect portable cluster schema by table presence while preserving nonempty legacy precedence. Compression handling stays with the store; ClawSweeper keeps import and deduplication policy.

## Review Disposition

The fail-closed digest dependency is deliberate and accepted in this implementation: snapshot publication does not authorize replacing executable code on the intake runner. A helper update requires a reviewed checksum update in this workflow; silently executing new bytes would undo that boundary. The repair documentation states this coordination requirement.

The pinned helper was inspected locally, and the actual helper bytes were used in the runtime proof. Its compressed branch checks archive basename, compressed size and SHA-256, decode budget, and decoded size and SHA-256. This PR does not claim stronger verification for the raw-file branch. The passing raw/gzip, hash-mismatch, and unreviewed-helper cases exercise the integration contract without depending on the reviewer having network access to the upstream checkout. No helper source is copied into this repository.

The current branch includes the publication cleanup already on main and preserves its human contributor credit in the changelog. The intake source and workflow hashes are unchanged after that merge; the clean-head nine-case proof was rerun, and the fresh committed-branch review reports no P0–P2 findings.

## User Impact

Raw and compressed snapshots can reach candidate import. Empty portable snapshots finish successfully with zero candidates. Repeated snapshots skip before materialization, forced reimports still work, and gzip archive/decoded hash mismatches stop before import admission.

## Evidence

- Focused SQLite, store, and intake-publication tests: 16 passed.
- Full `pnpm run check` on AWS through Crabbox: 4,986 tests, 4,968 passed, 18 skipped, zero failures; static checks, lint, builds and coverage gates passed.
- Independent Codex review before commit: no P0–P2 findings. The committed review identified execution of mutable publisher-supplied code; the digest pin, private copy, and passing tampering case resolve that finding.
- Changelog and repair documentation updated.

## Real Behavior Proof

**Claim and surface:** execute the actual workflow `Prepare intake` shell, built importer, and empty selector against synthetic SQLite snapshots. The proof does not substitute a decoder for the upstream materializer.

**Environment:** AWS through Crabbox, lease `cbx_76a9ed6493af`, image `ami-0461d919be7deb53c` in `eu-west-1`, Node `v24.18.1`, pnpm `11.10.0`. The hydration adapter did not support the pinned pnpm action, so the same lease used a frozen dependency install directly. No provider switch was made.

**Current source:** `7e9406ca78ad3ed1f4e690db6f83d65b9d3621c2`, clean tree. Current proof run: `run_e415ed055b8d`; full gate: `run_7f37c2f9d8f1`.

**Command:** after `pnpm run build:node`, run `node docs/proof/cluster-intake-portable/run-proof.mjs /tmp/portable-artifact.mjs .artifacts/cluster-intake-portable/candidate-synced.json`. Supply the store's existing materializer at the named temporary path. The same executable accepts a baseline ref as its third argument.

**Observed result:** all nine candidate cases pass: raw and gzip snapshots each produce one job; empty raw/gzip portable schemas produce no jobs and an empty selector decision; unreviewed materializer code is rejected before execution; archive and decoded hash mismatches reject admission; a processed SHA skips before decoding; forced reimport produces one job. Baseline `a9ed9b5ba7eb12357da7cc2360d87cc5397c3c36` accepts raw input, rejects gzip-only input, and fails on an empty portable schema with a legacy-table error.

**Artifact/trace:** reproducible harness and receipt format are in `docs/proof/cluster-intake-portable/`. Generated JSON records source/compiled hashes, dirty state, materializer hash, runtime and per-case outcomes. Candidate workflow SHA-256: `c42316d8a1214f5b7060c4ab3a43458a1a99afe99adb0e8393ee01e700ca6344`; importer source SHA-256: `0f694d4e80fc0bca10beb339451ba0a7866c828189c4619e310336908000d2c8`.

**Limits:** synthetic snapshots; no inference, live GitHub publication, or production intake was triggered. The baseline importer shares unchanged compiled dependencies. This does not add clustering to exports that omit it or strengthen the upstream helper's raw-file verification contract.

**OpenClaw Bay:** unaffected; this changes internal snapshot import behavior, with no public status or data-contract change.

### DIFF
--- a/.github/workflows/repair-cluster-intake.yml
+++ b/.github/workflows/repair-cluster-intake.yml
@@ -217,17 +217,11 @@ jobs:
           db_name="${TARGET_REPO,,}"
           db_name="${db_name//\//__}.sync.db"
           manifest_path="gitcrawl-store/data/${db_name}.manifest.json"
-          db_path="gitcrawl-store/data/${db_name}"
           ledger_path="results/cluster-repair-intake/${repo_slug}.json"
           if [ ! -f "$manifest_path" ]; then
             echo "::error::Missing gitcrawl-store manifest: ${manifest_path}"
             exit 1
           fi
-          if [ ! -f "$db_path" ]; then
-            echo "::error::Missing gitcrawl-store DB: ${db_path}"
-            exit 1
-          fi
-
           store_sha="$(node -e "const fs=require('fs'); const m=JSON.parse(fs.readFileSync(process.argv[1], 'utf8')); console.log(m.sha256 || '')" "$manifest_path")"
           exported_at="$(node -e "const fs=require('fs'); const m=JSON.parse(fs.readFileSync(process.argv[1], 'utf8')); console.log(m.exportedAt || '')" "$manifest_path")"
           previous_sha=""
@@ -238,7 +232,6 @@ jobs:
           {
             echo "target_repo=${TARGET_REPO}"
             echo "repo_slug=${repo_slug}"
-            echo "db_path=${db_path}"
             echo "manifest_path=${manifest_path}"
             echo "ledger_path=${ledger_path}"
             echo "store_sha=${store_sha}"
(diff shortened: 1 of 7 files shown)


steipete:
@clawsweeper re-review

The main PR body now contains the clean current-head proof and disposition of both review concerns: the helper checksum update dependency is deliberately accepted, and the pinned helper was inspected and exercised directly across successful, corrupted, empty, and tampered inputs. The digest gate remains intact.

PR #1452, opened 2026-09-05
fix: expose publication successor fence diagnostics

vincentkoc:
## What Problem This Solves

Resolves a problem where the authenticated publication-reconciliation response could report a stale command row as `terminal_missing` without identifying which successor uniqueness or lifecycle-admission fence prevented terminal disposition.

## Why This Change Was Made

The exact-review queue now derives one nullable `successor_fence_state` from the same successor and lifecycle-admission predicates used by the alarm path. The field is diagnostic only: `verified` does not grant mutation permission, and no supersede, terminal-disposition, or acknowledgement policy changes.

This is the Worker-first half of a staged rollout. The existing maintenance client accepts the additive Worker field without throwing but intentionally does not expose it yet. A follow-up will add the strict client type/parser and enum/coherence tests only after this Worker contract is deployed, avoiding a new strict client running against an older Worker.

Root cause: the queue already owned the complete successor fence, but the authenticated reconcile response exposed only the aggregate `terminal_missing` result.

Architectural owner: `ExactReviewQueue` remains the uniqueness, admission, and mutation-policy owner.

Canonical fix: one private successor-fence helper is shared by alarm evaluation and reconciliation sampling. No scan, schema, workflow, endpoint, public projection, or client change is included.

Production TypeScript delta: +18 net lines, justified by the new authenticated diagnostic contract and shared owner-boundary helper. Tests are counted separately.

## User Impact

The authenticated Worker response can distinguish a missing or ambiguous successor from each lifecycle-admission mismatch. The existing maintenance client remains compatible during deployment and will expose the field in the staged follow-up.

## OpenClaw Bay Impact

No Bay change. The new value exists only in the authenticated internal reconciliation response; it is not added to any public Bay projection or browser surface.

## Documentation Impact

Updated the active `docs/live-dashboard.md` runbook. Role owner: ClawSweeper maintainers. Source of truth: the authenticated Worker reconcile response at `6b896bf88912f4135d9d61eea1236c6d1f0e3c44`. Update triggers: changes to reconciliation sample fields, successor-fence predicates, or authentication/visibility of the route.

## Evidence

- `pnpm run build:all`: passed on Node 24.
- `node --test test/dashboard-worker-publication-lifecycle.test.ts test/dashboard-worker-queue-runtime.test.ts test/exact-review-publication-batches.test.ts test/repair/exact-review-batch-queue-client.test.ts test/repair/exact-review-batch-coordinator.test.ts`: 426/426 passed.
- `pnpm run check`: passed from a git-backed PR checkout on Crabbox AWS Linux, Node 24.18.1, run `run_aded5cabb9a5`; 5,103 tests, 5,085 passed, 18 skipped, 0 failed. A prior raw-rsync run was discarded because that workspace omitted `.git`, causing two unrelated E2E fixtures that call `git rev-parse HEAD` to fail.
- `git diff --check`: passed.
- Pre-commit and committed-head Codex reviews confirmed that the classifier preserves the existing alarm fence. Both raised the expected staged-rollout limitation that the current maintenance client drops the additive diagnostic. Disposition: deferred to the strict-client follow-up after this Worker deploys; adding it here would recreate the old-Worker/new-client incompatibility that required the split.
- Signed commit: `6b896bf88912f4135d9d61eea1236c6d1f0e3c44`, parented on deployed dry-run repair `50487d1848196e17e8b74e3dd2e7921b1a201588`.

## Real Behavior Proof

**Claim**

The actual signed production reconciliation handler reports the exact successor-fence diagnostic without mutating the tested queue state, and the unchanged production client accepts the additive field.

**Surface**

`POST /internal/exact-review/publications/reconcile` through the production Worker route and production `ExactReviewQueue`, backed by disposable workerd SQLite.

**Scenario**

Two synthetic publication rows share a target: the oldest row is stale with `acknowledgement_unavailable_reason=terminal_missing`; its unique successor has an exact matching lifecycle-admission record.

**Invocation**

- authenticated HMAC request
- `apply=false`
- `max_items=1`
- one request, no retry

**Environment**

- queue source SHA-256: `d87de8a66d8466b3e9ae2425b329ea86ee98f2fcbbf886f6e84d077cfb0b586d`, matching the committed file
- unchanged client source SHA-256: `eaacb158c94c5ecbf3be3b874b36c9a417bab0c8eb15c5519feb05243b254d4a`
- Node 24.16.0
- Miniflare 4.20260701.0
- workerd 1.20260701.1
- report SHA-256: `baf72a479229ed359f8ecaea7d3c10e2b976db50bba100896592c1f94055ef6f`

**Observed**

```text
reason=stale_revision
acknowledgement_unavailable_reason=terminal_missing
successor_fence_state=verified
changed=0
item_count_after=2
terminal_dispositions_after=0
external_calls=0
unchanged_client_sample_count=1
unchanged_client_acknowledgement_unavailable_reason=terminal_missing
```

**Limits**

The proof uses synthetic local rows and disposable workerd SQLite. It does not call GitHub, the deployed Worker, the production queue, or a workflow, and it performs no external mutation.

The independent `apply=false` safety repair in https://github.com/openclaw/clawsweeper/pull/1459 is merged and deployed at `50487d1848196e17e8b74e3dd2e7921b1a201588`. No live reconcile request is dispatched from this PR before the successor diagnostic Worker itself deploys; the post-deploy redacted one-shot remains a separate rollout gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ checkpoint, and status-only commits are intentionally omitted.
 - Reduce the review ceiling to 32, cap scheduled background reviews at eight, pace scheduled intake at 60 per hour with a six-item burst, and scan normal backlog hourly while retaining manual-review priority and existing leases.
 - Honor the configured Codex reasoning effort in sweep planning, event and shard reviews, assist answers, and close-coverage proofs instead of forcing high reasoning; assist uses the shared fast service tier.
 
+- Authenticated publication reconciliation now diagnoses missing, ambiguous, and mismatched successor fences without changing supersede behavior.
 - Publication reconciliation diagnostics now report the exact acknowledgement-unavailable reason while preserving fail-closed supersede behavior.
 
 - Admit the reviewed mocked marketplace telemetry-redaction and Gateway config CDP-redaction fixtures through exact URI, source-line, path, and decoder bindings without relaxing native scanner checks.
--- a/dashboard/exact-review-queue.ts
+++ b/dashboard/exact-review-queue.ts
@@ -6796,6 +6796,25 @@ export class ExactReviewQueue {
     return "routed_receipts_incomplete";
   }
 
+  private publicationSuccessorFenceState(successor: ExactReviewQueueItem | null | undefined) {
+    if (successor === undefined) return "missing" as const;
+    if (successor === null) return "ambiguous" as const;
+    const producer = successor.decision.publication!.producerDecision;
+    const admission = this.lifecycleProjectionStore.read(
+      `${producer.targetRepo}#${producer.itemNumber}`,
+      successor.key,
+      successor.revision,
+    )?.admission;
+    if (!admission) return "admission_missing" as const;
+    if (admission.commandOriginated !== exactReviewDecisionHasCommandContext(producer))
+      return "admission_command_mismatch" as const;
+    if (admission.statusMarker !== (producer.commandStatusMarker ?? null))
+      return "admission_marker_mismatch" as const;
+    if (admission.statusCommentId !== (producer.statusCommentId ?? null))
+      return "admission_comment_mismatch" as const;
+    return "verified" as const;
+  }
+
   private stalePublicationCandidatesSync(
     state: ExactReviewQueueState,
     activeBatchItemKeys: ReadonlySet<string>,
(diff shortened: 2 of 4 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

Issue #1471, opened 2026-09-07
Re-review run succeeded but durable review never updated; later review ran on stale/truncated data


TommyLei666:
### Re-review run reported "success" but never updated the durable review; the next review then ran on stale/truncated data

**Context:** I filed openclaw/openclaw#140010. On 2026-09-06T10:31:03Z I commented `@clawsweeper re-review` after adding evidence (issue comment 5558632346). ClawSweeper ACKed that the review was queued (comment 5558632692).

**What I observed (all links below):**

1. **First re-review run "succeeded" but its result never landed.**
   - Actions run https://github.com/openclaw/clawsweeper/actions/runs/34028984001 (created 2026-09-06T11:00:29Z) finished with `status=completed, conclusion=success` at 11:05:21Z.
   - Yet the durable review comment (https://github.com/openclaw/openclaw/issues/140010#issuecomment-5558487867) was NOT updated, and the ACK comment's progress block stayed at "State: Review in progress" (last updated 2026-09-06T11:01:56Z) forever.
   - Net effect: for ~14.5 hours the issue looked like it was being / had been re-reviewed, but no review result existed.

2. **A second request produced the review, but it ran on stale/truncated data.**
   - Second `@clawsweeper re-review` at 2026-09-07T01:02:16Z (comment 5563572174) -> run https://github.com/openclaw/clawsweeper/actions/runs/34071991807 (named "...terminal-finalization:openclaw/openclaw#140010:4 rev 4...") completed at 01:08:07Z; the durable review comment was finally updated at 01:07:17Z.
   - The completed review contains clearly stale/incorrect reads of the thread:
     - It states the supplied timing windows "are from 2026.9.1". They are dated 2026-09-05 and 2026-09-06; "2026.9.1" is a software version string in the issue body (Environment section).
     - Its details note "The truncated correction and unavailable September 7 comment leave a material discussion-coverage gap" - but the evidence comment is ~11 KB of plain Markdown, and the September 7 comment it says was unavailable (the re-review request, 5563572174, 01:02:16Z) predates the review's own timestamp (reviewed_at=2026-09-07T01:06:39Z) by ~4 minutes.
   - It re-asked for "expected vs actual behavior", which the evidence comment already contained in its first third.

**Expected:** a re-review run that concludes `success` updates the durable review comment promptly; a fresh review reads the current thread state (all comments present before the run started).

[...]

PR #1493, opened 2026-09-08
fix(review): preserve later discussion evidence

steipete:
## What Problem This Solves

A contributor's evidence comment could be present in the hydrated thread while its correction or proof was missing from the review prompt. Retained issue comments and inline PR comments were reduced to the first 6,000 characters without body-coverage metadata. This fixes the confirmed discussion-truncation part of https://github.com/openclaw/clawsweeper/issues/1471; the publication-delay investigation remains open.

## Why This Change Was Made

Reuse the existing bounded primary-body compactor for retained discussion comments. Comments through 12,000 UTF-16 units remain complete; larger comments carry the same bounded opening, later proof/output excerpts, source hash, and explicit omitted ranges as primary bodies. Inline-comment cache fingerprints include the complete source hash when coverage is incomplete, so edits in excerpts or omitted text cannot reuse the previous verdict. The existing 24 discussion / 40 inline-comment windows remain in place. Prompt guidance now describes comment coverage too.

## User Impact

Reviewers receive later corrections and evidence that previously disappeared after character 6,000. Remaining omissions are explicit, so incomplete reading cannot be mistaken for absent evidence. No publication, queue, scanner-admission, or mutation policy changes. OpenClaw Bay is unaffected because it does not consume review-input compaction.

Release note for the final notes PR: Preserve later evidence in retained discussion comments and expose incomplete coverage instead of silently truncating at 6,000 characters; thanks @TommyLei666.

## Real Behavior Proof

Claim: evidence beyond the former comment cutoff reaches the compiled hydration and final review-prompt path.

Environment and command: Node 24.20.0, pinned pnpm 11.10.0; `pnpm run build && node scripts/e2e/discussion-context.ts --live`.

The executable reads the original public evidence comment cited by issue 1471 through GitHub, then passes that response through the compiled context and prompt owners. All 10,827 characters reach both issue and PR prompts, preserving the final 4,827 characters that the previous cutoff omitted. The JSON receipt contains only lengths and hashes. Synthetic 6,590-character and 60,641-character cases prove late trace delivery and explicit incomplete coverage for the larger case. The old truncation operation loses the same synthetic trace.

The same executable drives the compiled scheduled-review cache decision for an unchanged long inline comment, an excerpt-only edit at offset 19,635, and an omitted-text edit at offset 60,640. The unchanged input hits the cache; both edits miss it while preserving the displayed prefix.

Limits: this proves input delivery and cache invalidation, not the model's interpretation, the secret scanner, the historical publication stall, or production deployment. It performs no GitHub writes and does not execute submitted evidence.

## Evidence

- `node --test test/primary-body.test.ts test/review-content-cache.test.ts`: 62 passed, including regression cases for the former cutoff, long-comment coverage, and both cache fingerprint paths.
- Full `pnpm run check`, current-head CI, and independent review results are recorded in the proof follow-up.
- No review findings are knowingly left unresolved. The change reuses the established coverage implementation rather than adding another excerpt algorithm.

## Review Finding Disposition

Accepted the P2 inline-comment cache-fingerprint finding. The shared digest owner now includes the host-generated full-source hash for comments with incomplete coverage. Regression tests cover excerpt-only and omitted-range edits through both hydrated revision fingerprints and fallback content fingerprints. The executable proof additionally checks the actual cache-hit decision. This also implements the corresponding rank-up suggestion.

### DIFF
--- a/docs/pr-review-comments.md
+++ b/docs/pr-review-comments.md
@@ -410,9 +410,13 @@ report representation; older unmarked attribution cannot regain verified status
 when comments are rendered. Stored reports and live comments are not rewritten by
 this reader change. OpenClaw Bay needs no change: no observer API or controls change.
 
-## Primary Body Coverage
+## Primary Body and Discussion Coverage
 
-Hosted primary issue and PR bodies up to 12,000 UTF-16 units remain intact.
+Hosted primary issue and PR bodies and retained discussion comments up to
+12,000 UTF-16 units remain intact. This includes inline PR review comments;
+the existing 24 discussion-comment and 40 inline-comment windows still apply.
+This replaces the former 6,000-unit discussion prefix, which could omit a
+contributor's later correction without reporting body coverage.
 Longer bodies retain an opening plus at most three source-ordered verbatim
 excerpts around proof and trace/output anchors, including inside details.
 The sibling `bodyCoverage` records the full-source SHA-256, original length,
@@ -438,7 +442,7 @@ retain curl's 90-second limit.
 Assist preserves coverage alongside the body. The report context ledger counts
 each primary record as one entry and includes its coverage in character totals;
 its list hydration counters do not describe body completeness. Related items,
-comments, patch content, local body overrides, proof statuses, and mutation gates
+patch content, local body overrides, proof statuses, and mutation gates
 are unchanged. This is reviewer input only: OpenClaw Bay needs no change because
 no observer API, public data contract, or action surface changes.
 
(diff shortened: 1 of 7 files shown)


steipete:
The inline-comment cache finding is addressed in `062175d618`: the shared fingerprint now includes the host-generated full-source hash when coverage is incomplete. Both hydrated and fallback cache paths have regression coverage.

Executed `node scripts/e2e/discussion-context.ts --live` on the updated head. The original 10,827-character evidence comment reaches both compiled prompts intact. Unchanged long inline evidence hits the scheduled-review cache; edits at offsets 19,635 and 60,640 miss it. The 62 focused compaction/cache tests pass. The main PR body now includes this proof and the finding disposition; the remaining publication-delay investigation stays open.

@clawsweeper re-review

steipete:
Validation is complete for `062175d618bb2f857b2d6698ee984da0aac79187`. The full CI workflow, including `pnpm check`, passed: https://github.com/openclaw/clawsweeper/actions/runs/34232289902.

The committed branch review is clean through P2, and the latest ClawSweeper review matches this head with sufficient proof and no actionable findings. Current native behavior proof and its limits are in the main PR body. This is prepared for maintainer landing; no merge was performed in this pass.

PR #1518, opened 2026-09-10
fix(review): hydrate blob sizes from bounded trees

RomneyDa:
## Summary

- hydrate historical review blob sizes from the existing bounded recursive-tree REST metadata route
- cache one complete metadata map per distinct base/head revision
- preserve fail-closed handling when a tree is truncated or a diff blob is absent

## Why

The private-workspace admission deployed in #1488 made exact OpenClaw reviews depend on per-object GraphQL lookups. Under the current review burst, those workers repeatedly failed before model execution with `review_checkout_unavailable` / “remote blob size metadata is unavailable.” The durable queue retried the same exact heads without producing artifacts.

Every blob in a base-to-head diff must belong to either the exact base tree or exact head tree. ClawSweeper already has a bounded recursive-tree resolver that rejects truncated responses, invalid entries, and repositories over the tracked-file limit, so source hydration can use that complete metadata instead of amplifying GraphQL requests per object batch.

## Behavior proof

- Claim: exact review source admission can resolve both sides of a real OpenClaw PR diff without per-object GraphQL metadata.
- Surface: `githubReviewTreeBlobSizes` through the revised hydration resolver.
- Scenario: OpenClaw Wave 13 completion PR base `46af4a9582b34444d11cc98310838060e1aae5ca` and head `4cc522493abf6cbacafc272251d832d52a2ed2a3`.
- Environment: local Node 24 process using authenticated live `gh api` REST reads.
- Result: both recursive trees were complete and untruncated, each yielded 39,880 unique blob identities, and the changed file’s base/head blobs resolved to 39,361 and 38,117 bytes.
- Limits: this proves live metadata acquisition and exact revision binding, not hosted workflow deployment; CI and a post-merge exact review provide that final proof.

## Validation

- `pnpm run build:all`
- `node --test test/review-blob-hydration.test.ts test/review-command-workflow.test.ts` — 122 passed
- focused live API behavior proof above
- committed-head autoreview P3 — clean, 0.91 confidence
- `pnpm run check` — 5,941 passed, 32 skipped; one unrelated macOS-only failure because system Bash lacks `mapfile` in `apply-drift-refresh` E2E. Linux CI is the platform-complete authority.

## OpenClaw Bay

Not affected. This changes private review source admission only; no queue, status, telemetry, or Bay data contract changes.

### DIFF
--- a/src/clawsweeper-context-hydration.ts
+++ b/src/clawsweeper-context-hydration.ts
@@ -14,7 +14,6 @@ import { createRelatedContext } from "./clawsweeper-related-context.js";
 import {
   ensurePullRequestReviewHead,
   ensureReviewTreeCommit,
-  githubReviewBlobSizes,
   githubReviewTreeBlobSizes,
   hydratePullRequestReviewBlobs,
   hydratePullRequestReviewHistory,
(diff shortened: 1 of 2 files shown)


RomneyDa:
@clawsweeper re-review

PR #1522, opened 2026-09-10
fix(review): project bounded tree metadata at source

RomneyDa:
## What this fixes

#1519 projected the Git Trees response used by final review-tree materialization, but exact-review source hydration has a separate recursive-tree call. Wave 13 still failed there before admission.

Apply #1519's validated projection to that remaining source-hydration call and share the projection constant between both paths. The projection preserves malformed entries for downstream rejection while removing unused path/URL fields from production-sized responses.

## Evidence

- `pnpm run build && node --test test/review-blob-hydration.test.ts` — 33 passed, including #1519's >8 MB capture regression and a new exact source-hydration argument assertion
- format, all builds, all lint, and changed-coverage stages — passed
- P3 exact-commit autoreview — clean, 0.99
- live OpenClaw tree: 10,692,481 bytes raw → 3,110,958 bytes projected; blob SHA/size inventory preserved

## Production historical-hydration proof

Ran the exact PR head `270b2471a67a9fcf8e978718546fbc745b9f1be6` locally on macOS against the real OpenClaw checkout and PR #143533:

```text
pnpm run review -- --local-only \
  --target-repo openclaw/openclaw \
  --item-number 143533 \
  --target-dir /Users/dallin/openclaw-wave13-completion \
  --output-retention summary \
  --result-format json \
  --codex-timeout-ms 120000
```

Observed completion:

- source/context hydration completed in 12,146 ms
- pinned base/merge-base `46af4a9582b34444d11cc98310838060e1aae5ca`
- pinned PR head `4cc522493abf6cbacafc272251d832d52a2ed2a3`
- fetched current-main comparison `1baeb74f01730d183e3500163f741ca235e2d6fc`
- review completed successfully in 1m 58s with `decision: keep_open`, `confidence: high`, and no findings
- retained summary artifact: `artifacts/local-review-143533/143533.md`
- checkout inspection verified; review status complete; terminal failure false

The run used the normal production source-preparation path and retained the existing admission limits: 200,000 files / 2 GiB projected and actual checkout usage, plus the 1 GiB filesystem reserve. No limit was raised or bypassed.

### DIFF
--- a/src/clawsweeper-context-hydration.ts
+++ b/src/clawsweeper-context-hydration.ts
@@ -45,6 +45,9 @@ import { isGitHubNotFoundError } from "./github-retry.js";
 import { type RepositoryProfile } from "./repository-profiles.js";
 import { compareCodeUnits, stableJson } from "./stable-json.js";
 
+const REVIEW_TREE_METADATA_JQ =
+  '{truncated, tree: (.tree | if type == "array" then map(if type == "object" then {type, sha, size} else . end) else . end)}';
+
 interface CreateContextHydrationDependencies {
   asRecord: (value: unknown) => Record<string, unknown>;
   CLAWSWEEPER_BOT_AUTHORS: Set<string>;
@@ -984,7 +987,7 @@ export function createContextHydration(dependencies: CreateContextHydrationDepen
         const sizes = githubReviewTreeBlobSizes({
           repository: targetRepo(),
           headSha: revision,
-          request: (path) => ghJson(["api", path]),
+          request: (path) => ghJson(["api", path, "--jq", REVIEW_TREE_METADATA_JQ]),
         });
         remoteTreeSizes.set(revision, sizes);
         return sizes;
@@ -1114,15 +1117,7 @@ export function createContextHydration(dependencies: CreateContextHydrationDepen
             headSha: options.headSha,
             // Path and URL metadata can exceed the CLI capture limit before admission runs.
             request: (path) =>
-              ghJsonOnce(
-                [
-                  "api",
-                  path,
-                  "--jq",
-                  '{truncated, tree: (.tree | if type == "array" then map(if type == "object" then {type, sha, size} else . end) else . end)}',
-                ],
-                timeoutMs,
-              ),
+              ghJsonOnce(["api", path, "--jq", REVIEW_TREE_METADATA_JQ], timeoutMs),
           });
           return new Map(
             objectIds.flatMap((objectId) => {
--- a/test/review-blob-hydration.test.ts
+++ b/test/review-blob-hydration.test.ts
@@ -1523,6 +1523,10 @@ test("source preparation reports unavailable historical blobs before restricted 
         targetRepo: () => "fixture/repository",
         ghJson: (args: string[]) => {
           assert.equal(args[0], "api");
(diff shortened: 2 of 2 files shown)


RomneyDa:
@clawsweeper re-review

RomneyDa:
@clawsweeper re-review

steipete:
Heads-up from ClawSweeper monitoring (UTC, 2026-09-10): openclaw/openclaw exact reviews are failing at roughly 60–70% since about 08:56 (67 failures vs 28 successes in one 70-minute window), while other target repositories (ocm, telemetry, CodexBar) review normally.

Failure-diagnostics manifests from ten recent openclaw failures:

- 4× `agent_input_scan / deadline`
- 3× `source_preparation / review_blobs_unavailable` (ETIMEDOUT, SIGTERM)
- 2× `source_preparation / review_commit_fetch_failed` (ETIMEDOUT, SIGTERM)
- 1× `source_preparation / source_incompatible`

Earlier windows today (02:00–08:55) show only `source_incompatible` at a low rate; the timeout classes appeared with the morning PR volume. From this machine a shallow fetch of an openclaw PR head takes one second and GitHub status is operational, so this looks like the recursive tree-metadata hydration for the large openclaw repository hitting `REVIEW_TREE_METADATA_DEADLINE_MS` (30 s) and the scan deadline under concurrency, rather than a GitHub outage. The affected PRs touch ordinary source files (no submodules or LFS).

Unrelated to this: the same morning had a separate ClawSweeper incident (#1525, rolled back and reverted via #1527); the openclaw timeout failures continued unchanged after that rollback, so they are attributed to the hydration path, not the queue.

Not reverting anything from this series myself; flagging so the owners can decide whether to bound or defer the tree hydration for large repositories. Evidence runs: 34461206619 (deadline), 34461206024 (fetch timeout), 34461188384 (source_incompatible).

steipete:
Follow-up data point: another burst at 11:36–11:40 UTC on the reverted head `fcaef5fe5`: 12 openclaw/openclaw exact reviews dispatched within about 4 minutes and failed, 13 of 97 completions in that window versus 1 of 83 in the window before. Sampled diagnostics are all `source_preparation / review_commit_fetch_failed` with `ETIMEDOUT` and `SIGTERM` (runs 34472195520, 34472279803). So the dominant remaining class is the PR commit fetch timing out when many openclaw reviews start concurrently, rather than the tree-metadata deadline alone. Between bursts the failure rate is near zero.

steipete:
Root-cause pointer for the fetch timeouts: `src/clawsweeper-review-blobs.ts` (around line 133) runs `git fetch --force --filter=blob:none --no-tags ... [--unshallow] origin <ref>` with a hard `timeout: 30_000`, and adds `--unshallow` whenever the target checkout is shallow. For openclaw/openclaw an unshallow fetch of the full history does not reliably finish in 30 s when a dozen runners start at once, which is exactly the `review_commit_fetch_failed` / `ETIMEDOUT` / `SIGTERM` signature in the 11:36 burst. Options for the owners: avoid `--unshallow` when the review only needs the PR head plus merge base (deepen to the merge base instead), or give the initial unshallow its own bounded budget separate from the incremental fetch. I am not changing this myself.

PR #1601, opened 2026-09-16
fix(scanner): unblock reviews of generated Git metadata

steipete:
## What Problem This Solves

Fixes review refusals when the native scanner mistakes generated Git blob IDs for credentials, including the dependency update in https://github.com/openclaw/openclaw/pull/149908.

## User Impact

Updates containing only these generated metadata findings can reach review. A conditional second native scan uses the existing deadline and staging limit.

The same-file combination of metadata findings and an already-reviewed unchanged URI fixture can still refuse when masked headers cannot satisfy the URI witness parser. Those cases were already blocked; URI-only reviews retain their existing single scan. This change does not expand that fixture policy.

## Why This Change Was Made

The scanner owner binds each eligible complete object ID to canonical raw-diff and patch fields, matching endpoints, paths, regular-file modes, and independently rehashed blobs. A matching value in any full blob, prompt, schema, additional input, or raw working bytes refuses classification.

These witnesses only request proof. HTML decoding can reconstruct the same value from source content and borrow a detector keyword from the patch filename, so a value found in a header does not prove where a decoded finding originated. The owner therefore scans a second complete patch copy with only proven metadata fields masked at their original lengths. Its actual masked bytes are retained; filenames and content are unchanged. Any remaining unclassified finding, verified result, metadata-proof request, native error, incomplete completion, or source drift blocks admission.

The complete primary patch, raw diff, full blobs, prompt, schema, and additional input remain scanned unchanged. There are no detector exclusions, new settings, decoder reimplementations, gate bypasses, or deployments.

## OpenClaw Bay Impact

Bay is unaffected: this changes scanner admission and host stderr notices, without changing dashboard data or action surfaces.

## Documentation Impact

Updated the active README safety contract, active reproducible proof guide, and Unreleased changelog. The scanner admission owner and pinned native implementation remain the source of truth. Requalify this proof when the scanner version, metadata witness, staging, or admission flow changes.

## Evidence

- Eleven focused owner/sibling tests pass, including mandatory supplemental execution, unchanged primary bytes, preserved patch content, verified-result refusal, and missing-completion/native-error refusal.
- Static/docs/format checks, all three TypeScript builds, and lint pass.
- Fresh precommit and committed-branch Codex reviews are scoped-clean through P2.
- The unchanged trusted scanner owner also admits this companion's exact committed source range with zero findings; the repair itself remains reviewable under current hosted policy.
- Exact-head [CI](https://github.com/openclaw/clawsweeper/actions/runs/35086232530) passed `pnpm check`, Windows launcher, network smoke, and sparse-build checks. [CodeQL](https://github.com/openclaw/clawsweeper/actions/runs/35086232475) passed. The full check succeeded on an unchanged-head retry after the first attempt passed all 6,253 active tests but hit an empty Node coverage-profile error during collection; no coverage gate was changed.
- Local full-check limit: 6,234 tests passed, two existing compiled-consumer loopback requests failed with `ECONNRESET`, and 36 skipped. Both consumers passed isolated reruns, and their fixed-profile requests do not invoke the scanner. The exact transport cause is unproved. A separate harness follow-up is to consume the unauthorized-claim response bodies in `scripts/e2e/command-proof-consumer-loopback.mjs`; this missing cleanup is not claimed as the cause of those resets.
- Production delta: +186 net lines. The existing blob witness is shared rather than duplicated; the added code owns explicit proof-required state, bounded native replay, and strict completion. Tests add 360 lines; the standalone native proof adds 128 lines.

## Real Behavior Proof

**Claim and surface:** the real `scanAgentInput` flow admits generated metadata without clearing matching decoded source content. No model is invoked by the proof.

**Environment:** macOS 27 arm64, Node 24.21.0, pnpm 12.4.1, checksum-qualified TruffleHog 3.97.4. The current native release still has the upstream metadata false-positive tracked in https://github.com/trufflesecurity/trufflehog/issues/3266.

**Command:**

```bash
pnpm run build:node
node scripts/setup-review-tools.mjs --timeout-ms 120000
node docs/proof/agent-input-scan-git-metadata/run-proof.mjs /path/to/proof.json \
  /path/to/isolated-openclaw \
  47d9a40df0afa279da03cc08e14031d05738d456 \
  b87e5b75ffcf4386e01789b512e53d73b94437da
```

**Observed results:** the unchanged trusted scanner owner refused the original complete source range with 16 metadata findings. The corrected owner admits that same range with two metadata notices after its supplemental scan.

| Native scenario | Result |
| --- | --- |
| Original complete dependency source range | Admitted |
| Generated metadata fixture | Admitted |
| Matching value split by HTML tags | Refused |
| Hexadecimal HTML entities | Refused |
| Decimal HTML entities | Refused |
| URL-encoded HTML attribute | Refused |
| Zero-width characters | Refused |
| Same value copied into additional input | Refused |

**Trace and limits:** the committed proof script writes the native classification/diagnostic JSON used for this matrix. Encoded fixture blobs contain no literal object ID, and their filename supplies the detector keyword. The tag-split case proves why an independent replay is required instead of guessed decoded line offsets. Output contains safe classifications and digests, never matched values. This controlled proof does not replace the hosted review's own prompt/schema/source scan.


steipete:
@clawsweeper re-review

The current head's full CI has passed. The PR body now records that result, the unchanged-head coverage-collection retry, and the separate local loopback-test limitation. Runtime source and native behavior proof are unchanged.

